PR #36, opened 2015-01-19
Backend v4

### LINKED ISSUE #32: Students code needs to be isolated from code used for grading.
This issue has already been raised offline. This bug is to track the progress and schedule the task at some milestone.

### LINKED ISSUE #33: Detect memory outage.
When a submission hits the memory limit on the container, it would be nice to have a distinct error message instead of "failed". 

I do not know if it is possible to detect this.
In particular, some code may end with a sucessful feedback even though some process was killed for its abusive memory consuption.
In any case, this is not an expected scenario.

GuillaumeDerval:
INGInious is now separated in three distinct parts: the ```frontend```, the ```backend``` and the ```agents```

Agents will be run by distant Docker daemon on the same machine as containers, allowing it to create unix sockets between it and running containers. 

This allows us to handle issue #32. A new command, available in containers, named ```run_student``` (see UCL-INGI/INGInious-containers#1), allows to run "sub-containers" with custom commands, separating completely the (partially untrusted) grading and the (completely) untrusted code execution.

As the agent runs on the same machine as the other containers, we have access to cgroups for more fine-grained info about the containers. This allows to fix the timeout management (that is currently insecure as done in the container) and will help to fix #33.

The new ```backend``` will simply redirect the requests to an agent.

The goal is to make the changes completely invisible for the administrators (sysadmins ...) of INGInious instances. The agent should be launched automatically by the backend on startup. This PR then implements a new system for synchronizing task directories between the backend and the agents (currently, this has to be done manually).

Todo-list
------------

- [x] Backend implementation
- [x] Basic agent implementation
- [x] ```run_student``` (#32)
- [x] Timeout detection
- [x] OOM detection (#33)
- [x] Real logging system (no more print)
- [x] Task directory sync between backend and agents
- [x] Docker container image for the agent
- [x] Script for Boot2Docker (simplifying the usage of INGInious on OS X)
- [x] Documentation (docstrings)
- [x] Unit tests for the backend
- [x] Remote agent autostart
- [x] Configuration for the agent
- [x] Documentation (Read the Docs)
-  ~~Unit tests for the agent~~ (... the test framework would be too huge... If someone feels motivated, we are open to PR!)


layus:
Any idea on when this will get merged ?

GuillaumeDerval:
It is more or less ready, but I *really* want to take time to make unit tests. I don't want to set a deadline for this yet.

GuillaumeDerval:
> It is more or less ready, but I really want to take time to make unit tests. I don't want to set a deadline for this yet.

Ok, it was not more or less ready at the time...
But it's merged now @layus ;-)

layus:
Nice ! I will test it on linux and do a pull-request :-)
(see: https://github.com/GuillaumeDerval/INGInious/commit/617cf5ad06be33a0d457bc199a084b46a0f56aa5)

Issue #3, opened 2014-08-17
Remove the "custom" type in the task editor


GuillaumeDerval:
The custom subproblem type only allows to create "boxes" in code subproblems.

There should be an interface to do it without editing json code manually.

anthonygego:
I suggest to extract or delete (is it even used somewhere ?) the "custom" problem type from the code base and rely on subproblems instead. This will remove the need for boxes that makes the code problem implementation quite complicated. In case of extraction the boxes code can be put in a plugin.

GuillaumeDerval:
This was a feature of Pythia 1, and if I remember well it was used only by a single exercise which may have been discarded since then.

It allowed to create things like 
```
for(int i = <empty box>; <empty box>; i++) {
    <empty box>
}
```

It can be really useful, but now that INGInious has evolved a lot, it will certainly better fit in a plugin.

Drumor:
This is not relevant anymore

PR #44, opened 2015-03-20
Yaml

### LINKED ISSUE #21: reStructuredText values in JSON strings drive me crazy :-)
It is nearly impossible to enter reStructuredText in json strings, and there are cases where the user enters a simple string that gets involuntarily evaluated as restructured text.

My point is that in tasj.json, strings should only be plain strings, or html chunks.
In task.rst, everything should be reStructuredText (obviously).

So
1) parsing of strings in task.json should be disabled by default, and
2) Either task.rst or "{context,name,header}IsHTML" should be documented to avoid asking the user to enter reStructuredText values in JSON strings.

What do you think of that ?

GuillaumeDerval:
Add a YAML parser for the tasks, for the course descriptions and even to replace configuration.json.
Modify a bit the studio to use YAML instead of JSON (more simple to modify manually...) (linked to #3)
Updated the documentation to reflect the new default.

This is still perfectly backward compatible. I indicated in the comments that the JSON and RST parsers for the task are (will be, after the merge) deprecated.

Fixes #21.

@31416r, what is your opinion on this?


31416r:
I like it, but I would prefer to improve on the RST parser than remove/deprecate it completely. This may be something that I could/should do myself. :-)

Maybe the best is to adopt the YAML parser as the main parser and to have other languages/formalisms as addons. What would be the cost to have some kind of "plugin mechanism" allowing to add new parsers ?

layus:
I think that the rst parser is not appropriate for the task at hand.
It is not easy to encode configuration options in reStructuredText, and it has not been designed for that task.

Ideally, we would use some lightweight markup syntax that describes the task. But we have reached a state where we support too many option that disrupt the flow of a RST document.

Yaml is a very good alternative as it is designed to write efficiently configuration options.
I see it a a relaxed json.

This does not mean that we have to abandon RST, as we will need it for some string values.
For example, the following task.yaml contains markdown strings. (freely extended :-), the syntax might not be recognised by the current yaml parser)

```yaml
---
  environment: default
  taskfs: test/test.sfs
  author:
    - Guillaume Derval
    - Pierre Dupond
  limits: {time: 60, memory: 1Gi, disk: 32Mi, output: 100ki}
  
  name: Un titre d'énoncé farfelu
  context: |
    Contexte de la mission
    ======================
    
    Ceci est le contexte écrit en _markdown_ ou en _reStrucuredText_,
    selon ce que vous préférez.

    Théoriquement, on peut y inclure
    * des listes,
    * avec des items
    
    et même
    1. les numéroter
    2. ou non ;-)

  problems: 

    # Premier problème sur les ...
    idFirstExercise:
      name: The first exercice
      type: code
      header: The first exercice consists of blablabla
      headerType: markdown
      language: java

    idSecondExercise:
      name: The second exercice
      type: code
      header: "<p>The second exercice consists of blablabla</p>"
      headerIsHTML: true
      boxes: 
 ...
```

for comparison, here is the equivalent task.json

```json
{
  "environment": "default", 
  "context": "Contexte de la mission\n======================\n\nCeci est le contexte \u00e9crit en _markdown_ ou en _reStrucuredText_,\nselon ce que vous pr\u00e9f\u00e9rez.\n\nTh\u00e9oriquement, on peut y inclure\n* des listes,\n* avec des items\n\net m\u00eame\n1. les num\u00e9roter\n2. ou non ;-)\n", 
  "name": "Un titre d'\u00e9nonc\u00e9 farfelu", 
  "limits": {
    "output": "100ki", 
    "disk": "32Mi", 
    "time": 60, 
    "memory": "1Gi"
  }, 
  "author": [
    "Guillaume Derval", 
    "Pierre Dupond"
  ], 
  "taskfs": "test/test.sfs", 
  "problems": {
    "idSecondExercise": {
      "header": "<p>The second exercice consists of blablabla</p>", 
      "boxes": null, 
      "type": "code", 
      "name": "The second exercice", 
      "headerIsHTML": true
    }, 
    "idFirstExercise": {
      "header": "The first exercice consists of blablabla", 
      "type": "code", 
      "name": "The first exercice", 
      "language": "java", 
      "headerType": "markdown"
    }
  }
}
```

It would make sense to also compare with the RST version, but I am unable to write it correctly.

GuillaumeDerval:
@31416r I moved RST and JSON to plugins with b161e6b71b60c6b167b9eeb4689222b63049112f :-)

@layus that's exactly what is implemented right now (without the Markdown) :-)

Implementing alternatives "markup languages" is clearly something I want to do. Currently we have all these headerIsHTML, contextIsHTML, ... that are used for backward compatibility with Pythia, but I would prefer to have a single field ```markup``` per task file that indicates what is the markup language for the whole task.

It will then be simpler to implement the usage new markup languages like Markdown.

I'm going to create a tool that converts all configuration files (configuration.json, course.json, task.json and task.rst) to their YAML equivalent, should I take the opportunity to implement this ```markup``` field as it will probably not be backward compatible? Or should I leave it for later?

layus:
@GuillaumeDerval Please do not implement a one-to-one conversion tool. It is much more preferable to have an internal representation of a task/course/configuration and implement two methods (import/export) for each supported configuration language.
When you have that, a conversion tool is trivial, and you can reuse the code to generate Yaml from the web configuration pane.

I suppose that it is already what you intend to do, but one is never too sure ;-)

GuillaumeDerval:
That's what is done and what I plan to do ;-)

31416r:
I agree with Layus on the converter thing and the way to add new configuration languages (@Layus : a Markdown plugin soon ?)

With respect to the deprecation of JSON and RST, I think that the Markdown option is so much more convenient that they will fade out be themselves without any need to formally push them out.

This seems good to me : good work !

PR #58, opened 2015-06-29
Use the test task folder for docker testing

layus:
This fixes an error when starting inginious.
This does not have to be merged if this is not the correct way to fix the problem.


GuillaumeDerval:
Not relevant anymore :-(

PR #59, opened 2015-06-29
Fix database access error ?

layus:


GuillaumeDerval:
Can you upgrade pymongo and check again?
They changed a lot how the lib works and I had to made a lot of changes too (see c356ef7be3bdf3d21c858018b1b0f93290917887), so that's maybe the problem here :-)

layus:
Ok, seems fixed indeed.
Anoter question :  why do you need to use list(...) everywhere ?

PR #142, opened 2016-08-28
Patch add wsgi

dirkcgrunwald:
Patches to add WSGI scripts and documentation to show how to use those (as well as some troubleshoot for LTI).

The configuration of the WSGI scripts leaves something to be desired (editing the file) but it appears it's not easy to pass in this information from the Apache config file.


dirkcgrunwald:
test this please

anthonygego:
Thank you. I merged the scripts so there is only one file to use per application. I tested this with Apache and lighttpd. Tell me if you have any problem.

PR #143, opened 2016-08-30
Patch lti grading methods

dirkcgrunwald:
In the current version of the LTI interface, the lis_outcome_manager always reports the MAX score to the TC (i.e. your LMS gradebook).  I initially added another field to the task items (lti_grader_method) and then saw that I should really use the 'evaluate' method.

I'm not completely happy with this because I would like to have the option of setting the grading method (last/best) in the custom_grading field so that instructors can use the same questions in formative and summative assessment. However, in the current setup, the LTI parameters are not available when doing the evaluation, so this will have to do.


anthonygego:
OK I merged this in eae7631ff6. Thank you. Btw, I suppose the grade sort should have been done in descending order to ensure the best is kept. I fixed this in a later commit.

PR #145, opened 2016-08-30
Patch specify lti username

dirkcgrunwald:
the LTI standard uses "user_id" as the user identifier, but most LMS's also support ext_user_username, which is typically the unique identifier for the person (e.g. in moodle, this is the account name).

We use a SSO system with ldap + shib, and would like to have the data in the Moodle and Inginious key off the same name rather than the numeric ID of the moodle user.

grade changes back to the Moodle TC still work with this change. Users can specify lti_user_name: ext_user_username in the YAML config file.

Forgot to add a doc for it, but can do it if this patch is OK.


anthonygego:
LGTM. Thank you ! Some words about this in the docs would be great.

Issue #146, opened 2016-08-30
Unifying LTI / webapp view of users


dirkcgrunwald:
I would like to have users who only access the system through the LTI interface have their userid added to the course and then have their scores visible in the webapp interface.

In our scenario, students may use the LTI to submit their data, but the teachers / admins could use the webapp interface to e.g. download all submissions or run bulk grading options.

This means that for us, course enrollment would be set up by Moodle and we wouldn't have to add them to Inginious as well.

Is there a reason this doesn't already work? Is it as simple as adding the users to the course in mongo?

dirkcgrunwald:
I'd be happy to explore other ways to resolve this feature request.

The specific need to integrate the Web and LTI frontends have to do with allowing the instructor to perform batch operations, such as running the JPLAG checker or downloading all of the assignments.

Some options for solutions include:
1) Could add all LTI submissions into a 'group' defined by the LTI context information (i.e. same launching point from the LTI consumer). This would disambiguate the web and LTI uses.

2) Could provide "LTI web interface" when the instructor visits the LTI link using the LTI provided role information. That interface could provide the batch operations.

I don't like solution #2 because, in practice, an instructor is going to be aware of the web interface when designing tasks.

I'd like to find some design that could be implemented in the next 20-40 days so that we could deploy INGInious at our university.

anthonygego:
I'm not sure to get the point for your first solution. Could you be more specific ?

For now, I think the patch you comitted could do the trick. Everything depends on your webapp usage. As long as you don't configure a task for grouped submission on the LTI, there will be no problem. And if this would happen, it won't crash: you'll just probably see weird results in the submissions as the LTI will allow a unique student to submit to a task the webapp would disallow if the task is configured as grouped task.

This really depends on your usage: you can, for instance, tell the students to submit on Moodle, use your patch to make them register automatically, set the courses not accessible to the students on the webapp and use it for administration only. This setup should work fine. 

In this case, by working a little bit your patch and moving the functions modifying data in the the DB into the common files, I think it is possible to make it less sensitive to the possible changes we would make until finding a better solution.

We'll probably separate the batch containers and studio from the webapp in the near future. A download feature is available for the LTI, it is not as complete, but we'll try to make it better. Common statistics would however require more work, but we are currently investigating new architecture for the frontends.

dirkcgrunwald:
This scenario:

[...]

PR #148, opened 2016-09-02
enroll LTI users if autoenroll is true

dirkcgrunwald:
This PR implementations a solution to the issue I posted. If your LTI config file sets "autoenroll: True", then requests submitted via LTI cause the student to be enrolled in the course and to have a users_task record to be added to the MongoDB.

The end result is that you can download the submissions via the webapp and I think you should be able to run "batch items" against the submissions (haven't tried).

Ideally, this would have some method to create a "classroom" that uses the LTI provided course information.

Also included documentation updates.

### DIFF
--- a/doc/install_doc/config_reference.rst
+++ b/doc/install_doc/config_reference.rst
@@ -62,6 +62,24 @@ The different entries are :
     See :ref:`plugins` for detailed information on available plugins, including their configuration.
 
 .. _configuration.example.yaml: https://github.com/UCL-INGI/INGInious/blob/master/configuration.example.yaml
+
+LTI Configuration reference
+=======================
+The LTI interface uses most of the same configuration options as the webapp as well as the following:
+
+``lti``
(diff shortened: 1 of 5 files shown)


anthonygego:
We clearly understand your need as explained in issue #146 and think that it could be nice for people using both webapp and LTI with same user base.

I merged this with our recent changes in a separate branch. However, I'll not merge it into our master now as this is only compatible with some specific configurations. Actually, the submission username is an array because grouped submissions authors (classroom groups or teams) are stored like that in DB. In the case of grouped submissions either from the webapp or the LMS, inconsistency is likely to happen. Moreover, it somehow strengthen the link between the  webapp and LTI that we'd like to eliminate. 

We are planning to rethink the way the webapp and LTI work in the future in order to ease the access to the information provided by the webapp to both users (submission downloads e.g.), and not to force LTI-only users to set up the webapp for task edition.

PR #159, opened 2016-10-18
INGInious.css: One-click selection of SSH info

oliviertilmans:


GuillaumeDerval:
Thanks!

The CSS are generated from LESS files, notably this one:
https://github.com/UCL-INGI/INGInious/blob/master/inginious/frontend/common/static/css/common.less#L123

Please update this file and regenerate all the CSS files (with the LTI ones) using the grunt file provided in utils.

PR #169, opened 2016-11-09
Improve the usage of TemplateHelper outside of the main source (for p…

GuillaumeDerval:
…lugins) by selecting by default the base layout.

Remove improper passing of the inginious package root path from the app to the helper, and the method added previously to get this path from TemplateHelper, which made no sense either.
Add instead a function get_root_path() in the INGInious package itself, which is of course always available.

This may break some plugin as the newly "layout" param for get_custom_renderer has not the same signification;
I then changed the name of the method from get_custom_template_renderer to get_custom_renderer to make other devs notice the change.
The new name is also more consistent with the name of the other methods.

@anthonygego you should be happier now.

### DIFF
--- a/inginious-lti
+++ b/inginious-lti
@@ -87,7 +87,7 @@ if __name__ == "__main__":
     application = CustomLogMiddleware(application, logging.getLogger("inginious.lti.requests"))
 
     # Launch the app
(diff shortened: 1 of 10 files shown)


GuillaumeDerval:
Hey, it seems I have to sign my own license agreement :D
This PR also contains a bonus bugfix in the second commit.

PR #183, opened 2017-02-07
Implement a way to get back the job queue from the backend

GuillaumeDerval:
Add a status page. Add info about the waiting queue when a submission is running.

Also fix an old bug with the "kill submission" button not showing when reloading the task page.


inginious/frontend/common/static/js/task.js
```diff
@@ -518,13 +523,27 @@ function getLoadingAlertCode(content, submissionid)
...
-    task_alert.html(getLoadingAlertCode("<b>Verifying your answers...</b>", submissionid));
-    $('html, body').animate({
-        scrollTop: task_alert.offset().top - 100
-    }, 200);
+    var content = '<i class="fa fa-spinner fa-pulse fa-fw" aria-hidden="true"></i> ';
+    if(submission_wait_data != null && "nb_tasks_before" in submission_wait_data && "approx_wait_time" in submission_wait_data) {
+        var nb_tasks_before = submission_wait_data["nb_tasks_before"];
+        var wait_time = Math.round(submission_wait_data["approx_wait_time"]);
+        if(nb_tasks_before == -1 && wait_time <= 0)
+            content += "<b>The genius snake is evaluating your answer right now.<b/> (almost done, we promise!)";
```
anthonygego: All those geeky messages should be made more serious : this will exacerbate the fact younger students think that everything is magic and make serious users go away from using this in a serious environment.

inginious/frontend/webapp/pages/queue.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INGInious. See the LICENSE and the COPYRIGHTS files for
+# more information about the licensing of this file.
+
+""" Maintenance page """
```
anthonygego: Maintenance...

inginious/frontend/common/static/js/task.js
GuillaumeDerval: You have no humor ;-)

PR #188, opened 2017-02-25
Make LDAP logs more detailed

bogem:


bogem:
@INGIniousBot ok to test

PR #190, opened 2017-03-01
More stability in LdapAuthMethod/UserManager.get_users_info

bogem:

PR #192, opened 2017-03-01
Fix changes of #190

GuillaumeDerval:
PR #190 did not LGTM.

@anthonygego can you test (;-)) this commit asap?


inginious/frontend/webapp/plugins/auth/ldap_auth.py
```diff
@@ -75,9 +75,9 @@ def auth(self, login_data):
                 realname = user_data[self._cn][0].decode('utf8')
                 return (username, realname, email)
             except KeyError as e:
-                self._logger.info("Can't get field " + str(e) + " from your LDAP server")
+                self._logger.error("Can't get field " + str(e) + " from your LDAP server")
             except Exception as e:
-                self._logger.info("Can't get some user fields: " + str(e))
```
GuillaumeDerval: I changed it to `exception`, which [will show the whole exception message and stacktrace](https://docs.python.org/3/library/logging.html#logging.Logger.exception) :-)

PR #196, opened 2017-03-26
Added Nginx documentation

izayoi-ami:
A quick setup for Nginx is written.


izayoi-ami:
ok to test

anthonygego:
Configuring the webserver as a proxy to the the webapp application does not seem to be a good idea. You'll probably be able to distribute the load among several instances but performance will be worse as you'll still use the Python-integrated HTTP stack and thus only add overhead, except for the static files.

To achieve much more performance, you should probably look at ways to use FastCGI or WSGI with Nginx.

GuillaumeDerval:
Web.py provides a small cookbook about nginx: http://webpy.org/cookbook/fastcgi-nginx

It's probably a good starting point :-)

GuillaumeDerval:
Closing this for now.

PR #199, opened 2017-04-09
Add cursor:pointer for .list-group-item

bogem:
On some machines the cursor isn't correct

### DIFF
--- a/inginious/frontend/webapp/static/css/INGInious.css
+++ b/inginious/frontend/webapp/static/css/INGInious.css
@@ -285,6 +285,9 @@ form#task label {
 .list-group.list-group-progress * {
   line-height: 22px;
 }
+.list-group-item {
+  cursor: pointer;
+}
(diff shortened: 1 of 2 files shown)


GuillaumeDerval:
This is more a problem with BootStrap than with our theme; moreover, it shouldn't be a pointer when the list-item does not hold a link.

Can you be more specific on the affected platforms/browsers/pages?

bogem:
But where it can hold not a link?
I saw it on some student's PC. It was Windows, but I don't remember the browser. Chrome or Opera.

bogem:
I don't have this issue on my macOS with last Chrome

anthonygego:
There are several list group items that are not links such as in the "Submitting as" list in the task page or the student list in classroom pages.

GuillaumeDerval:
Closing this for now. If you spot the affected browser, please open a new issue :-)

PR #202, opened 2017-04-17
Add a mechanism to support cookieless sessions

GuillaumeDerval:
To create a cookieless session, prefix the url with `/@@` (`http://www.instance.be/@@/index` for example).
Each page everywhere in the app will then be prefixed with the session_id, in the form `/@session_id@`.

This is completely transparent from the app point of view, everything is handled by the web.py Application class and its associated Session object.


GuillaumeDerval:
Close this in favor of #226

PR #242, opened 2017-08-24
Add Google auth

0xDCA:
This PR adds Google as an authentication method. It also supports a domain option to allow restricting the authentication to a specific domain.

Note: I'm not sure if using the logo without the rest of the button is allowed by their branding guidelines (and I think this applies to the rest of the sign-in buttons as well).


anthonygego:
LGTM. Actually FontAwesome already embed the Google logo under ``fa-google`` class, so there's no need to include it. Can you update this ? There shouldn't be any issue as the brand icons clearly refers and link to the associated company services.

0xDCA:
I am aware of the Google icon in font-awesome, but it does not have the original colors, so it looks really weird (given that the other icons are colored). What do you think?

0xDCA:
Friendly ping

anthonygego:
I test this as soon as I can ;)

anthonygego:
It's OK for me. However, in order to keep compatibility with Google accounts that do not include the Google+ service (for instance, some academic accounts), I'll switch the userinfo URL back to the simple OAuth2 way. This seems to imply that the domain can not be checked anymore.

PR #260, opened 2017-10-18
INGInious : Tags

BIG-COW:
With the tag mechanism, we can tag tasks and submissions with tags. For example, we can create a tag 'Compile' that will be set to true if the student code compile or false if not. This mechanism can be useful to manage a huge set of submissions (for example sort submission, I only want to see submissions of students that compile, ...) and get statistics about submissions (how many submissions compile ?, ...).
We can also tag tasks for organizational purposes. For exemple, a tag 'Mission 1' can be tagged in all tasks of the mission 1. We can then filter tasks based on it.

If you want to test this pull request, you can define your tags in the **_Edit task_** > **Tags** section of an INGInious task.

To set tags to _true_ (green) we should use the tag API present in the feedback API.
For example, with a run script in Python: `feedback.set_tag("compile", True)`.
If the tag is set explicitly to False or is not set at anytime, the tag will be blue (considered as False).

Tags are displayed in **_View submissions_** page to help tutors and administrators to identify difficulties of student easily.
Also, we can find statistics about tags in classroom results.


inginious/agent/docker_agent/__init__.py
```diff
@@ -490,7 +490,7 @@ def close_and_delete(student_container_id=student_container_id_loop):
                             raise Exception("Feedback file is badly formatted.")
                         elif accepted_types[key] == dict:
                             for sub_key, sub_item in item.items():
-                                if not id_checker(sub_key) or isinstance(sub_item, dict):
+                                if not id_checker(sub_key): # or isinstance(sub_item, dict): Why? Very restrictive.
```
GuillaumeDerval: This protects the key "custom", "problems" and "tests". I agree that "custom" should be able to contain anything, but "problems" and "tests" must contain a precise format, that should be checked here. Please modify the code to something like

```
elif accepted_types[key] == dict and key != "custom": #custom can contain anything
    for sub_key, sub_item in item.items():
        if not id_checker(sub_key) or isinstance(sub_item, dict):
            raise Exception("Feedback file is badly formatted.")
```

inginious/common/tags.py
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INGInious. See the LICENSE and the COPYRIGHTS files for
+# more information about the licensing of this file.
+
+class Tag:
```
GuillaumeDerval: A tag could be a tuple (id, name), thus avoiding the unneeded class. `parse_tag_str` could be a simple function available in this file. It would be even better with a function name like `parse_tags_from_string`.

BIG-COW: I create a class for the tag for extensibility reasons. For example, we could add lots of fields like 'is_shown_to_student' to keep the tag hidden to students (only for admin), 'color' if we want to customize the tag rendering, 'explanations' to get information of the purpose of the tag with a tooltip HTML for example. I think this would be easier to add these things with a class instead of a tuple. What do you think ?

GuillaumeDerval: Ok then :-)

inginious/agent/docker_agent/__init__.py
BIG-COW: Thank you, I just updated the code with your suggestion !

anthonygego:
I think we should merge ``rdict["custom"][tags"]`` with ``rdict["tests"]`` as those features are quite similar. Ideally, a displayed tag is a test that is shown to students. Current tests should therefore become tags that are only visible for admins. Those could eventually be displayed in the frontend too. What do you think ?

How are tags exactly defined in the ``task.yaml`` file ? It would be nice to have them listed as dictionary as it will ease the string extractions for translations.

anthonygego:
@BIG-COW Commits are coming and coming, please tell us when you think something is mergeable. This will be easier.

anthonygego:
@BIG-COW Can you pull the master branch on your local repo branch ``<branch>`` and perform a rebase using ``git rebase <branch>`` in your working branch ? You will probably have some conflicts to resolve on the way as you performed merges before... This will clean the PR commit history. Thanks.

BIG-COW:
@anthonygego I have lots of strange conflits in `inginious/frontend/plugins/auth/twitter_auth.py` when I rebase. Can I come tomorrow at your office to resolve that with you ?

anthonygego:
@BIG-COW OK

base-containers/base/inginious/feedback.py
```diff
@@ -86,6 +86,27 @@ def set_problem_feedback(feedback, problem_id, append=False):
     save_feedback(rdict)
 
 
+def set_tag(tag, value):
```
anthonygego: Maybe we should merge this completely with ``tests``, by removing the ``inginious.tests`` module. Moreover, as those values are now always returned and not only for debug, putting them in a separate file is no more useful. You can merge it directly in the feedback dictionary. You'll have something like : 

    rdict = load_feedback()
    tests = rdict.setdefault("tests", {})
    tests[tag] = value == True
    save_feedback(rdict)

base-containers/base/inginious/feedback.py
```diff
@@ -86,6 +86,27 @@ def set_problem_feedback(feedback, problem_id, append=False):
...
+    :param value: should be a boolean 
+    :param tag: should be the id of the tag 
+    """
+    if (value == True):
+        inginious.tests.set_result(tag, True)
+    else:
+        inginious.tests.set_result(tag, False)
+        
+_current_auto_tag = 0
+def tag(value):
```
anthonygego: ``funcname.variable`` is better for such static variables.
I suggest to prepend ``auto-tag`` with a reserved char to avoid user to set them manually.
Actually, if only a distinction between tags and auto-tags are needed, I would recommend to replace the tag id by a reserved char and a random hash, because static variables will be reset if we call the API using the shell, which is a desired feature for non-Python users.

I'd re-use the definetest code to define a new shell command such as ``tag-result`` or ``tag-set``.

inginious/frontend/courses.py
```diff
@@ -10,7 +10,11 @@
 
 from inginious.common.courses import Course
 from inginious.frontend.accessible_time import AccessibleTime
+from natsort import natsorted
 
+_all_tags_cache = {}
```
anthonygego: Caches are accessed using ``self._name`` key, meaning that only the course-related value is required for for a given course. Why aren't such variables not class attributes ?

By the way, can tags be loaded in the class ``__init__`` method ?

inginious/frontend/pages/course_admin/aggregation_task.py
```diff
@@ -38,7 +39,23 @@ def page(self, course, aggregationid, task):
...
-        return self.template_helper.get_renderer().course_admin.aggregation_task(course, aggregation, task, data)
+            
+        # Get best submissions (submission for evaluation)
+        # Need this for statistics computation
+        # Not really optimized for now...
+        user_tasks = list(self.database.user_tasks.find({"username": {"$in": aggregation["students"]}, 
+                                                       "courseid": course.get_id(),
+                                                       "taskid": task.get_id()},
+                                                       {"submissionid": 1, "_id": 0}))
+        best_submissions_list = [] # list containing Ids of best submissions
```
anthonygego: ``best_submissions_list = [ut["submissionid"] for ut in user_tasks]`` would be better processed by Python. However, I'm wondering if only one field is required if mongo function ``distinct`` would not do the job better, as submissionids are unique.

inginious/frontend/pages/course_admin/statistics.py
```diff
@@ -0,0 +1,55 @@
...
+    total_submission_best = 0
+    total_submission_best_succeeded = 0
+        
+    for submission in data:
+        if "best" in submission and submission["best"]:
+            total_submission_best = total_submission_best + 1
+            if "result" in submission and submission["result"] == "success":
+                total_submission_best_succeeded += 1
+        
+    statistics = []
```
anthonygego: Don't be afraid to put the code on several lines rather than calling ``append`` several times

inginious/frontend/pages/course_admin/task_edit.py
```diff
@@ -205,8 +206,38 @@ def POST_AUTH(self, courseid, taskid):  # pylint: disable=arguments-differ
...
+                        if tags[k2]["type"] != 2 and tags[k2]["id"] == id:
+                            count = count+1
+                    if count > 1:
+                        return json.dumps({"status": "error", "message": _("Some tags have the same id! The id of a tag must be unique.")})
+            #Force no id if organisational tag
+            for k in tags:
+                if tags[k]["type"] == 2:
+                    tags[k]["id"] = ""
+                
+            data = {key: val for key, val in data.items() if not key.startswith("problem") and not key.startswith("limits") and not key.startswith("tags")}
```
anthonygego: There are several identical loops on ``tags`` here. Can this be simplified ?

inginious/frontend/pages/tasks.py
```diff
@@ -308,6 +310,16 @@ def submission_to_json(self, data, debug, reloading=False, replace=False):
...
             tojson["input"] = data.get('input', {})
 
+        if("tests" in data):
+            tojson["tests"] = {}
+            for tag in tags[0]+tags[1]: # Tags only visible for admins should not appear in the json for students.
+                if (tag.is_visible_for_student() or debug) and tag.get_id() in data["tests"]:
+                    tojson["tests"][tag.get_id()] = data["tests"][tag.get_id()]
+            if debug: #We add also auto tags when we are admin
+                for tag in data["tests"]:
+                    if tag.startswith("auto-tag-"):
```
anthonygego: This can thus be replaced by a reserved char, as mentioned before.

inginious/frontend/templates/course_admin/aggregation_task.html
```diff
@@ -73,6 +129,27 @@ <h2>$:_("Submissions for the task {}").format(task.get_name(user_manager.session
...
+                    </span>
+                $for tag in task.get_tags()[1]:
+                    <span style="line-height: 18px;">
+                    $if "tests" in submission and tag.get_id() in submission["tests"] and submission["tests"][tag.get_id()]:
+                        <span class="badge alert-danger" id="$tag.get_id()">$:_(tag.get_name())</span> 
+                    </span>
+                $# Auto-tags not defined in task
+                $if "tests" in submission:
+                    $for tag in submission["tests"]:
+                        $if tag.startswith("auto-tag-"):
```
anthonygego: Same remark.

inginious/frontend/templates/course_admin/edit_tabs/tags.html
```diff
@@ -0,0 +1,148 @@
...
+    </thead>
+    <tbody>
+    $ i=-1
+    $for tag in task_data.get('tags', {}):
+        $ i=i+1
+        $ id = task_data.get('tags')[tag]["id"] if "id" in task_data.get('tags')[tag] else "unknow_id"
+        $ name = task_data.get('tags')[tag]["name"] if "name" in task_data.get('tags')[tag] else "Unknow name"
+        $ description = task_data.get('tags')[tag]["description"] if "description" in task_data.get('tags')[tag] else ""
+        $ type = task_data.get('tags')[tag]["type"] if "type" in task_data.get('tags')[tag] else 0
+        $ visible = task_data.get('tags')[tag]["visible"] if "visible" in task_data.get('tags')[tag] else False
```
anthonygego: You should replace the loop by something like ``for key, tag in task_data.get('tags', {}).items():``

inginious/frontend/templates/course_admin/edit_tabs/tags.html
```diff
@@ -0,0 +1,148 @@
...
+                <select class="form-control" name="tags['+ new_id +'][type]" onchange="change_id_field(this, '+ new_id +')"> \
+                    <option value="0" '+checked0+'>$:_("Common")</option> \
+                    <option value="1" '+checked1+'>$:_("Antitag")</option> \
+                    <option value="2" '+checked2+'>$:_("Organisation")</option> \
+                </select> \
+            </td> \
+        </tr>'
+    $$('#table').find('tbody').append(new_row);
+}
+</script>
```
anthonygego: I'd rather have the script part in the ``studio.js``file. However, I'd put the big html snippet directly into the html code as a hidden template and set the appropriate field in JS after copying the object.

inginious/frontend/templates/course_admin/student_task.html
```diff
@@ -73,6 +75,27 @@ <h2>$:_("Submissions for the task {}").format(task.get_name(user_manager.session
...
+                    </span>
+                $for tag in task.get_tags()[1]:
+                    <span style="line-height: 18px;">
+                    $if "tests" in submission and tag.get_id() in submission["tests"] and submission["tests"][tag.get_id()]:
+                        <span class="badge alert-danger">$:_(tag.get_name())</span> 
+                    </span>
+                $# Auto-tags not defined in task
+                $if "tests" in submission:
+                    $for tag in submission["tests"]:
+                        $if tag.startswith("auto-tag-"):
```
anthonygego: Same remark.

inginious/frontend/templates/task.html
```diff
@@ -75,7 +75,42 @@ <h3>$:_("Information")</h3>
...
+    $for tag in task.get_tags()[0]:
+        $if tag.is_visible_for_student() or user_manager.has_staff_rights_on_course(course):
+            $ number_of_visible_tags += 1
+    $for tag in task.get_tags()[1]:
+        $if tag.is_visible_for_student() or user_manager.has_staff_rights_on_course(course):
+            $ number_of_visible_antitag += 1
+
+    $# Auto tags may appear for admins. So we force the h3 title even if there is not predefined tags
+    $# We do not show the tag header for students if they can not get tags
+    $if (number_of_visible_tags > 0) or (number_of_visible_antitag > 0) or user_manager.has_staff_rights_on_course(course):
```
anthonygego: Using list comprehension and filtering to form new ``visible_tags`` and ``visible_antitags`` lists would probably avoid iterating several times here.

Something like : 

    visible_tags = [tag for tag in task.get_tags()[0] if tag.is_visible_for_student() or user_manager.has_staff_rights_on_course(course)]

And then ``if len(visible_tags)`` and ``for tag in visible_tags:``.

BIG-COW:
Hello, about the reserved char, the id has to pass trough the function 
https://github.com/UCL-INGI/INGInious/blob/master/inginious/common/base.py#L16
calling by
https://github.com/UCL-INGI/INGInious/blob/master/inginious/agent/docker_agent/__init__.py#L491

The only char allowed are `-` and `_` so I can not really use a reserved char.
What can I do ?

anthonygego:
I guess that we can add an exception to allow an additional beginning char when ``key == "tests"``. 

Maybe a more elegant way would be to associate a sub-key validating function to each key when the expected type is dict. ``tests`` key would have a different checking function than ``id_checker`` that allows such a pattern. What do you think ?

BIG-COW:
Mabye we could modify the existing function instead of create an new for each type of key ?
Like:
```
def id_checker(id_to_test, original_key=None):
    """Checks if a id is correct"""
    if original_key == "tests":
        return bool(re.match(r'[a-z0-9\-_\-*]+$', id_to_test, re.IGNORECASE))
    else:
        return bool(re.match(r'[a-z0-9\-_]+$', id_to_test, re.IGNORECASE))
```

BIG-COW:
I also need to use the `_("str")` function of gettext in `inginious/frontend/courses.py` but I have the error `UnboundLocalError: local variable '_' referenced before assignment`. What is the trick to have access to the shortcut `_` ? I use gettext.gettext() for the moment.

anthonygego:
You won't have to create a new function for each type of key, you'll simply assign the appropriate function to each key.

    keys_fct = {"problems": id_checker, "custom": id_checker, "tests": a_new_func}

An then you'll replace:

    elif accepted_types[key] == dict:
        for sub_key, sub_item in item.items():
            if not id_checker(sub_key) or isinstance(sub_item, dict):
                raise Exception("Feedback file is badly formatted.")

by something like that :

    elif accepted_types[key] == dict:
        for sub_key, sub_item in item.items():
            if not keys_fct[key](sub_key) or isinstance(sub_item, dict):
                raise Exception("Feedback file is badly formatted.")

If the string you want to translate is course-related, you won't use ``_`` which is the webapp gettext, but the ``Course`` class has its own gettext function. Have a look at ``get_name`` for an example.

BIG-COW:
Okay, thank you ! :)

BIG-COW:
@anthonygego, I think all your requested changes are done now!

anthonygego:
I'll do a squash and merge as there are many commits editing the same files, to have a cleaner view for us.

We'll probably need to find more attractive and helpful names for each tag type such as "skill", "misconception" and "category". I keep thinking that categorization should be separated from skill concepts even if they share common ideas. But we can keep it like this for now.

Thanks a lot for this contribution. This will help to extract information from the submissions more easily.

PR #266, opened 2017-11-26
Pluggable subproblems

anthonygego:
Those changes require changing problem types containing dashes like "multiple-choice", "code-file" and "code-single-line" in YAML files, and renames them into their underscore equivalent such as "multiple_choice". This to allow javascript methods using the problem type name. 

Don't know if it is relevant to add code in custom yaml to ensure retrocompatibility and automatically rename them.

The deprecaed rst task file reader was removed from the repo as it is too dependent from the problem types.

This PR requires careful check.


anthonygego:
Any update ?

anthonygego:
@GuillaumeDerval These changes need to be merged for the end of the month.

inginious/common/tasks_problems.py
```diff
@@ -172,8 +207,28 @@ def __init__(self, task, problemid, content, translations=None):
...
             self._create_box("", {"type": "file", "max_size": content.get("max_size", None), "allowed_exts": content.get("allowed_exts", None)})]
 
-    def get_type(self):
-        return "code-file"
+    def input_type(self):
+        return dict
+
+    @classmethod
+    def get_type(cls):
+        return "code_file"
```
GuillaumeDerval: Let's replace it by "file", as it is not only used to transfer files containing human-readable code.

inginious/frontend/app.py
```diff
@@ -148,7 +150,15 @@ def get_app(config):
...
-    course_factory, task_factory = create_factories(fs_provider, plugin_manager, WebAppCourse, WebAppTask)
+    default_problem_types = {
+        problem_type.get_type(): problem_type for problem_type in [DisplayableCodeProblem,
+                                                                   DisplayableCodeSingleLineProblem,
+                                                                   DisplayableCodeFileProblem,
+                                                                   DisplayableMultipleChoiceProblem,
+                                                                   DisplayableMatchProblem]
+    }
+
+    course_factory, task_factory = create_factories(fs_provider, default_problem_types, plugin_manager, WebAppCourse, WebAppTask)
```
GuillaumeDerval: Just a side note: there should be a hook to add new types of problem here. It is problematic to use plugins as they are not initialized yet. Maybe we should use the same mechanism we use for the fs providers? (https://github.com/UCL-INGI/INGInious/blob/master/inginious/common/entrypoints.py). 

Ideally, we should also update the plugin system to discover plugins this way, but that's for another PR.

anthonygego:
Done.

PR #274, opened 2017-12-22
Updating example in doc + Sphinx-tab for more visibility

Drumor:
I tried to work a bit on the doc :)


anthonygego:
I'm wondering if there is still no better way to manage both docs. The text above examples is for the Bash API and not the Python one. So for me there is limited interest in displaying it for Python users.

anthonygego:
Any...suggestion ?

anthonygego:
After discussion we'll separate the two APIs in different pages.

PR #300, opened 2018-03-01
Input random

BIG-COW:
This add a way to generate inputs randoms for exercises. With this mechanism, each student can have a different exercises, avoiding copy/paste by students.
Here is an example of task using random inputs: https://www.dropbox.com/s/8m932u0s35tssdb/task_input_random_01.zip?dl=1

### DIFF
--- a/inginious/frontend/templates/task.html
+++ b/inginious/frontend/templates/task.html
@@ -257,13 +257,6 @@ $# Start content
     </div>
 </h2>
 $:task.get_context(user_manager.session_language())
-
-<!-- HTML Tags to edit by the task author when using random inputs-->
-<!-- height attribute is set by default to 0% to avoid blank spaces when not using it-->
-<svg id="svg" height="0px"></svg>
-<canvas id="canvas" height="0px"></canvas>
-<div id="editable_div"></div>
-
 <hr/>
 <div id="task_alert"></div>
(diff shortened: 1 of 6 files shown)


inginious/frontend/templates/task.html
```diff
@@ -383,3 +393,21 @@ <h4 class="modal-title" id="socialModalLabel">$:_("Share my result on: ")</h4>
...
 </script>
+
+$# Input random
+$# set up some useful variable that may be used by task author
+<script>
+    var __input_random_list = $input_random_list
+    var __username = "$user_manager.session_username()"
+    var __lang = "$user_manager.session_language()"
+    //Shortcut to replace some elements of the page.
+    function replace(old_str, new_str){
```
anthonygego: Actually I would merge my two previous comments : we'd let the user specify everything he wants through the ``..raw:: html`` directive and parse ``@@@field@@``-like strings along with the RST, or something else as long as it does not break rst parsing. What do you think ? 

This would require passing the fields dictionary to the rst parse function. The latter may be called outside of the frontend where data like username, lang, etc. have no sense. In this case we can pass an empty dict. 

Actually, the only way to access task data in the future without a web browser should be via the webapp API, which ensures the fields are defined.

PR #305, opened 2018-03-10
Feedback hooks

GuillaumeDerval:
Add two new hooks, `feedback_text` and `feedback_script`, to allow modifying feedback and executing scripts

`feedback_script` is separated from `feedback_text` for several reasons:
- using `<script type="text/javascript"></script>` is ugly
- plugins may override previous feedback in `feedback_text` (with the help of the new `call_hook_recursive` func) and we do not want that to happen most of the time. At least it gives more flexibility
- it ensures everything is loaded before executing the script.

A nice demo of `feedback_script` is here: https://github.com/UCL-INGI/INGInious-highlight-plugin
The news hooks have been documented.


doc/dev_doc/plugins.rst
```diff
@@ -216,6 +216,18 @@ Each hook available in INGInious is described here, starting with its name and p
 
     Adds a new helper to the instance of TemplateHelper. Should return a tuple (name,func) where name is the name that will
     be indicated when calling the TemplateHelper.call method, and func is the function that will be called.
+``feedback_text`` (``task``, ``submission``, ``text``)
```
anthonygego: If I understand well, this feature is meant for allowing later modifications on (global included) feedback based on custom fields, right ?

It is interesting for global feedback but concerning subproblems feedback I remain convinced that it would be simpler and architecturally nicer to replace the old reload feedback function, while finishing the subproblems refactoring, so that they can be simply called with JSON, or a dict. That means that current feedback would be a JSON string (or be replaced by a dict) and that problems will either parser the json either simply display the feedback as it is working now.

Therefore highlighting a line in a code-problem can be performed by the code-problem feedback loading function. If a Python pre-processing is required, maybe we can add such a method the the DisplayableProblem class. This way feedback will only be treated in a unique process whatever the problem is. This would be more similar to what we've seen on Dodona : every problem has its own way displaying feedback.

inginious/frontend/static/js/task.js
```diff
@@ -650,9 +650,9 @@ function loadOldSubmissionInput(id, with_feedback)
             {
                 updateMainTags(data);
                 unblurTaskForm();
-                if(with_feedback)
-                    loadOldFeedback(data);
                 load_input(id, data['input']);
+                if(with_feedback) // load feedback in second place as it may affect the input
```
anthonygego: Well seen :-)

doc/dev_doc/plugins.rst
```diff
@@ -216,6 +216,18 @@ Each hook available in INGInious is described here, starting with its name and p
...
+    Returns : {"task": ``task``, "submission": ``submission``, "text": ``modified_text``}
+
+    Modifies the feedback to be displayed. This hook is called each time a submission is displayed. You have to return
+    the origin ``task`` and ``submission`` objects in the return value. ``text`` is in HTML format.
+``feedback_script`` (``task``, ``submission``)
```
anthonygego: I'm wondering, as the first hook accept HTML, what is the real difference with this one, except that is does not require ``script`` tag ?

inginious/common/hook_manager.py
```diff
@@ -31,3 +31,13 @@ def add_hook(self, name, callback):
     def call_hook(self, name, **kwargs):
         """ Call all hooks registered with this name. Returns a list of the returns values of the hooks (in the order the hooks were added)"""
         return [y for y in [x(**kwargs) for x in self.hooks.get(name, [])] if y is not None]
+
+    def call_hook_recursive(self, name, **kwargs):
```
anthonygego: It looks more like a piping system than a recursive call...am I wrong ? Nice, but we should be careful on the way hooks are sequenced here...hooks must be independent.

doc/dev_doc/plugins.rst
anthonygego: OK as it must be evaluated at the end of subproblems feedback

doc/dev_doc/plugins.rst
GuillaumeDerval: You mean having something like this in the return value of the container?
```
{ "id": "submission_id",
  #...
  "feedback": {
    "@global": {
      "text": "The global feedback text in RST"
      "custom_field_1": "a custom field"
      "custom_field_2": {"something": "another custom field with a dict"}
    },
    "subproblem_1": {
      "result": "ok",
      "text": "the feedback for sp1",
      "custom_field3": "a custom field only for this subproblem"
    },
    #...
  }
}
```

PR #308, opened 2018-03-12
Share the folder "course/common" and "course/common_student" inside the containers

GuillaumeDerval:
Files stored in the subdirectory "common" of a course are available to all tasks, inside the container, at the path "/course/common". This directory is only mounted in the grading container.

Files stored in the subdirectory "common_student" of a course are available to all tasks, inside the container, at the path "/course/common_student". This directory is mounted in the grading container and in student containers.

All these folder are read-only inside the containers.

### DIFF
--- a/inginious/agent/docker_agent/__init__.py
+++ b/inginious/agent/docker_agent/__init__.py
@@ -214,13 +214,13 @@ class DockerAgent(Agent):
 
             # Copy common and common_student if needed
             # TODO: avoid copy
-            if course_fs.from_subfolder("common").exists():
-                await self._loop.run_in_executor(None, lambda: course_fs.from_subfolder("common").copy_from(None, course_common_path))
+            if course_fs.from_subfolder("$common").exists():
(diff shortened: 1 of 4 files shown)


inginious/agent/docker_agent/__init__.py
```diff
@@ -188,14 +189,20 @@ def environments(self):
                 return
 
             task_path = os.path.join(container_path, 'task')  # tmp_dir/id/task/
+            course_path = os.path.join(container_path, 'course')
+
             sockets_path = os.path.join(container_path, 'sockets')  # tmp_dir/id/socket/
             student_path = os.path.join(task_path, 'student')  # tmp_dir/id/task/student/
             systemfiles_path = os.path.join(task_path, 'systemfiles')  # tmp_dir/id/task/systemfiles/
 
+            course_common_path = os.path.join(course_path, 'common')
```
anthonygego: Non-tasks folders should not be confused with tasks folder. I suggest to use the same naming a for translations files, ``$common`` and ``$common_student``.

PR #311, opened 2018-03-15
Input random

BIG-COW:
Some bug fix about input random and submission viewer.


inginious/frontend/pages/tasks.py
```diff
@@ -97,11 +98,11 @@ def GET(self, courseid, taskid, isLTI):
 
         self.user_manager.user_saw_task(username, courseid, taskid)
         
-        # Generate random inputs and save ot into db
+        # Generate random inputs and save it into db
         random_input_list = []
-        if username is not None:
-            random.seed(username + taskid + courseid)
-            random_input_list = [random.random() for i in range(task.get_number_input_random())]
+        random.seed(str(username if username is not None else "") + taskid + courseid + str(time.time() if task.regenerate_input_random() else ""))
```
anthonygego: If I understand well, if the student opens the same task in two tabs, say Tab1 and Tab2, and then submit a correct answer in Tab1, it will fail because the random inputs have been regenerated. Maybe we should add a mechanism to, at least, refuse the submission for an outdated set of random inputs. What do you think ?

We can for instance simply return the generated set in the form and match with what was stored in DB.

BIG-COW: Yes, nice remark, I will fix this.

inginious/frontend/templates/course_admin/submissions.html
```diff
@@ -42,7 +42,7 @@ <h4>$:_('Submissions for task "{0}" - {1} ({2})').format(task_name, complete_nam
             $if str(c['_id']) in old_input['aggregation']:
                 <h4>$:_('Submissions for the task "{0}" - {1}').format(task_name, c["description"])</h4>
 
-<form id="select_form" method="post">
+<form id="select_form" method="get">
```
anthonygego: URL have size limitation, but the form can content all the usernames,... I think there is no alternative to POST the form.

BIG-COW: Yes I wanted to use the GET to enables the bookmarking of options in the page and to avoid inconsistency between the use of GET method when using other links in INGInious like 
"/admin/CS1-Java/submissions?user=bon&task=m01Q11" and the POST method when we perform queries.
I just resolve the inconsistency so, ok for the POST only.

inginious/frontend/templates/course_admin/submissions.html
```diff
@@ -426,7 +426,10 @@ <h4 class="modal-title">$:_("Download format")</h4>
 <!-- Submission table END -->
 
 </form>
-
+<script>
+    //Since we use post, we remove GET parameters in the URL to avoid inconsistency between POST and GET
+    window.history.replaceState({}, document.title, window.location.pathname);
```
anthonygego: If you really want to avoid this, it's better not to parse the web.input in ``submissions.py`` in both GET and POST cases...Actually I don't think it's a real issue...

PR #333, opened 2018-07-03
add support for python3.7

Maxmawt:
this commit adds the support for python3.7 and corrects problem due to the non compatibility of web.py with python3.7

### DIFF
--- a/inginious/common/asyncio_utils.py
+++ b/inginious/common/asyncio_utils.py
@@ -10,7 +10,6 @@
 import asyncio
 import threading
 from functools import wraps, partial
-from types import ModuleType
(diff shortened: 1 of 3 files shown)


Maxmawt:
ok to test

GuillaumeDerval:
Sorry to be so annoying; but I strongly believe we will go back to this code in one year and completely forgot why we did this :D

Can you add something like:
```
# This code was copied from web.py in order to fix it for Python 3.7+.
# Due to https://www.python.org/dev/peps/pep-0479/, Python 3.7+ will not accept StopIteration inside generators.```

Issue #331, opened 2018-06-20
Create task.yaml file directly after clicking on create task


anthonygego:
Currently this avoid to add files before a first save because the task folder is not yet created.

anthonygego:
Behaviour changed in #608

PR #335, opened 2018-07-12
Hard coding default task (Issue 331 "Create task.yaml file directly after clicking on create task")

amicis31:
#331 @anthonygego 
I write this branch as an example of creating directly the task.yaml file after clicking on create task. In this example a default configuration of the task.yaml file is hard coded in the source, this could be change to use a file instead of a hard coded configuration but i cannot think where to put this configuration file.

Also to illustrate the problem of just adding the file directly:
- The user doesn't know that the task has already been created. (In the case he decides to cancel the task creation or close the browser in the half of the process).
- Create a configuration file or add some configuration to the database has "bad smell".

This is why I want to add a default field in the entries on the database like in the "create classroom section". And to show the creation of the file I want to remove the redirection to the "edit task section".

### DIFF
--- a/inginious/frontend/pages/course_admin/task_edit.py
+++ b/inginious/frontend/pages/course_admin/task_edit.py
@@ -40,7 +40,39 @@ class CourseEditTask(INGIniousAdminPage):
         except:
             task_data = None
         if task_data is None:
-            task_data = {}
+            # Default task
+            data = OrderedDict([
+                ('accessible', True), 
+                ('author', ''), 
+                ('context', ''), 
+                ('environment', 'mcq'), 
+                ('evaluate', 'best'), 
+                ('groups', False), 
+                ('input_random', '0'), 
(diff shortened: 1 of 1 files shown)


GuillaumeDerval:
Thanks for the PR!

> And to show the creation of the file I want to remove the redirection to the "edit task section".

I think it would effectively be better to be redirected to the "Task list" page after a new task has been created. However, I don't really like the "default" name on the task list. The best would be to either use the `taskid` as default name or to add another `<input type="text">` field on the task page to directly indicate the name of the task. (@anthonygego any opinion on this?). 

Moreover, it would be nice to have an info message "The task has been created, click here to edit it" one the "task list" page when the new task is created.

We will keep your PR on hold until these tasks are done. We will probably take time to do it in the (not-so-near) future, but feel free to add commits to this PR if you feel motivated :-)

PR #339, opened 2018-07-19
Fixes #338

### LINKED ISSUE #338: Input fields grayed out after submission feedback if the answer to a single-line code is incorrect
When creating a subproblem with type "single-line code", if the student answers incorrectly, the input boxes for all subproblems of the task will still be grayed out after the submission result, with the "Status" field at the right side still indicating a running submission, and with the impossibility of going to another submission.
I'm currently fixing this problem (looks like a name conflict in Python and in JS).

Jimvy:
Solves bug where the input fields would still be grayed out after the
end of the running task if the answer to a single-line code is
incorrect.

### DIFF
--- a/inginious/frontend/task_problems.py
+++ b/inginious/frontend/task_problems.py
@@ -96,7 +96,7 @@ class DisplayableCodeSingleLineProblem(CodeSingleLineProblem, DisplayableProblem
         header = ParsableText(self.gettext(language, self._header), "rst",
                               translation=self._translations.get(language, gettext.NullTranslations()))
         return str(DisplayableCodeSingleLineProblem.get_renderer(template_helper)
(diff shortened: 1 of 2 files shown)


inginious/frontend/task_problems.py
```diff
@@ -92,11 +92,12 @@ def get_type_name(cls, gettext):
         return gettext("single-line code")
 
     def show_input(self, template_helper, language, seed):
+        print("DEBUG: !!!!!!!!!!!!!!!!!!!!!!! THIS SHOULD FAIL")
```
GuillaumeDerval: Please remove this :D

Jimvy: Oops

anthonygego:
Thanks !

PR #341, opened 2018-07-26
Add explicit encoding to long_description

mlhetland:
Without an explicit encoding, install via pip fails in environments where the default encoding is ASCII.


GuillaumeDerval:
Thanks for the PR!

PR #351, opened 2018-09-14
adding multi-port gestion to introduce ssh-container

FlyearthR:


GuillaumeDerval:
Thanks @FlyearthR :-)

There are some things to be fixed, but you are getting closer.

It actually improves the genericity of the code with no side effects, so I don't see why we wouldn't merge this once the small issues are fixed.

inginious/agent/docker_agent/__init__.py
```diff
@@ -422,8 +425,8 @@ def __new_job_sync(self, message, future_results):
                                                                                      hard_time_limit, share_network, write_stream))
                             elif msg["type"] == "ssh_key":
                                 # send the data to the backend (and client)
-                                self._logger.info("%s %s", self._running_ssh_debug[container_id], str(msg))
-                                await self.send_ssh_job_info(job_id, self._ssh_host, ssh_port, msg["ssh_key"])
+                                self._logger.info("%s %s", container_id, str(msg))
+                                await self.send_ssh_job_info(job_id, self._address_host, ports['22'], msg["ssh_key"])
```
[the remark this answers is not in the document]
GuillaumeDerval: Similarly, we will probably want that this message contains the port to which to connect. Something like msg["port"], which will default to 22. 

I don't know if we need to do this now or later. It is probably a very simple modification to do. @anthonygego, any opinion on this?

inginious/agent/docker_agent/__init__.py
```diff
@@ -524,17 +527,19 @@ def __new_job_sync(self, message, future_results):
...
+                    self._external_ports.add(p)
+                del self._assigned_external_ports[container_id]
 
             # Verify if the container was killed, either by the client, by an OOM or by a timeout
             killed = await self._timeout_watcher.was_killed(container_id)
             if container_id in self._containers_killed:
                 killed = self._containers_killed[container_id]
-                if container_id in self._running_ssh_debug:
-                    self._ssh_ports.add(self._running_ssh_debug[container_id])
+                if container_id in self._assigned_external_ports:
```
GuillaumeDerval: I think this is dead code and can be safely removed. This is already done at line 530.

inginious/agent/docker_agent/_docker_interface.py
```diff
@@ -29,19 +29,22 @@ def get_containers(self):
             "name": {                          #for example, "default"
                 "id": "container img id",      #             "sha256:715c5cb5575cdb2641956e42af4a53e69edf763ce701006b2c6e0f4f39b68dd3"
                 "created": 12345678            # create date
+                "ports": [22, 434]             # list of ports needed
```
GuillaumeDerval: Here you indicate that the ports are integer...

inginious/agent/docker_agent/_docker_interface.py
```diff
@@ -29,19 +29,22 @@ def get_containers(self):
...
             }
         }
         """
 
-        # First, create a dict with {"id": {"title": "alias", "created": 000}}
-        images = {x.attrs['Id']: {"title": x.labels["org.inginious.grading.name"], "created": datetime.strptime(x.attrs['Created'][:-4], "%Y-%m-%dT%H:%M:%S.%f").timestamp()}
+        # First, create a dict with {"id": {"title": "alias", "created": 000, "ports": [0, 1]}}
+        images = {x.attrs['Id']: {"title": x.labels["org.inginious.grading.name"],
+                                  "created": datetime.strptime(x.attrs['Created'][:-4], "%Y-%m-%dT%H:%M:%S.%f").timestamp(),
+                                  "ports": x.labels["org.inginious.grading.ports"].split(",") if "org.inginious.grading.ports" in x.labels else []}
```
GuillaumeDerval: ... but here they are clearly strings ;-)
I prefer integers.

inginious/frontend/pages/course_admin/task_edit.py
```diff
@@ -205,9 +205,9 @@ def POST_AUTH(self, courseid, taskid):  # pylint: disable=arguments-differ
 
             # Parse and order the problems (also deletes @order from the result)
             if problems is None:
-                return json.dumps({"status": "error", "message": _("You cannot create a task without subproblems")})
-
-            data["problems"] = OrderedDict([(key, self.parse_problem(val))
+                data["problems"] = OrderedDict([])
+            else:
+                data["problems"] = OrderedDict([(key, self.parse_problem(val))
                                             for key, val in sorted(iter(problems.items()), key=lambda x: int(x[1]['@order']))])
```
GuillaumeDerval: Indentation is wrong here

inginious/agent/docker_agent/__init__.py
```diff
@@ -422,8 +425,8 @@ def __new_job_sync(self, message, future_results):
                                                                                      hard_time_limit, share_network, write_stream))
                             elif msg["type"] == "ssh_key":
                                 # send the data to the backend (and client)
-                                self._logger.info("%s %s", self._running_ssh_debug[container_id], str(msg))
-                                await self.send_ssh_job_info(job_id, self._ssh_host, ssh_port, msg["ssh_key"])
+                                self._logger.info("%s %s", container_id, str(msg))
+                                await self.send_ssh_job_info(job_id, self._address_host, ports[22], msg["ssh_key"])
```
GuillaumeDerval: @anthonygego shouldn't we directly update the "ssh_key" message code to be able to set another port?

inginious/agent/docker_agent/_docker_interface.py
```diff
@@ -29,19 +30,28 @@ def get_containers(self):
...
+        images = {}
+        for x in self._docker.images.list(filters={"label": "org.inginious.grading.name"}):
+            try:
+                title = x.labels["org.inginious.grading.name"]
+                created = datetime.strptime(x.attrs['Created'][:-4], "%Y-%m-%dT%H:%M:%S.%f").timestamp()
+                ports = [int(y) for y in x.labels["org.inginious.grading.ports"].split(
+                    ",")] if "org.inginious.grading.ports" in x.labels else []
+                images[x.attrs['Id']] = {"title": title, "created": created, "ports": ports}
+            except:
+                logging.getLogger("inginious.agent").warning("Container " + title + " requested an invalid port")
```
GuillaumeDerval: Actually, the error can come from various things. Can you use getLogger.exception with "Container %s is badly formatted" ?

inginious/agent/docker_agent/_docker_interface.py
```diff
@@ -61,16 +71,15 @@ def get_host_ip(self, env_with_dig='ingi/inginious-c-default'):
             return None
 
     def create_container(self, environment, network_grading, mem_limit, task_path, sockets_path,
-                         course_common_path, course_common_student_path, ssh_port=None):
+                         course_common_path, course_common_student_path, ports={}):
```
GuillaumeDerval: This is a bad smell in Python. Set `ports=None` and then in the code below add the lines `if ports is None: ports = {}`. 

If you don't see why it's bad to do it like that, check in the python doc ;-)

GuillaumeDerval:
LGTM

inginious/agent/docker_agent/__init__.py
anthonygego: That is not critical but could be nice.

inginious/agent/mcq_agent/__init__.py
```diff
@@ -75,7 +75,10 @@ def environments(self):
             text.append("\n\n" + _("Among them, you have {} invalid answers in the multiple choice questions").format(mcq_error_count))
 
         nb_subproblems = len(task.get_problems())
-        grade = 100.0 * float(nb_subproblems - error_count) / float(nb_subproblems)
+        if nb_subproblems == 0:
+            grade = 100.0
```
anthonygego: If I understand well, the MCQ agent will return a score of 100 if there are no problem added. Wouldn't this be rather considered as an error ?

PR #354, opened 2018-09-24
Fixes contains_is_html method for non-string keys

mpiraux:
A YAML task description can also contains integer keys, which will be parsed as such by the YAML parser. The fix prevents an error from being raised and returns the correct value.


anthonygego:
@mpiraux Please sign the CLA.

PR #361, opened 2018-10-04
Make the webdav a standalone app

anthonygego:
Quick changes to allow clients to ask OPTIONS/PROPFIND on ``/``.

``inginious-webdav`` is quite similar to ``inginious-webapp``. Maybe they can even be merged actually.

``webdav_available`` is replaced by the use of ``webdav_host`` set in ``configuration.yaml``.

### DIFF
--- a/doc/commands.rst
+++ b/doc/commands.rst
@@ -9,6 +9,7 @@ Frontend commands
 
     commands_doc/inginious-install
     commands_doc/inginious-webapp
+    commands_doc/inginious-webdav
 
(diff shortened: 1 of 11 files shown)


GuillaumeDerval:
LGTM, but could you update the documentation with everything that is needed for running inginious-webdav with nginx (https://inginious.readthedocs.io/en/v0.5/install_doc/installation.html)?

anthonygego:
Done.

PR #364, opened 2018-10-16
bootstrap 4

anthonygego:
Should be OK but need to be tested. As plugins templates should be updated too, I suggest to keep old plugins templates into v0.5 branches for a while in order to ease the plugin install for v0.5 users.


anthonygego:
Friendly reminder :-)

anthonygego:
@GuillaumeDerval

PR #376, opened 2018-11-29
Fix some issues saving profile

crdgonzalezca:
This PR fixes some issues/bugs while saving the profile, here are the problems fixed:

1. When you register using _social authentication_ and then try to add a custom password, the application says your profile was saved correctly with the new password but it wasn't updated actually.
This is a problem when you want to update your bindings, as you cannot remove them because you need a password but you cannot update the password. 
**The fix proposed for this is to check the length of new password rather than the old password so that way we know when the user wants to update the password.**

2. If the name is updated, the current session is not updated with the new name. Also, it doesn't check if the new name is empty.

3. If you update all fields at the same time (password, name and language), either the password or language is updated but not both at the same time. Now, it updates those fields at the same time and are not dependent on each other.

Also, some refactoring on `save_profile` was done.

### DIFF
--- a/inginious/frontend/pages/preferences/profile.py
+++ b/inginious/frontend/pages/preferences/profile.py
@@ -19,35 +19,9 @@ class ProfilePage(INGIniousAuthPage):
         """ Save user profile modifications """
         result = userdata
         error = False
-        msg = ""
 
-        # Check input format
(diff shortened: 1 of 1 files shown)


anthonygego:
LGTM. Thanks !

PR #377, opened 2018-11-30
Enhanced mcq

ogoletti:
We just tested this PR with master and it seems to work. 
Could you check if it could be merged ?

### DIFF
--- a/inginious/frontend/submission_manager.py
+++ b/inginious/frontend/submission_manager.py
@@ -382,7 +382,7 @@ class WebAppSubmissionManager:
             submission["input"] = inp
             return submission
 
-    def get_feedback_from_submission(self, submission, only_feedback=False, show_everything=False,
+    def get_feedback_from_submission(self, task, submission, only_feedback=False, show_everything=False,
                                      translation=gettext.NullTranslations()):
         """
             Get the input of a submission. If only_input is False, returns the full submissions with a dictionnary object at the key "input".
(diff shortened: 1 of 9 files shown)


GuillaumeDerval:
@mpiraux can you check if it does not break your code?

inginious/frontend/task_problems.py
```diff
@@ -49,6 +49,11 @@ def show_editbox(cls, template_helper, key):
     def show_editbox_templates(cls, template_helper, key):
         return ""
 
+    @classmethod
+    @abstractmethod
+    def prepare_feedback(cls, feedback, show_everything, translation):
```
GuillaumeDerval: A small doc here indicating what is the purpose of this method would be nice

inginious/frontend/submission_manager.py
```diff
@@ -364,15 +364,21 @@ def get_feedback_from_submission(self, submission, only_feedback=False, show_eve
         if "text" in submission:
             submission["text"] = ParsableText(submission["text"], submission["response_type"], show_everything, translation).parse()
         if "problems" in submission:
-            for problem in submission["problems"]:
-                if isinstance(submission["problems"][problem], str):  # fallback for old-style submissions
-                    submission["problems"][problem] = (submission.get('result', 'crash'), ParsableText(submission["problems"][problem],
-                                                                                                       submission["response_type"],
```
GuillaumeDerval: It feels like we are silently dropping support for "response_type" in subproblem-specific feedback here. Is it expected?

anthonygego: The feature is deprecated since v0.4 but they are still people using it through the undocumented YAML fields. So yes maybe this should be kept for now.

Drumor:
Dunno where we are with this PR but at least it's rebase.

anthonygego:
I'll close this as the code is completely obsolete now.

PR #378, opened 2018-11-30
Show a soft deadline for tasks

qdeconinck:
Sometimes, students might be tempted to perform homeworks at the very
last moment, which can be too late. This commit implements a soft
deadline mechanism that is publicly communicated, and keep the previous
deadline as hard deadline, which is kept private but determines until
when the task is available.


GuillaumeDerval:
LGTM. A nice addition would be to send the information about the soft deadline being reached to the grader, but that's for another PR.

I'll wait for @anthonygego 's comments, then we can merge this.

anthonygego:
Another nice addition would be to ease the course administrators to see which submissions were late.

PR #379, opened 2018-12-02
Add a new stats page in the administration

GuillaumeDerval:


inginious/frontend/templates/course_admin/stats.html
```diff
@@ -0,0 +1,147 @@
...
+        </div>
+    </div>
+    <button type="submit" class="btn btn-primary">Update</button>
+</form>
+
+<h2>$:_("Stats between {} and {}").format(daterange[0].strftime("%d/%m/%Y %H:%M:%S"), daterange[1].strftime("%d/%m/%Y %H:%M:%S"))</h2>
+
+<h3>$:_("Submissions")</h3>
+<canvas id="canvas"></canvas>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/Chart.js/2.7.3/Chart.min.js"></script>
```
anthonygego: We were used to provide a copy of each library, so if the licence allows to redistribute it, maybe it would be more coherent to do so.

anthonygego:
Oops, I didn't see this was updated.

PR #381, opened 2018-12-07
as bootsrap requires popper, its better to get popper before bootstrap

Drumor:
Should resolve Js error on submission.


anthonygego:
Could you therefore regenerate the minified js ?

PR #382, opened 2018-12-11
Fix bug adding student to a course but username is empty

crdgonzalezca:
When you try to add a student to a course but if the username in the input is empty, an exception like this is shown:
image

The this happens when you are in **course administration** in **students** tab, the you try adding a student but the inserted username is empty.


anthonygego:
LGTM. Thanks !

PR #386, opened 2019-01-31
Autotest Script

Maxmawt:
Non final version of autotesting script for code checking

### DIFF
--- a/inginious-autotest
+++ b/inginious-autotest
@@ -1,0 +1,304 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# This file is part of INGInious. See the LICENSE and the COPYRIGHTS files for
+# more information about the licensing of this file.
+#
+
+"""Test the output of a given input from a yaml inginious output file and compare with the old one"""
+
+import argparse
+import os
+import time
+import json
+import sys
+
+from inginious.common.tags import Tag
+from inginious.frontend.tasks import WebAppTask
+from inginious.frontend.parsable_text import ParsableText
+from inginious.common.log import init_logging
+from inginious.common.course_factory import create_factories
+from inginious.client.client_sync import ClientSync
+from inginious.frontend.arch_helper import start_asyncio_and_zmq, create_arch
+from yaml import load
+from inginious.common.filesystems.local import LocalFSProvider
+from inginious.frontend.task_problems import *
+
+
+def import_class(name):
+    m = name.split('.')
+    mod = __import__(m[0])
+
+    for comp in m[1:]:
+        mod = getattr(mod, comp)
+    return mod
+
+
+def create_client(config, course_factory, fs_provider):
+    """
+    Create a new client and return it
+    :param config: dict for configuration
+    :return: a Client object
+    """
+    zmq_context, t = start_asyncio_and_zmq(config.get("debug_asyncio", False))
+    return create_arch(config, fs_provider, zmq_context, course_factory)
+
+
+def compare_all_outputs(output1, output2, keys):
+    """
+    Compare both the output according to each key strategy
+    :param output1: output from the yaml
+    :param output2: output from the new job
+    :param keys: list of string, representing the keys for each value in the output
+    :return: a dict, whose the form is
+        {k: t,...} where k is a key whose the associated values in the outputs are different and t is a tuple containing
+        the values from both the outputs
+    """
+    return {
+            keys[i]: (output1[i], output2[i]) for i in range(len(keys))
+            if compare_output(output1[i], output2[i], keys[i]) and output1[i] is not None and output2[i] is not None
+        }
+
+
+def compare_output(output1, output2, key):
+    """
+    Compare atomic elements from the outputs
+    :param output1: atomic data structure from the yaml
+    :param output2: atomic data structure from the new job
+    :param key: key associated
+    :return: True if output1 and output2 are different, False otherwise
+    """
+
+    def result_compare(output1, output2):
+        """
+        Compare for the result key
+        :param output1: string
+        :param output2: tuple (string 1, string 2)
+        :return: True if output1 is different from the string 1 for output2
+        """
+        return output1 != output2[0]
+
+    def problems_compare(output1, output2):
+        """
+        Compare for the problems key
+        :param output1: dict whose each value is a list
+        :param output2: dict whose each value is a tuple
+        :return: True if output1 != output2, False otherwise
+        """
+        try:
+            for k in output1:  # same keys in output1 and output2
+                l = output1[k]  # in output1, values are lists of size 2
+                t = output2[k]  # in output2, values are tuples of size 2
+                if l[0] != t[0] or l[1] != t[1]:
+                    raise BaseException("")
+            return False
+        except:  # KeyError or BaseException => different outputs
+            return True
+
+    def archive_compare(output1, output2):
+        """
+        Return False because archives are always different
+        """
+        return False
+
+    def generic_compare(output1, output2):
+        """
+        Generic compare for keys
+        :param output1: string or different object (dict, list, ...)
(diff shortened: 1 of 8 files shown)


inginious-autotest
```diff
@@ -0,0 +1,255 @@
...
+    """
+    Test the task by comparing the new outputs with the old ones
+    :param yaml_data: dict corresponding to the yaml output file for the task
+    :param task: Task object corresponding to the yaml data
+    :param client: backend client of type Client
+    :param client_sync: ClientSync object
+    :return: dict whose the format is specified in compare_all_outputs function doc
+    """
+    if task.get_environment() not in client.get_available_containers():
+        time.sleep(10)
```
GuillaumeDerval: Let's try with 1 rather than 10 ;-)

inginious-autotest
```diff
@@ -0,0 +1,255 @@
...
+
+def get_task(course_id, task_id, course_factory):
+    """
+    Create a Task object from the data contained in the yaml data and config
+    :param course_id: string for the id of the course
+    :param task_id: string for the task id
+    :param course_factory: CourseFactory object
+    :return: Task object corresponding to the task output for the yaml
+    """
+    return course_factory.get_task(course_id, task_id)
```
GuillaumeDerval: Not sure this function is needed

inginious-autotest
```diff
@@ -0,0 +1,255 @@
...
+            yaml_data,
+            task.get_fs(),
+            task.get_translation_fs(),
+            task.get_hook(),
+            config["default_problem_types"],
+            debug=True
+        )
+        web_task.get_context('English')
+    except BaseException as e:
+        res = yaml_data
```
GuillaumeDerval: This cannot work

inginious-autotest
```diff
@@ -0,0 +1,255 @@
...
+        if res != {}:
+            output[path] = res
+
+
+def test_task_yaml(path, output, course_factory, task_name, course_name, config):
+    res = {}
+    with open(path, 'r') as yaml_file:
+        yaml_data = load(yaml_file)
+    test_web_task(yaml_data, get_task(course_name, task_name, course_factory), res, config)
+    if res != {}:
```
GuillaumeDerval: res is not modified inside `test_web_task`, so this cannot work

anthonygego:
Don't forget to add some docs ;-)

inginious/frontend/parsable_text.py
```diff
@@ -122,7 +122,7 @@ def starttag(self, node, tagname, suffix='\n', empty=False, **attributes):
 class ParsableText(object):
     """Allow to parse a string with different parsers"""
 
-    def __init__(self, content, mode="rst", show_everything=False, translation=gettext.NullTranslations()):
+    def __init__(self, content, mode="rst", show_everything=False, translation=gettext.NullTranslations(), debug=False):
```
GuillaumeDerval: I would rather define debug inside `parse(debug=False)`, and call it manually.

inginious/frontend/tasks.py
```diff
@@ -17,7 +17,7 @@
 class WebAppTask(Task):
     """ A task that stores additional context information, specific to the web app """
 
-    def __init__(self, course, taskid, content, task_fs, translations_fs, hook_manager, task_problem_types):
+    def __init__(self, course, taskid, content, task_fs, translations_fs, hook_manager, task_problem_types, debug=False):
```
GuillaumeDerval: We should avoid defining `debug` here, it's not really needed nor relevant.

inginious-autotest
Maxmawt: removed

inginious-autotest
Maxmawt: changed

inginious-autotest
Maxmawt: changed too

inginious/frontend/parsable_text.py
Maxmawt: done

inginious/frontend/tasks.py
Maxmawt: done

inginious-autotest
Maxmawt: Done and it's working locally and on Travis

inginious-autotest
```diff
@@ -0,0 +1,281 @@
...
+        # Init logging
+        init_logging()
+
+    default_problem_types = {
+        problem_type.get_type(): problem_type for problem_type in [DisplayableCodeProblem,
+                                                                   DisplayableCodeSingleLineProblem,
+                                                                   DisplayableFileProblem,
+                                                                   DisplayableMultipleChoiceProblem,
+                                                                   DisplayableMatchProblem]
+    }
```
GuillaumeDerval: There should be a way to add new problem types. See the `--ptype` arg in inginious-agent-mcq.

inginious/common/tags.py
```diff
@@ -94,7 +94,13 @@ def create_tags_from_dict(tag_dict):
                 elif(type == 1):
                     tag_list_misconception.insert(int(tag), Tag(id, name, description, visible, 1))
                 else:
-                    tag_list_common.insert(int(tag), Tag(id, name, description, visible, 0))
+                    if not debug:
```
GuillaumeDerval: Duplicate code. replace with not-nested elif.

```python
elif not debug or type == 0:
    tag_list_common.insert(int(tag), Tag(id, name, description, visible, 0))
else:
    raise BaseException("Tag type not correct")
```

inginious/frontend/tasks.py
GuillaumeDerval: Not, it's not ;-)

Is it possible to check tags elsewhere?
Why not defining a class method `check_format` that can be called externally ?
It may be useful for other uses.

inginious/frontend/tasks.py
```diff
@@ -59,7 +59,7 @@ def __init__(self, course, taskid, content, task_fs, translations_fs, hook_manag
         self._regenerate_input_random = bool(self._data.get("regenerate_input_random", False))
 
         # Tags
-        self._tags = Tag.create_tags_from_dict(self._data.get("tags", {})) 
+        self._tags = Tag.create_tags_from_dict(self._data.get("tags", {}), debug=debug)
```
GuillaumeDerval: remove the debug here.

inginious/frontend/parsable_text.py
```diff
@@ -141,16 +141,19 @@ def original_content(self):
...
             try:
                 if self._mode == "html":
                     self._parsed = self.html(self._content, self._show_everything, self._translation)
                 else:
-                    self._parsed = self.rst(self._content, self._show_everything, self._translation)
+                    self._parsed = self.rst(self._content, self._show_everything, self._translation, debug=debug)
             except:
-                self._parsed = self._translation.gettext("<b>Parsing failed</b>: <pre>{}</pre>").format(html.escape(self._content))
+                if debug:
+                    raise BaseException("Parsing failed")
```
GuillaumeDerval:
```python
except Exception as e:
    if debug:
        raise BaseException("Parsing failed") from e
    ...
```

Maxmawt: done

Maxmawt:
* Add a little spelling correction in the doc
* Add robustness in case of diverging problems. For example, for a ``submission.test`` which is compiling and the test now returns a compilation failure

Maxmawt:
Add exception catching

Maxmawt:
Should be ok now

Maxmawt:
Change printing of failure and remove dev printings

Maxmawt:
@GuillaumeDerval , @anthonygego would like a new approval for the latest changes ;-)

anthonygego:
@GuillaumeDerval Did you validate the changes ?

GuillaumeDerval:
LGTM

PR #387, opened 2019-02-01
Record late submissions

qdeconinck:
This is done by adding a @late tag in the input of the submission.

This PR also shows the fact that submissions are late in the task submission list, but this requires adding its own field in the submission db. This is done by the second commit, that indicates it either at page load time (html) or at submission time (js + tasks.py files).


qdeconinck:
Closed by #388, as it does not duplicate fields in the DB.

PR #388, opened 2019-02-05
Record late submission v2

qdeconinck:
This is done by adding a `@late` tag in the input of the submission.

Compared to #387, this PR does not duplicate the `late` field to show that a submission is late or not on the task page. This is easy at submission time (js + submission_to_json), but when statically GETting the task html page, we need to fetch the tag from the input.


anthonygego:
LGTM

PR #391, opened 2019-02-26
Added lti_auto_bind feature

NerLOR:
New course setting: "lti_auto_bind"
Added function lti_auto_bind() in course
Added test in lti page, which checks if lti_auto_bind() of the course is True and if so, binds the lti-provided user to the course automatically

### DIFF
--- a/inginious/frontend/pages/lti.py
+++ b/inginious/frontend/pages/lti.py
@@ -112,9 +112,26 @@ class LTILoginPage(INGIniousPage):
         except:
             return self.template_helper.get_renderer().lti_bind(False, "", None, "Invalid LTI data")
 
-        user_profile = self.database.users.find_one({"ltibindings." + data["task"][0] + "." + data["consumer_key"]: data["username"]})
+        if course.lti_auto_bind():
+            lti_email = data['email']
+            lti_name = data['realname']
+            lti_id = lti_email[0:lti_email.find('@')]
+            user_profile = self.database.users.find_one({"username": lti_id})
+            if not user_profile:
+                # New user, create an account using email address
+                user_profile = {"username": lti_id,
+                                "realname": lti_name,
+                                "email": lti_email,
+                                "bindings": {data["consumer_key"]: [lti_id, {}]},
+                                "language": self.app.get_session().get("language", "en")}
+                self.database.users.insert(user_profile)
+        else:
+            user_profile = self.database.users.find_one({"ltibindings." + data["task"][0] + "." + data["consumer_key"]:
+                                                         data["username"]})
+
         if user_profile:
-            self.user_manager.connect_user(user_profile["username"], user_profile["realname"], user_profile["email"], user_profile["language"])
+            self.user_manager.connect_user(user_profile["username"], user_profile["realname"],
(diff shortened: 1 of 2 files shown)


rolfis:
This patch is a genius patch. I've added the option to the UI as well, is this possible to integrate somehow with this pull request or should I add a new one from this?

anthonygego:
I'll look at this as soon as I can. I'm still pretty busy on other projects for now.

inginious/frontend/pages/lti.py
```diff
@@ -112,9 +112,26 @@ def GET(self):
         except:
             return self.template_helper.get_renderer().lti_bind(False, "", None, "Invalid LTI data")
 
-        user_profile = self.database.users.find_one({"ltibindings." + data["task"][0] + "." + data["consumer_key"]: data["username"]})
+        if course.lti_auto_bind():
+            lti_email = data['email']
+            lti_name = data['realname']
+            lti_id = lti_email[0:lti_email.find('@')]
+            user_profile = self.database.users.find_one({"username": lti_id})
```
anthonygego: I'm afraid this will lead to a severe security issue enabling account hijacking. This may seem harmless in a closed environment such as a university Moodle instance where all students have the same email domain, but in case an LTI course is used in a public environment (such as our edX Louv1.x on inginious.org), ``somebody@somedomain`` may be bound to the ``somebody@anotherdomain`` account, or, actually, simply an alreay existing account with corresponding user id.

A more secure and acceptable auto-binding feature would be to silently accept the binding in place of the user **once logged in**.

NerLOR: Of course this is right and this feature can lead to serve security issues, but as you can see, this feature has to be manually activated in the ```course.yaml```.
Solutions regarding to the other points you made:
- whitelisting only one domain for this feature
- implement mapping of domains to some pre-/postfixes (```alice@gmail.com``` -> ```alice_gmail```, ```alice@company.com``` -> ```alice_company```). I know, admins will have to keep track over their mappings to ensure the uniqueness of the name.

What would you think, would be the best solution? If necessary, I will implement one of these features or something else, if you have better ideas.

GuillaumeDerval:
In general, INGInious is made to avoid trusting things made by exercise creators (i.e. the people that can modify the course.yaml). So it would be preferable to be implemented system-wide in the configuration.yaml.

Even with that, it would remain a security problem, as individual courses (managed by potentially unknown actors) would be able to hijack accounts at will. The attack surface is huge: you can have a malicious course creator, or a hacked course creator, or an hack of the webdav, or a malicious LTI provider, ..., ..., ... and it would lead to an open door to one of the most important collection of the database (that is, the user collection, and their data).

Sadly, we'll then not be able to integrate this PR in INGInious. Feel free, however, to transform it in a plugin, we'll gladly put a link to your plugin inside our INGInious-plugin repository.

AugustH:
would you accept a version like:

* keep the patch mostly, just set auto-binding always to false
* at the end of `WebAppCourse __init__`: add a hook ("course_created")
* add a plugin which uses this hook to modify the auto-binding

--> no change if the plugin isn't loaded.

PR #392, opened 2019-02-26
Improved LDAP Authentication

NerLOR:
added _ldap_request() in ldap_auth.py which tries one ldap authentication
POST function now uses the old configuration in course.yaml and calls _ldap_request() once
or calls _ldap_request() multiple times


anthonygego:
I'm closing this as #423 is going to be merged.

PR #396, opened 2019-04-08
Add IPython inside containers + Modify the way `run` files are chosen + refactor the API (run_student mainly + small fixes)

GuillaumeDerval:
- [x] Add IPython
- [x] Make it work
- [x] Import INGInious libs by default in the environment + some vars
- [x] Update documentation on the commands
- [x] Add documentation about IPython env
- [x] Modify `run_student` to make it easily accessible in Python
- [x] Deprecate `archive`. Make the folder /archive accessible directly instead.
- [x] Add a `run_cmd` option in task.yaml
- [x] Allow to use files such as `run.py` or `run.sh` as run script, avoiding the shebang and being a bit clearer, while maintaining retrocompatibility
- [x] Allow to put run script inside the common folder


GuillaumeDerval:
@anthonygego done :-)

base-containers/base/bin/INGInious
```diff
@@ -251,6 +254,19 @@ class INGIniousMainRunner(object):
         f.write(json.dumps(data))
         f.close()
 
+    def _detect_and_start_run(self):
+        """ By default, we run the `run` script under IPython. This function checks
```
anthonygego: I'd rather make this explicit to avoid confusion and to ensure retro-compatibility the right way. This patch may make support harder in the future as problems will leverage from the fact that some users will sometimes forget this rule, or will simply be distracted and won't look at that line particularly. Just think about a missing "!" in the shebang. In the previous behaviour, Linux would simply say that this is not executable and this will be fixed very quickly. If this is therefore executed by ipython, this will be harder to debug for a common user.

base-containers/base/bin/inginious-ipython
```diff
@@ -0,0 +1,24 @@
...
+sys.argv = sys.argv[1:]  # correct the args
+
+# Add the utility function to the global namespace
+shell.run_cell("from inginious_container_api.input import *", store_history=False)
+shell.run_cell("from inginious_container_api.feedback import *", store_history=False)
+shell.run_cell("from inginious_container_api.lang import *", store_history=False)
+shell.run_cell("from inginious_container_api.rst import *", store_history=False)
+shell.run_cell("from inginious_container_api.run_student import *", store_history=False)
+
+shell.safe_execfile_ipy(sys.argv[0])
```
anthonygego: I guess that this would prevent running a ipython script with several arguments from a Bash run file ?

base-containers/base/Dockerfile
```diff
@@ -5,12 +5,14 @@ ENV LANG en_US.UTF-8
 ENV LANGUAGE en_US:en  
 ENV LC_ALL en_US.UTF-8
 
-# Install python, needed for scripts used in INGInious
+# Install python, needed for scripts used in INGInious + locale support
 RUN     yum clean metadata && \
         yum -y install https://centos7.iuscommunity.org/ius-release.rpm && \
         yum -y upgrade && \
         yum -y install python python35u-pip python35u zip unzip tar sed openssh-server openssl bind-utils iproute file && \
```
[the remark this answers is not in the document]
GuillaumeDerval: This is orthogonal, but yes we should :-)

base-containers/base/bin/inginious-ipython
GuillaumeDerval: It doesn't. Line 15 correct the args, such that the script see the correct ones. For example, if you run `/bin/inginious-ipython myscript.py something somethingelse`, then `myscript.py` will see the following as `sys.argv`:
`['myscript.py', 'something', 'somethingelse']`

anthonygego:
LGTM.

PR #397, opened 2019-04-11
Plugin to receive a notification on task completion

sprt:
I started working on a new plugin that would allow students to receive a notification when a task they've submitted has finished. As a student, when I work on a longer-running task, I find myself having to check the INGInious tab multiple times to see if my task has finished. I also often end up coming back to the task well after it has finished, and losing precious time, hence why I think such a feature has great value.

This uses the [Notifications API](https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API), which is [supported by 75% of browsers](https://caniuse.com/#feat=notifications). It's clearly an MVP for the moment and just a proposal to see if you'd be interested by this plugin.

# How it works

There's a single setting to enable/disable notifications (stored on the client with [`localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API)). For this MVP, I used the `main_menu` hook for the interface and the result looks like the screenshot below. The first time notifications are enabled, we trigger the native browser dialog requesting notification permission. After that, the user can toggle them as they please.

Current interface to toggle notifications
*Current interface to toggle notifications.*

One disadvantage with this interface is that, when a user is on a task page, they would have to navigate to the course list if they wanted to toggle notifications. For this reason, I think it's better to put that checkbox in the navbar, or in the dropdown of that navbar. But this would probably require a new hook. **What do you think about this?**

Regarding the actual notifications, one is sent each time a task that the user just submitted (ie. *not* reloaded) has finished, whether successfully or not. The screenshot below is an example of such a notification. An important feature to add would be not to receive a notification if you're currently focused on the tab, as INGInious already provides visual feedback.

Example notification
*Example notification. (edit April 28: now includes INGInious' logo)*

# Implementation details

To make this work, I had to change the behavior of the `feedback_script` hook to make it also trigger for new submissions, as opposed to just submissions that are being reloaded. I also added a new parameter to the hook, `reloading`, which is a boolean that is true iff the feedback regards a reloaded submission.

# Todo

Before this can be merged into master, at least the following tasks should be completed:

- [ ] Investigate the best UI for the setting toggle
- [ ] Update the documentation
- [x] Don't send a notification if the tab is focused
- [x] Focus the tab when a notification is clicked
- [x] Warn the user if their browser doesn't support notifications
- [x] Warn the user if they try to enable notifications but they're blocked at the browser level
- [x] Make the checkbox checked on page load if notifications are enabled
- [x] Include the INGInious logo in the notification
- [x] When the user enables notifications for the first time, send them a notification explaining how the feature works

... and anything else we can think about.

---

Waiting for any feedback. Should I go forward with this?


anthonygego:
Of course this is interesting. I think there's a hook for each menu the student can see (main_menu, course_menu, task_menu) but you can add a hook for the navbar and see if it's better that way.

sprt:
I added a hook for the navbar and put the toggle there. Looks much more user-friendly as you don't have to navigate to another page to toggle it now. Here's what it looks like:

Navbar toggle

sprt:
I've implemented everything I had in mind for this feature. If you're happy with the UI and the way the new navbar hook is implemented, I'll add some documentation and mark the PR as ready for review.

GuillaumeDerval:
This is really a nice feature, and the implementation is very small. 

For once I think this should not even be a plugin, as this is something that we might want to propose to other plugins as well.

We actually have an idea of plugin with @pschaus that may use this kind of feature. We have for a long-time wanted to do a peer-review plugin, and this of things can be super-useful in this context.

It lacks some things IMHO:
- This only works if we are on the page of the submission when the submission is done. With a kind of active (but slow) recurrent querying (let's say, once each minute + each time we refresh for a submission), we should be able to gather new messages from the frontend and to display them on any webpage.
- It lacks a nice python interface to create message from everywhere.

An interface like this would be very nice:
```python
import inginious.frontend.notification as notification

# Add a notification that will be sent, one day, to the user. Stores it temporarily in DB. Keep it in DB for max 3600 seconds
notification.notify(userid, message, timeout=3600)

# Get all the notifications that the user need to see, and delete them for DB
notifications = notification.get_notifications_for_user(userid)
```

Let's say, that, for now, we don't mind if we loose notifications. This simple design can easily be extended if in the future we want to use also the [push API](https://developer.mozilla.org/en-US/docs/Web/API/Push_API) (`notification.notify(userid, message, push=True)`).

We should not let this die. I don't know if you are still motivated to contribute on this @sprt, but if not I'll definitely try to spend some time on it in the future (although it's sometime difficult to find time ;-) )

sprt:
I probably won't have time to work on this new API any time soon since I'm starting a 12-week internship next week and will be working on my thesis after that, unfortunately.

anthonygego:
I'm closing this since it is no more actively developed since almost 2 years. Feel free to open it again some day.

PR #398, opened 2019-04-17
Redirect from logout to courselist

amicis31:
Just a little tweak for preventing bad redirection on logout

Instead redirect to /courselist.


anthonygego:
LGTM

PR #400, opened 2019-05-27
Drag and drop files into file list (admin task)

Drumor:
This lets course administrators to drag and drop files into file list.


anthonygego:
LGTM.

PR #401, opened 2019-05-27
added @attempts value to inputdata

pickettbd:
Hi! I've come across a situation where it would be nice to know which attempt is currently being graded. Of course, the database can't be accessed by the grading container (and appropriately so), so I need a way to get the value. The `@username` and `@lang` keys in the `inputdata` dictionary are very helpful. I thought I could solve my own problem with `@attempts`. I've implemented this for our course by changing the source code, but it is an older version. I have tested it and found that it works well. I thought I'd future-proof things by adding the change here so I can upgrade without a problem. I have not actually tried setting up a completely new system with this new code.

My changes are very simple. I hope you'll review them and consider including them (or something similar of your own). Thank you!

### DIFF
--- a/inginious/frontend/submission_manager.py
+++ b/inginious/frontend/submission_manager.py
@@ -176,6 +176,9 @@ class WebAppSubmissionManager:
             username = self._user_manager.session_username()
             submission["username"] = [username]
             submission["submitted_on"] = datetime.now()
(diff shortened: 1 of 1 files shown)


anthonygego:
LGTM. There is probably some other data that are worth being added in those additional fields, such as the deadline, which could make our custom ``hidden-util`` RST directive more convenient. Thanks.

PR #402, opened 2019-06-12
Tagging system rewrite to improve global performance

anthonygego:
In order to improve the memory footprint and to reduce runtime computation on tags, the tagging system has been refactored. This PR requires the ``course.yaml`` and ``task.yaml`` files to be modified in order to reflect the new arch.

Before
----------

The ``task.yaml`` file contained an ordered list of tags.

    tags:
      '0':
          visible: true
          type: 0
          description: This tag validate that your close are correctly performed.
          name: Close
          id: close
      '1':
          description: This tag validate that you are able to copy permission of a file.
          type: 0
          visible: true
          id: permission
          name: Permission
      ...

The course tags were computed at runtime by merging all the task tags together. Tasks were only allowed to use the task tags, and one needed to import tags from one task to another. Category tags were special because they did not have ids.

After
------

Tags are now a course wide feature, and can be used to categorize tasks.

The ``course.yaml`` file contains a dict where key is the tag id and the value the actual tag dict.

    tags:
      banned_funcs:
          type: 1
          name: Banned functions
          description: You use some banned functions.
          visible: true
      close:
          name: Close
          type: 0
          visible: true
          description: This tag validate that your close are correctly performed.
      ...

The ``task.yaml`` file contains a ``categories`` field which list the tags :

    categories:
    - close
    - copy
    - open
    - s5
    - file
    - permission

This still allows to use of auto-tags as they were independent. It also allows to use any tag as a categorizer instead of only "category tags".


inginious/frontend/templates/course_admin/tags.html
```diff
@@ -0,0 +1,94 @@
...
+        </tr>
+    </thead>
+    <tbody>
+
+    <!-- An empty row so that we can copy it to add new rows. All 'disabled' are important to not send this row to INGInious when saving the task-->
+    <tr id="NEW" style="display:none;">
+        <td><input disabled type="text" ID_STOP id="id_NEW" class="form-control" name="tags[NEW][id]" value="ID_REPLACE"></td>
+        <td><input disabled type="text" class="form-control" name="tags[NEW][name]" value="NAME_REPLACE"/></td>
+        <td><textarea disabled class="form-control" onfocus="studio_expand_tag_description(this);" onblur="studio_expand_tag_description_not(this);" name="tags[NEW][description]" rows="1" style="resize:none">DESCRIPTION_REPLACE</textarea></td>
+        <td><input disabled type="checkbox" class="form-control" name="tags[NEW][visible]" VISIBLE_REPLACE)/></td>
```
Drumor: Design remark : I think it's better to remove the form-control class here. I have a huge square:
image

inginious/frontend/templates/course_admin/edit_tabs/basic.html
```diff
@@ -48,6 +48,13 @@
         <input type="text" class="form-control" id="author" name="author" placeholder="$:_('Your name')" value="$a"/>
     </div>
 </div>
+<div class="form-group row">
```
Drumor: Probably a good idea to propose a list of available id's.

PR #417, opened 2019-07-30
Attempt to fix a Docutils warning.

jy95:
"No MathJax URL specified, using local fallback"
This causes the debug mode of the ParsableText class to fail.


jy95:
Example : 

```rst

    Reminder : Fibonacci's sequence is computed as follows

    .. math::

         f(n) =\begin{cases}0 & n=0 \\1 & n = 1\\ f(n-1) + f(n-2) & n > 1\end{cases}
```

inginious/frontend/parsable_text.py
```diff
@@ -266,7 +266,7 @@ def rst(cls, string, show_everything=False, translation=gettext.NullTranslations
             'translation': translation,
             'raw_enabled': True,
             'file_insertion_enabled': False,
-            'math_output': 'MathJax'
+            'math_output': 'MathJax /this/does/not/need/to/exist.js'
```
anthonygego: Did you investigate to know what would be a real fix ?

jy95: As you can read from the [documentation](http://docutils.sourceforge.net/docs/user/config.html#math-output) : 
**"A URL pointing to a MathJax library should be appended after whitespace. A warning is given if this is missing."**
Feel free to use one of the three given solutions inside the documentation.

anthonygego: @GuillaumeDerval confirmed to me that this link is actually not used further. So I think it's OK for me.

PR #420, opened 2019-08-05
Use cards instead of alerts

### LINKED ISSUE #419: Use cards rather than alerts for submission with feedback
I just drop this here, waiting for approval, before actually coding it.

I find that the current situation (using alerts for displaying feedback) is making the feedback difficult to read. See this example:

Capture d’écran 2019-08-02 à 17 25 21

I would propose that instead of using a bootstrap "alert", we use instead a bootstrap card. Here is a screenshot of what it would look like:

Capture d’écran 2019-08-02 à 17 26 25

Here is the code used to produce this:
```html
<div class="card border-danger">
    <div class="card-header bg-danger text-white">
        There are some errors in your answer. [...]
        <button type="button" class="close" data-dismiss="alert" style="color: white;"><span aria-hidden="true">×</span><span class="sr-only">Close</span></button>
    </div>
    <div class="card-body">
        <!-- table here -->
     </div>
</div>
```

### Demo success
Capture d’écran 2019-08-02 à 17 30 17

### Demo timeout/other warning
Capture d’écran 2019-08-02 à 17 30 51

GuillaumeDerval:
Fixes #419

Capture d’écran 2019-08-05 à 15 23 56


inginious/frontend/static/css/INGInious.css
```diff
@@ -270,7 +285,3 @@ ul.checked-list-box input[type="checkbox"] {
   -webkit-box-shadow: inset 0 1px 1px rgba(0, 0, 0, 0.075), 0 0 8px rgba(102, 175, 233, 0.6);
   box-shadow: inset 0 1px 1px rgba(0, 0, 0, 0.075), 0 0 8px rgba(102, 175, 233, 0.6);
 }
-.dragin{
```
Drumor: why removing this?

GuillaumeDerval: It's what happens when you edit directly a generated file @Drumor 
https://github.com/UCL-INGI/INGInious/commit/59b490f3ea59c799ea72a88936fc2e1c64da0c24

The CSS files are generated from the LESS ones. You should not directly update the CSS, but rather the LESS file and run lessc to regenerate the css.

I'll add another commit to fix the bug ;-)

PR #423, opened 2019-08-29
Ldap auth

AugustH:
Improved LDAP Authentication v2

- extended config
- allow multiple configs
- documentation

### DIFF
--- a/inginious/frontend/plugins/auth/ldap_auth.py
+++ b/inginious/frontend/plugins/auth/ldap_auth.py
@@ -61,7 +61,8 @@ class LDAPAuthenticationPage(AuthenticationPage):
     def GET(self, id):
         settings = self.user_manager.get_auth_method(id).get_settings()
         return self.template_helper.get_custom_renderer('frontend/plugins/auth').custom_auth_form(settings,
-                                                                                                         False)
+                                                                                                  False)
+
     @staticmethod
     def _ldap_request(login, password, host, port=None, encryption=None, bind_dn=None, bind_password=None,
                       bind_password_file=None, base_dn=None, request=None, start_tls='NONE', cn='cn', mail='mail',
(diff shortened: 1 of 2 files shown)


inginious/frontend/plugins/auth/ldap_auth.py
```diff
@@ -61,7 +61,100 @@ class LDAPAuthenticationPage(AuthenticationPage):
...
+            raise Exception('Unable to contact host')
+        except Exception as e:
+            logger.error('Unable to get user data: {}'.format(str(e)))
+            conn.unbind()
+            raise Exception('Unknown User')
+
+        # 'Log in' (bind) to the ldap connection
+        if conn.rebind(user_data['dn'], password=password):
+            try:
+                email = user_data['attributes'][mail]
```
anthonygego: Here we used to take the first item of the returned array. Are the attributes not returned as an array in your case ?

inginious/frontend/plugins/auth/ldap_auth.py
```diff
@@ -61,7 +61,100 @@ class LDAPAuthenticationPage(AuthenticationPage):
...
+                username = login
+                realname = user_data['attributes'][cn]
```
anthonygego: Same comment here.

AugustH: i see - but i just rechecked against our ldap (windows server): both cn and userPrincipalName (which is used for mail) are strings, not lists. Sadly i can't check the original code as our ldap server needs TLS_BEFORE_BIND. 

Two ways to resolve:
- add [0], which works even in my case, gives bad realnames and mail
- additional `if isinstance` check

Any chance that you/anybody checks the old code: are there really lists or is only the first character taken?

anthonygego: I'll check this soon. AFAIK both the LDAP directories we used returned lists. But there may be differences between openLDAP and Active Directory.

anthonygego:
Superseeded by #554

PR #429, opened 2019-09-04
Change input field to register for a course as 'password' type

crdgonzalezca:
This changes the input field to register for a course as a 'password' type rather than a plain text type. This gives a better security to the course password.


anthonygego:
Thanks !

PR #446, opened 2019-10-09
Update from Weblate

INGIbot:


anthonygego:
Merging current translation state to avoid further conflicts.

Issue #434, opened 2019-09-20
Allow to reset submission state


anthonygego:
It should be very useful to be able to reset a student ``user_task.state`` field as a course administrator.

Drumor:
Solved in #449

PR #449, opened 2019-10-11
Updating state of user tasks

Drumor:
It's more than probably not mergeable as it is but I'd like to know if I understood well the #434 issue? Is it possible to get extra information about this feature request? Thx


anthonygego:
I'd rather have this in a POST method with a confirmation via a modal as for other critical features such as removing a student from the course, to avoid any wrong manipulation. I don't think it is neccesary to put the POST processing in another page, the routing dictionary is already long enough. You can just check for the form submission in the concerned page and call the appropriate function at that time. This will avoid to play with redirects too.

Drumor:
I think i have made changes you asked for.

Drumor:
I need to review if all is okay after rebase of master and also change page situation of button

Drumor:
I think i've done here.

PR #452, opened 2019-10-14
Cut long chains in task debug information

jadinm:
When loading long chains of debug data, the browser load is increasing
drastically. This is either due to an excessive amount of prints or
to the binary value of uploaded files that are printed out in the debug
dialog. This commit cuts all the chains that exceeds 1000 characters
(if this is a string) or bytes (if this is a byte chain).


GuillaumeDerval:
Cutting things in the debug view seems "ok". If more information is needed, the best things to do is either to download the submission and analyse the `submission.test` file & the other files in the archive or to start a SSH debug session.

PR #473, opened 2019-11-20
removing unecessary code

Drumor:
As https://github.com/webpy/webpy/pull/419 seems to be resolved and was packaged into pip, I think we can remove fix_webpy_cookies


anthonygego:
LGTM

PR #474, opened 2019-11-21
Add support for Jinja templates

anthonygego:
This adds support for jinja templates in the template manager as an opt-in in order to slowly migrate from web.py templates to jinja templates.

The renderer object is no more cached as the jinja initialization does not keep the globals dict as a reference. This should not have a significant impact compared to the whole page generation process.

While the web.py template renderer takes its arguments as a list, the jinja ones takes its arguments as a dictionary.


GuillaumeDerval:
LGTM

mpiraux:
There could be an opportunity for a test tool that validates the HTML output. I have no experience in that matter but it would be worth to have some kind of automatic validation of the produced HTML. Maybe a 1:1 translation is not mandatory ?

anthonygego:
Ok for me

PR #476, opened 2019-11-26
Environment parameters

GuillaumeDerval:
Add a mechanism to have different "environment parameters" for different "environment types".

An agent announces multiple environment, each having an environment type, which can be different.
The frontend then displays in the studio parameters specific for this particular environment type.

Most of the time I expect that one agent = one environment type, but that is not mandatory. 

The original problem was that MCQ tasks do not need a time limit, and that we had in the past wanted to add specific parameters for the MCQ to modify the behavior of the agent.

This PR modifies the task.yaml format. All the docker-related parameters are now environment parameters. All environment parameters are stored inside a key `environment_parameters`.

This implies that we have to modify every task.yaml for all the existing tasks. We have the choice of either doing in on-the-fly in INGIninous (currently done in the PR: https://github.com/UCL-INGI/INGInious/compare/env_parameters?expand=1#diff-7225411b5e409292ee4524fe95f36102R163)
or to do a script to update everything.


inginious/common/tasks.py
```diff
@@ -181,3 +159,14 @@ def _create_task_problem(self, problemid, problem_content, task_problem_types):
...
+
+    def _migrate_from_v1(self, content):
+        if "environment" in content:
+            content["environment_id"] = content["environment"]
+            content["environment_type"] = "docker" if content["environment_id"] != "mcq" else "mcq"
+            del content["environment"]
+            content["environment_parameters"] = {"limits": content.get("limits", {}),
+                                                 "run_cmd": content.get("run_cmd", ''),
+                                                 "network_grading": content.get("network_grading", False),
+                                                 "response_id_html": content.get('responseIsHTML', False)}
```
anthonygego: response_i*d*_html ?

inginious/frontend/environment_types/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+    """ env_obj is an instance of FrontendEnvType """
+    __env_types[env_obj.id] = env_obj
+
+
+def register_base_env_types(hook_manager: HookManager):
+    hook_manager.add_hook("register_env_type", register_env_type, 0)
+
+    # register standard env types here
+    register_env_type(DockerEnvType())
+    register_env_type(MCQEnvType())
```
anthonygego: Does this mean the environnement are loaded in the frontend even if the agents are not launched ?

GuillaumeDerval: Yes they are. For now this seems like a good middleground. In the long term, we may want that agents send themselves what they want as arguments, however this is complex to handle. Particularly when the frontend is starting or when an agent is not available. In that case, how would we be able to edit tasks, if we cannot get the parameters needed?

This architecture solves this problem, but forces the agent and the plugin "describing" the agent to be at the same version at all times.

PR #481, opened 2019-12-01
Previewable pages

GuillaumeDerval:
Add a mechanism to preview auth pages (i.e. pages that globally behave as auth pages, but are previewable under some circumstances).

Now the Course page asks for a login if the course is not previewable, rather than saying "you're not registered".
The same thing goes for the Task page.


GuillaumeDerval:
Please prefer PR #482 ;-)

PR #482, opened 2019-12-01
Course enrollment

GuillaumeDerval:
Create a dedicated webpage for course enrollment. It is displayed instead of the page "course unavailable".

Adds a mechanism to preview auth pages (such as the course/task page of a previewable course) that correctly handles non-connected users when the preview mode is not enabled.

Capture d’écran 2019-12-01 à 21 47 33

Capture d’écran 2019-12-01 à 21 46 39


anthonygego:
LGTM.

PR #487, opened 2019-12-15
translating non-english word

ThaSami:
just a fast translation..


Drumor:
Hello,

Thx for your contribution. We clearly appreciate that! :-)

Could you please use the translator syntax ( $:_('"Mytext"') ) to let the parser detect your string?

Thx a lot!

PR #494, opened 2020-01-14
Add a workaround for a starvation problem

GuillaumeDerval:


GuillaumeDerval:
This actually solves the starvation problem, but causes high cpu usage when the queue is filled with a lot of jobs (during real-life tests, the backend was not responsive enough and triggered false disconnection when under >30k tasks, which, while being unusual, happens).

We need to find something better.

PR #501, opened 2020-01-24
Fix tests

Drumor:
Let's tests run again...


inginious/__init__.py
```diff
@@ -5,10 +5,15 @@
 #
 
 import os
+import gettext
+import builtins
 
 __version__ = "0.6.dev0"
 
 
+builtins.__dict__['_'] = gettext.gettext
```
GuillaumeDerval: @anthonygego is that ok for you? For now we use _ as a builtin everywhere, but it needs to be defined inside each script that use a part of the inginious library. This seems to be a nice fallback.

PR #506, opened 2020-01-31
Fix LTI relative urls

mpiraux:
Using links or images in the description of tasks and serving them over LTI doesn't work. This patch changes the RST HTML writer so that it rewrites relative URLs to use an intermediate page that will redirect the request to the correct asset URL using an HTTP 301.


anthonygego:
LGTM

PR #511, opened 2020-02-12
fixing code language place holder

ThaSami:
the place holder says C++ 
where c++ return invalid language. however cpp works.


GuillaumeDerval:
LGTM :-)
Thanks!

PR #518, opened 2020-03-24
Add a way to kill job from the job queue

Drumor:
It allows administrators to manage and review jobs and agents.


anthonygego:
I'll make further tests but LGTM

PR #519, opened 2020-03-25
If usernames is None, data from all users will be returned. Its a lie!

Drumor:
I try to set usernames as None like it's specify :

> If usernames is None, data from all users will be returned.

but this goes in an error.


anthonygego:
Those lines should be put after the mongo request to avoid reducing performance.

Issue #515, opened 2020-02-23
Parameter "encoding" was removed in msgpack 1.0.0


misolietavec:
This means, in `inginious/common/message_meta.py` for that version of msgpack  references to `encoding = 'UTF-8'` should be removed.

PR #520, opened 2020-05-02
Add upper bound for msgpack and web.py

sargreal:
With the current configuration a clean install is not possible as the dependencies have made breaking changes

With the specified upper bounds a installation is possible again.

see #515 and webpy/webpy#635


anthonygego:
Thank you !

Issue #524, opened 2020-05-07
Internal error on task submission (clean install, tutorial course)


zootos:
Installing INGInious version 0.6 on a clean Ubuntu 18.04.4 LTS installation, following the provided [installation guide](https://inginious.readthedocs.io/en/v0.6/install_doc/installation.html) with all defaults selected when running `inginious-install`, results in most of the tasks in the [tutorial course](https://github.com/UCL-INGI/INGInious-demo-tasks) producing internal errors when ran. The only tasks not failing are the three tasks using the `mcq` environement. For the remaining tasks, the following error is given in the web interface:

```
An internal error occurred. Please retry later. If the error persists, send an email to the course administrator. [Submission #5eb3da069949682704d3cb19]
The grader did not return a readable output : 'NoneType' object has no attribute 'items'
```

Based on the log, the code that crashes seems to lie on line 574 in `inginious/agent/docker_agent/__init__.py`.  Based on my probing of the problem, it seems like there is something wrong with the `ingi/inginious-c-base` docker container.

### Steps to reproduce:

1. Install Ubuntu 18.04 LTS on a machine, VM or similar
2. Follow the installation guide for version 0.6 of INGInious
3. In the tutorial course, try to submit task 1 with the content `print("Hello World!")`

zootos:
If it helps with debugging, I can add that the problem was discovered when working on a setup, using version 0.5, that has an installation script that work perfectly fine in November 2019 and has not been changed since then. I do realize that there is only two versions (dev and latest) of the docker containers, and that the **lastest**-version is the one meant for version 0.6. Updating the setup to version 0.6 resulted in the same error both for the tutorial course and the course we used INGInious during the fall of 2019.

anthonygego:
Are you able to perform a ``docker logs`` on the container ?

I'm wondering if this could be due to the msgpack update and a possible version mismatch. Version 1.0 contains breaking changes. Did you downloaded the container from docker hub ? If this is the case, can you try building the two base containers from the v0.6 branch of the repo, ensuring the msgpack version that is used is below 1.0 ?

Maybe you'll have to change https://github.com/UCL-INGI/INGInious/blob/master/base-containers/base/Dockerfile#L15 to tell msgpack-python==0.5.6

zootos:
I am not able to perform a docker logs on the container. The container instance seems to be immediately killed, and it seems like the logs disappear when you kill the instance?

[...]

PR #525, opened 2020-05-08
Specify version for msgpack-python in the Docker container

zootos:
Recently, in #520, the version of `msgpack-python` and `web.py` was specified for INGInious itself in version 0.6. For `msgpack-python`,  this is due to the release of version 1.0, which changes how encoding is handled. To allow the docker containers to communicate with INGInious, we must ensure that the same major version of the library is used for both the container and INGInious itself. This PR places the same restriction on the major version of the library in the base container, as was done for INGInious in #520. If this change is not performed, running all tasks, except mcq, will result in errors like this one

```
An internal error occurred. Please retry later. If the error persists, send an email to the course administrator. [Submission #5eb3da069949682704d3cb19]
The grader did not return a readable output : 'NoneType' object has no attribute 'items'
```

This fixes the bug described in #524

### DIFF
--- a/base-containers/base/Dockerfile
+++ b/base-containers/base/Dockerfile
@@ -12,7 +12,7 @@ RUN     yum clean metadata && \
         yum -y install epel-release && \
         yum -y upgrade && \
         yum -y install python python36-pip python36 zip unzip tar sed openssh-server openssl bind-utils iproute file && \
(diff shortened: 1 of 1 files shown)


GuillaumeDerval:
Thanks!

Issue #523, opened 2020-05-05
Filter submissions with timeout and interal errors


jadinm:
To debug our grading scripts, it would be useful to have a way to see only submissions that triggered an internal error or a timeout.

When we have 100+ submissions, it becomes impossible to spot them without waiting for students to report them.

PR #527, opened 2020-05-18
Submissions list: add a filter for crashes and timeouts

jadinm:
Fixes Issue #523 

This adds a single checkbox to filter submissions that crashed or timed out.

Do you prefer a single checkbox like this or a radio group with three choices (i.e., timeout, crash, no filter)?


anthonygego:
LGTM.

PR #528, opened 2020-05-19
Return error to user if form has invalid deadlines

jadinm:
Otherwise, it silently sets the soft deadline to the hard one and the user might think that it worked


Drumor:
LGTM

PR #530, opened 2020-05-22
Harden event stream

GuillaumeDerval:
This modification ensures that the event stream always restart if Docker shuts down, and check its state.


anthonygego (approved):
LGTM. Is it still ok for you @GuillaumeDerval ?

anthonygego:
UP

GuillaumeDerval:
I'll UP myself here

PR #531, opened 2020-05-23
Update lighttpd configurations

ThaSami:
We learned these the hard way after an exam failing in production when there are 400+ students entering the exam.

### DIFF
--- a/doc/install_doc/installation.rst
+++ b/doc/install_doc/installation.rst
@@ -369,7 +369,14 @@ from another path than the specified one. In this case, lighttpd forces to set a
 while a root access if wanted, in order to serve static files correctly. Therefore, this environment variable is set
 to an empty string in addition to the rewrite rule.
 
+.. note::
 
+   The Default configuration doesn't optimize Inginious for performance, please refer to 
+   https://redmine.lighttpd.net/projects/lighttpd/wiki/Docs_Performance for more about performance optimising
+   you may also change 'max-procs' and append "PHP_FCGI_CHILDREN" => "someValue" inside "bin-environment"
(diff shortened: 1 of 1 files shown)


anthonygego:
Sorry to learn that. The documentations are indeed written for a quick, basic, deployment, on a single machine. They also lack some basic optimizations as static serving of ``public`` folders to avoid wasting fastcgi/wsgi sockets just to open the file and serve it without any cache, which can result into catastrophies in case you serve big static files via the tasks ``public`` folder with a lot of users logged in.

A single frontend process is not enough if you plan to have several tens or hundreds of requests at the same time. Moreover the behaviour differs from webservers: some just keeps the requests in a growing buffer and the frontend will take time to respond, in this case you can expect more requests to appear in the buffer as the frontend is taking time to respond because students tend to refresh the page when it takes time to respond, and it never ends. Some others just throw the requests when the application process buffer is full (typically ~100 requests), resulting in a lot of "backend unavailable" messages.

Our current setup at UCLouvain is an nginx instance with uwsgi running 4 inginious-webapp instances. This seems to be fine with 500 students logged in simultaneously.

I'll merge this warning but I think it could be useful to write a more generic one which is webserver-independent.

javawolfpack:
How do you run `inginious-webapp` as a uwsgi instance? I'm trying to configure our deployment with uwsgi & nginx as well; however, the documentation only mentions uwsgi once. I know how to run uwsgi with flask/Django as there is usually a wsgi.py file and app name but not sure how to do it with a command/binary. Only thing I can see is that the `inginious-webapp` says, "This command can run a standalone web server (see --host and --port options), but also as a FastCGI or WSGI backend."

Can you share details of how I can configure uwsgi to work with the ingenious apps? Thanks.

PR #547, opened 2020-07-08
adding kata agent

FlyearthR:
Adding a kata-container agent allowing a new type of containers requiring an isolated kernel to be used


GuillaumeDerval:
There is **_a lot_** of code duplication here. You should probably take time to:
1) extract the "common part" of the docker and kata agents, which is probably 90% of the code
2) make this a new class
3) inherit this class to differenciate the two behaviors.
Any other way that avoid code duplication should be ok too, if you see something better, feel free.

inginious/agent/docker_agent/_docker_interface.py
```diff
@@ -21,6 +21,14 @@ class DockerInterface(object):  # pragma: no cover
         Docker-py itself is already well tested.
     """
 
+    def __init__(self, type, runtime):
+        """
+        :param type: type of the container
+        :param runtime: runtime used by docker
+        """
+        self.type = type
```
GuillaumeDerval, line 29: Please document what are the available types here

base-containers/default/Dockerfile
```diff
@@ -3,3 +3,5 @@
 #inherit from the base container, which have all the needed script to launch tasks
 FROM    ingi/inginious-c-base
 LABEL org.inginious.grading.name="default"
+LABEL org.inginious.grading.agent_version=2
```
GuillaumeDerval, line 6: I don't think this needs to be repeated (it's already in `base`)

inginious/agent/docker_agent/_docker_interface.py
```diff
@@ -21,6 +21,14 @@ class DockerInterface(object):  # pragma: no cover
...
+        self.runtime = runtime
```
GuillaumeDerval, line 30: You should check the runtime. As its given directly to docker, it would be beneficial to avoid having possible injections here.

GuillaumeDerval:
I can't say I'm really happy with the architecture of the solution. Some open questions:
- Why can't "normal" docker containers run on kata? Why do we need a separate type of containers?
   It looks like to me that its simply another runtime that (as it starts a VM) allows to run the content as root. This could merely be a checkbox in the "docker environment" page indicating "start the container as root". This would avoid having two agents for docker and kata. A flag in containers could indicate that they *need* to be run as root, forcing the use of kata for example.
- Does the "student container" setup work with kata? Can a "normal" docker container start a "kata" student container?
  If its a limitation of the agent (or a too difficult task to manage right now), you should explicitly document it.

Also, I don't really like the duplicate of the INGInious base container source code in https://github.com/UCL-INGI/INGInious-containers/pull/45. Isn't there a way to avoid that?

GuillaumeDerval:
After informal discussion with @FlyearthR on a UCLouvain's private slack, he explained why we need two agents; mainly, its because the best docker storage backends for kata and for vanilla docker are different. In order to have optimal performances, we have two allow the kata "containers" to run on different environment/docker instance.

Still, I believe containers should be interoperable :-)

FlyearthR:
More precisely, by default, vanilla docker with runc as runtime uses overlay2 as storage driver, which is the best one for most of the INGInious usecases.
Kata-containers offers two different runtime : kata-fc, which uses firecracker for the kernel virtualization, and kata-qemu, which uses qemu. The kata-fc runtime needs devicemapper as storage driver while kata-qemu supports both devicemapper and overlay2 (among others).
In this PR, kata-qemu is used because, for now, it's the most effective runtime. Even if it supports overlay2, it is highly recommended to run it with devicemapper because there is a huge difference of performance for INGInious usecases (no need to have large storage operations to notice a performance difference).
You can find more details and explanation in the master thesis of Guillaume Everarts de Velp available here https://github.com/edvgui/LEPL2990-Manuscript/blob/master/manuscript.pdf

As for performance reasons we need two different storage drivers, we also need two different environments. So it makes sense to have two different instances of the Docker agent, one with the default runtime and another with a kata runtime. For the same reasons, this PR does not allow a normal docker container to run a kata student container.

For now, a kata-container cannot run a student container because it uses unix sockets, which are managed by the kernel, to communicate between the two containers, which have two isolated kernels. It could somehow be fixed but this is not implemented in the PR.

inginious/frontend/environment_types/kata.py
```diff
@@ -0,0 +1,50 @@
...
+                        limits['disk'] <= 0:
+                    raise Exception("Invalid limit")
+            except:
+                raise Exception("Invalid limit")
+        out["limits"] = limits
+
+        return out
+
+    def studio_env_template(self, templator, task, allow_html: bool):
+        return templator.course_admin.edit_tabs.env_docker(task.get("environment_parameters", {}), allow_html)
```
anthonygego, line 50: Here you call the ``env_docker`` template but you define a ``env_kata`` below... ?? If it's the same template, maybe it would be more clear to call it ``env_container`` or something like that.

anthonygego:
As discussed informally, it would be useful to have a small addition in the docs to know how to setup a Kata agent (how to install Kata and configure the Docker client to use it as backend...)

FlyearthR:
About https://github.com/UCL-INGI/INGInious/pull/547/files/974b05998c8a26d6746e2a80cceffc4bf4c16c3d#diff-946c5edce045a44d80e59770b4e5ce25R50 : here I should call `env_kata`.
But as you said, for now they are the same templates and I do not see any case where it could be interesting to distinguish them, do you?
If not, I suppose the best would be to
- remove inginious/frontend/templates/course_admin/edit_tabs/env_kata.html
- rename inginious/frontend/templates/course_admin/edit_tabs/env_docker.html to env_container.html
About the documentation, if you use a recent enough version of docker, the kata documentation is complete. Maybe a explanation on how to set up the default storage driver as device mapper could be useful. Except that, a link to kata documentation and a warning about `run_student` with kata environment, do you see anything else missing in the documentation?

anthonygego:
Once again, code shouldn't be copy/pasted. So if there is a need for them to be different one day, the two files should inherit from a common one, or the common one should be included in the two seperate files.

The INGInious documentation supposes the user is not familiar with Docker and only mentions it as a dependency. It would be nice to have a minimalist guide to how to setup kata as the container solution. (which file to modify,...)

anthonygego:
Please avoid doing merges of ``master`` branch into PR branches, this may complicate rebasing in other branches. A good practice is to perform a rebase when you're working on a PR, and to perform a merge when you want to merge a PR into the master branch.

I'm still annoyed by your note concerning ``run_student``. We need (*must* have), at least, a mechanism to avoid students to simply modify the ``run`` file and replacing it by ``feedback-result success``. Robust grading is part of the INGInious identity, we cannot afford to merge such a security threat.

I'm also still waiting for a clean version of your PR at UCL-INGI/INGInious-containers#45.

FlyearthR:
You advise me to mount the task folder in read-only mode (to ensure student cannot modify the runfile), what I understand as "append 'mode': 'ro' to https://github.com/UCL-INGI/INGInious/blob/master/inginious/agent/docker_agent/_docker_interface.py#L113 ". But if I do this, all the tasks using templates to fill in will not work anymore as those templates are located in `/task`.
I could also mount another folder in read-only mode for the runfile, but it would break all the existing tasks as they do not use this files architecture.
Or should I only mount `/task` in read-only mode for tasks using kata environment?

Also, should I revert my merge of `master` or is this remark a good practice I'll use next time?

anthonygego:
After discussing with @GuillaumeDerval I realize that we have two independant issues:

* Your kata solution does not support ``run_student`` because the two containers would'nt be running the same kernel. However, there should be different ways to mitigate and mimick a similar behaviour. @GuillaumeDerval suggests running a Docker container into the Kata VM **as a first hint**.
* Your SSH tasks container actually allows everything once the student becomes root and access the grading files.

I suggest to focus and work on this first issue so that we can progress with this PR. This would already allow to use Docker or Kata with the current existing exercices.

FlyearthR:
According to this https://github.com/kata-containers/runtime/issues/358 both running Kata in Docker and running Docker in Kata work.
If we run Docker into our Kata VM, we will need to grant privileged mode to our container to let it have the same capabilities as Kata. I don't exactly get why, but this was precisely what we wanted to avoid by using Kata. In this case though, the privileged container is run on Kata kernel and not on host kernel. Does it fix the problem using privileged containers?
If we run Kata into a Docker container, we perfectly isolate the student code which need its own kernel from the teacher tests, but we need the Docker container to be privileged (precisely with `--security-opt seccomp=unconfined` in https://github.com/daniel-noland/docker-in-kata). Again this was precisely what we wanted to avoid by using Kata. Is the protection from the student code sufficient? In this case, are there still use cases for the kata agent, which launch Kata containers?
With those elements, according to you which approach is the best?

GuillaumeDerval:
I think you are right, my suggestion makes no sense :-)
Then I propose to do exactly what we did with run_student: start another "container" on the agent directly, and link both containers.

In the current version of `run_student`, this is done by transmitting the file descriptors (stdin/err/out) through a Unix socket using black unix magic. Of course, a kata containers are actually vm, they do not share a kernel with the host, so we can't use FDs directly.

They are other ways to do that, I think the best one being opening a UDP socket between the two containers and send all the stdin/stdout/stderr/signals i/o inside this socket.

I believe this is the most versatile thing to do. We can even imagine starting "docker" containers that use run_student with a nice `--root` arg, that starts a "kata" container.

GuillaumeDerval:
(an expert told me that TCP is actually a better choice)

GuillaumeDerval:
I pulled your work on a kata branch locally to see what's still to be done and fix some things. I thus close this PR, the works continues in #600.

PR #549, opened 2020-08-03
Add structure to course

bensim602:


anthonygego:
Can you check that your branch is rebased to master ? There shouldn't be any master branch commits here, but there are. If you are not used to rebase, backup your branch before trying this :-)

bensim602:
It should be fixed

inginious/common/tasks.py
```diff
@@ -81,7 +78,11 @@ def input_is_consistent(self, task_input, default_allowed_extension, default_max
 
     def get_order(self):
         """ Get the position of this task in the course """
-        return self._order
+        tasks_id = self._course.get_toc().get_tasks()
```
anthonygego, line 81: There are still some parts of the code that only rely on the ``common`` files and not the ``frontend`` ones. So calling ``get_toc`` which is defined in ``FrontendCourse`` won't work in some rare cases. But actually I think the order field is only used by the frontend. So this function can be moved into ``FrontendTask``

inginious/frontend/pages/course_admin/task_list.py
```diff
@@ -2,89 +2,88 @@
...
+        errors = []
+        user_input = web.input()
+        try:
+            new_toc = json.loads(user_input["course_structure"])
+            valid, message = check_toc(new_toc)
+            if valid:
+                self.course_factory.update_course_descriptor_element(courseid, 'toc', new_toc)
+                course, __ = self.get_course_and_check_rights(courseid, allow_all_staff=False)  # don't forget to reload the modified course
+            else:
+                errors.append("Invalid table of content: " + message)
```
anthonygego, line 36: Do not forget to wrap strings with ``_()``  to allow their translations

inginious/common/tasks.py
bensim602: Indeed thanks. It's done

inginious/frontend/pages/course_admin/task_list.py
bensim602: Oups I forgot when I changed the errors. It's done

anthonygego:
Great job ! Thanks.

PR #551, opened 2020-08-05
fix: exception if hard_time limit is set

### LINKED ISSUE #550: task not shown if hard_time limit is set
if hard_time limit is set (to an integer value) there is an exception in docker.py -> task will not be shown

AugustH:
this fixes  #550


GuillaumeDerval:
LGTM. We can merge this as soon as you sign the CLA (sorry, mandatory step :-))

AugustH:
signed the cla

PR #555, opened 2020-08-14
Merge submissions

bensim602:


inginious/frontend/static/js/checked-list-group.js
```diff
@@ -12,6 +12,8 @@ $(function()
             $checkbox.attr('name', $widget.data('name'));
         if($widget.data('value'))
             $checkbox.attr('value', $widget.data('value'));
+        if($widget.data('class'))
+            $checkbox.addClass($widget.data('class'));
```
anthonygego, line 16: Is there no other way than modifying this file ? We may forget to make this change in case we update boostrap as it was fetch as is from the Internet.

inginious/frontend/pages/course_admin/utils.py
```diff
@@ -51,26 +51,106 @@ def get_course_and_check_rights(self, courseid, taskid=None, allow_all_staff=Tru
...
+            if self.user_manager.session_username() in audience["tutors"]:
+                tutored_users += audience["students"]
+
+        limit = params.get("limit", 50) if params.get("limit", 50) > 0 else 50
+
+        return users, tutored_users, audiences, tutored_audiences, tasks, limit
+
+    def get_users(self, course):
+        user_ids = self.user_manager.get_course_registered_users(course)
+        users = {user: self.user_manager.get_user_realname(user) for user in user_ids}
```
anthonygego, line 77: We may still have some users with no name. It'd be worth replacing this by ``self.user_manager.get_user_realname(user) or ''`` here.

inginious/frontend/static/js/checked-list-group.js
bensim602: I will just revert back to the way it was selected before

inginious/frontend/pages/course_admin/utils.py
bensim602: yes thanks I will do it

bensim602:
requested changes done and rebase to master

PR #558, opened 2020-08-18
Privacy policies

Drumor:
This allows management of privacy policies and terms of service on INGInious.

### DIFF
--- a/inginious/frontend/app.py
+++ b/inginious/frontend/app.py
@@ -220,7 +220,8 @@ def get_app(config):
                                      'frontend/templates/layout_lti',
                                      config.get('use_minified_js', True))
 
-
+    is_tos_defined = fs_provider.exists(os.path.join(config.get("static_directory", "./static"), "terms.yaml")) and \
+                     fs_provider.exists(os.path.join(config.get("static_directory", "./static"), "privacy.yaml"))
 
     # Init web mail
     smtp_conf = config.get('smtp', None)
@@ -242,6 +243,7 @@ def get_app(config):
     template_helper.add_to_template_globals("user_manager", user_manager)
     template_helper.add_to_template_globals("default_allowed_file_extensions", default_allowed_file_extensions)
     template_helper.add_to_template_globals("default_max_file_size", default_max_file_size)
+    template_helper.add_to_template_globals("is_tos_defined", is_tos_defined)
     template_helper.add_other("course_admin_menu",
                               lambda course, current: course_admin_utils.get_menu(course, current, template_helper.get_renderer(False),
                                                                                   plugin_manager, user_manager))
--- a/inginious/frontend/app.py
+++ b/inginious/frontend/app.py
@@ -220,8 +220,7 @@ def get_app(config):
                                      'frontend/templates/layout_lti',
                                      config.get('use_minified_js', True))
 
-    is_tos_defined = fs_provider.exists(os.path.join(config.get("static_directory", "./static"), "terms.yaml")) and \
-                     fs_provider.exists(os.path.join(config.get("static_directory", "./static"), "privacy.yaml"))
+    is_tos_defined = config.get("privacy_page", "") is not "" and config.get("terms_page", "") is not ""
 
     # Init web mail
     smtp_conf = config.get('smtp', None)
@@ -283,6 +282,8 @@ def get_app(config):
     appli.allow_deletion = config.get("allow_deletion", True)
     appli.available_languages = available_languages
     appli.welcome_page = config.get("welcome_page", None)
+    appli.terms_page = config.get("terms_page", None)
+    appli.privacy_page = config.get("privacy_page", None)
     appli.static_directory = config.get("static_directory", "./static")
     appli.webdav_host = config.get("webdav_host", None)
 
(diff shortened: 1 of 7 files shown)


inginious/frontend/app.py
```diff
@@ -220,7 +222,8 @@ def get_app(config):
                                      'frontend/templates/layout_lti',
                                      config.get('use_minified_js', True))
 
-
+    is_tos_defined = fs_provider.exists(os.path.join(config.get("static_directory", "./static"), "terms.yaml")) and \
```
anthonygego, line 225: As explained, I'd rather have the two pages specified in the ``configuration.yaml`` file, as this is already the case for ``welcome_page`` instead of hardcoding their names. This would make behaviour more uniform. The ``is_tos_defined`` can therefore become a check if those paramter are specified in the ``configuration.yaml``file.

inginious/frontend/pages/register.py
```diff
@@ -71,6 +71,7 @@ def register_user(self, data):
             r')@(?:[A-Z0-9](?:[A-Z0-9-]{0,61}[A-Z0-9])?\.)+[A-Z]{2,6}\.?$', re.IGNORECASE)  # domain
 
         # Check input format
+        print(data)
```
anthonygego, line 74: I guess this should be removed

inginious/frontend/pages/utils.py
```diff
@@ -133,6 +133,23 @@ def logger(self) -> logging.Logger:
         """ Logger """
         return logging.getLogger('inginious.webapp.pages')
 
+    def is_tos_defined(self):
```
anthonygego, line 136: You shouldn't need this if you rely on the same mechanisms as for ``welcome_page``

inginious/frontend/pages/register.py
```diff
@@ -101,7 +106,11 @@ def register_user(self, data):
                                             "password": passwd_hash,
                                             "activate": activate_hash,
                                             "bindings": {},
-                                            "language": self.user_manager._session.get("language", "en")})
+                                            "language": self.user_manager._session.get("language", "en"),
+                                            })
+                if self.is_tos_defined():
+                    self.database.users.find_one_and_update({"username": data["username"]},
```
anthonygego, line 112: This could probably be simplified and put in one single insert request.

    ...
    "tos_accepted": self.is_tos_defined() and "term_policy_check" in data
    ...

inginious/frontend/pages/utils.py
```diff
@@ -151,7 +168,9 @@ def GET(self, *args, **kwargs):
         Otherwise, returns the login template.
         """
         if self.user_manager.session_logged_in():
-            if not self.user_manager.session_username() and not self.__class__.__name__ == "ProfilePage":
+            user = self.database.users.find_one({"email": self.user_manager.session_email()})
```
anthonygego, line 171: Even if there many places in the code where this is already the case, it would be better to avoid requesting the user again from database (this method is checked for every page). Two solutions:
- Either we keep this piece of information in the session object and define a new function ``session_tos_accepted``
- Either we keep the complete user profile object in the session.

Another lightweight solution to explore would be to set the username to ``''`` at login if the TOS are not accepted, mimicking the behaviour of a non-set username.

inginious/frontend/user_manager.py
```diff
@@ -345,6 +345,14 @@ def get_user_email(self, username):
             return info[1]
         return None
 
+    def has_accept_tos(self, username):
```
anthonygego, line 348: You directly send a request to mongodb in ``utils.py``. It would be better to make a choice between the two ways to avoid duplicate code.

inginious/frontend/pages/register.py
```diff
@@ -83,9 +83,13 @@ def register_user(self, data):
         elif data["passwd"] != data["passwd2"]:
             error = True
             msg = _("Passwords don't match !")
+        elif self.app.terms_page is not None and self.app.terms_page is not None and "term_policy_check" not in data:
```
anthonygego, line 86: Two times same condition here.

inginious/frontend/pages/utils.py
anthonygego: So this line should be useless now.

inginious/frontend/pages/utils.py
```diff
@@ -151,7 +151,11 @@ def GET(self, *args, **kwargs):
...
+            if (not self.user_manager.session_username() or (self.app.terms_page is not None and
+                                                             self.app.terms_page is not None and
```
anthonygego, line 156: Two times same condition ``terms_page``

inginious/frontend/pages/preferences/profile.py
```diff
@@ -97,6 +97,12 @@ def save_profile(self, userdata, data):
             return result, msg, error
 
         msg = _("Profile updated.")
+
+        #updating tos
+        if self.app.terms_page is not None and self.app.terms_page is not None:
```
anthonygego, line 102: You still have two identical conditions here.

inginious/frontend/user_manager.py
```diff
@@ -121,6 +121,12 @@ def session_realname(self):
             return None
         return self._session.realname
 
+    def session_tos_signed(self):
+        """ Returns True if the current user has signed the tos"""
+        if not self.session_logged_in():
+            return None
+        return self._session.tos_signed
```
anthonygego, line 128: It might be more secure to do a ``get("tos_signed", False)``. What happens if the user is logged in but hasn't accepted the ToS yet ? (in case of external login for instance)

anthonygego: Any news ?

PR #561, opened 2020-08-27
Move course/task factories to frontend

anthonygego:
This is a very first step in order to ease the course and task storage in database as we currently want to avoid the backend to rely on the frontend database. Was done:

1.  **Making the backend files independant of the course/task classes.** This means the ``task.yaml`` is not opened anymore by the backend files. The only usage done was in MCQ agent to check the student answers with expected ones. I decided to send the problem set as an additional message, next to the student input. The task ``check_answer`` function was moved to the MCQ agent, which is more coherent with the way agents are designed. 
1. **Merging Course and WebAppCourse classes, and moving CourseFactory to frontend**. This is essentially file moving with some glue.
1. **Merging Task and WebAppTask classes, and moving TaskFactory to frontend**. This is essentially file moving with some glue
1. **Moving Task/Course tests to frontend**. They still work. Pretty neat, huh ?
1. **Merging the HookManager and the PluginManager**. The HookManager has never been actually used in the backend. This makes things more clear and coherent, which is not a bad thing considering the current state of the code.

This PR is essentially code moving but already quite big. So I think this is a good checkpoint.


GuillaumeDerval:
I did not try it, but after a first read, it LGTM. I'll see during this week-end to try it.

Drumor:
Didn't test it but for seems good to me.

Drumor (approved):
LGTM

Drumor:
Just finish testing. Except plugin issues (they need to be review) I saw nothing weird.
LGTM

PR #573, opened 2020-10-02
[FRONTEND] Fix courselist presentation - Issue with long description

Drumor:


anthonygego:
It would be better to define a new class, to avoid redefining the behaviour of bootstrap ones.

PR #582, opened 2020-10-30
Dev doc

Drumor:
review submission doc and add evaluation documentation.


doc/dev_doc/submissions.rst
```diff
@@ -0,0 +1,51 @@
...
+
+The backend stores the job in a waiting queue. When an agent released and the job is the next one in the queue, The job is moved to the running queue, a *BackendNewJob* message is sent to the agent.
+
+Agent treats then the job and once it's over, returns a *AgentJobDone* message to the backend. This one removes job from the running job queue and send a *BackendJobDone* to the client. The client end the process by displaying the result within the frontend and by updating information in the database.
+
+State
+-----
+
+Submission is represented as a dictionary from the database. Easy to share and structured.
+The concept of Submission have no specific object class in INGInious but has no necessity of that.
```
anthonygego, line 25: It would be more interesting to have the structure explained instead of discussing the relevance of an object-mapping that would be, actually, necessary.

doc/dev_doc/submissions.rst
```diff
@@ -0,0 +1,51 @@
...
+Submission come with an interesting property : State.
+
+State is dynamic property that lets you store data related to the submission.
+State can be viewed as a bag. You can transport things with no condition about what is in the bag. 
+You can set state with basic or complex data types.
```
anthonygego, line 30: Only strings. It should be viewed as a register. This is the closest concept to compare to. The output value of a submission is the input value of the next submission. To be more precise: the output value of the evaluation submission is the input value of the next submission. That's the definition that should be given. And it is actually only managed by the frontend since it is stored in database along with the evaluation submission for a given user/taskid.

doc/dev_doc/submissions.rst
```diff
@@ -0,0 +1,51 @@
...
+State can also be used in the frontend part because it has a task input property. You can so, with basic Javascript, use this.
+
+Evaluation
+----------
+
+Submission evaluation is possible in two different ways : By evaluating the best one or the last one.
+As main part of the evaluation is done within the backend with only one submission, the selection part is done on frontend part.
+If the user run a new submission and that it suits with the evaluation, the submission id is store in database in the user task collection.
+The evaluation part really starts when the job is treated by the agent. The agent receives a *BackendNewJob* message and have to handle it. It starts its new_job function.
+For MCQ agent, it starts by getting files from task and course file system and the translation files. Then it loops over the question and calculate the number of good answers to return the correct feedback.
```
anthonygego, line 41: This should be more precise. There is a ``check_answer`` function for each problem type that is called by the main function in the MCQ agent.

doc/dev_doc/submissions.rst
```diff
@@ -0,0 +1,51 @@
...
+For Docker agent, it start by synchronously generate the needed element. It begins with the file system, copying files (especially task files and $common files) and creating the container. It continues by adding new info of the new container and then of course starts this one.
+All this work is done within the Asyncio event loop.
```
anthonygego, line 43: We need to describe more precisely what is put in the event loop. An illustrated execution example would be even better one detailed.

Drumor: Not sure that it suits your expectation :-)

anthonygego:
I think this was globally OK. I'll merge it and I'll review it a second time while working on a new table of contents

PR #583, opened 2020-11-03
Fix a race condition in the submission manager

GuillaumeDerval:
The callback can actually be called BEFORE the new_job function returns.


anthonygego:
LGTM

PR #586, opened 2020-11-03
Use another function to render jinja templates

GuillaumeDerval:
* This allows to call subfolder when using use_jinja=True
* and will allow to simply remove get_renderer and get_custom_renderer when the migration is done.


anthonygego:
LGTM

PR #592, opened 2020-11-10
Fix jobid handling

GuillaumeDerval:
Previously, the client "job id" and the backend "job id" were different.
The backend job id was actually a tuple containing the address of the client and the jobid it gave.

A very simple form of authetication was then used to allow (or not) clients to access data or kill a job: the Backend job id must be the same (i.e. the request should come from the original client).

In practice, this is a safety guarantee that is never used (at least we never seen it used in the wild) and we have problem with uwsgi-like setups where multiple frontend processes (and thus clients...) are started.

This PR then simply modify the backend job id to be the one sent by the client. The client should thus ensure that the id created is unique. uuid4(), as currently done, should be sufficient.

### DIFF
--- a/inginious/backend/backend.py
+++ b/inginious/backend/backend.py
@@ -6,7 +6,10 @@ import asyncio
 import logging
 import queue
 import time
+from collections import namedtuple
+
 import zmq
+from typing import Dict
 from zmq.asyncio import Poller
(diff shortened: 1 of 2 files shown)


anthonygego:
LGTM

PR #596, opened 2020-11-17
Add a way to configure session cookies.

### LINKED ISSUE #575: Bug - <class 'KeyError'> at /auth/callback/uclouvain
Hello,

It seems this [line](https://github.com/UCL-INGI/INGInious/blob/master/inginious/frontend/plugins/auth/saml2_auth.py#L86) needs additional check (like `"redir_url" in auth_storage`) as I rarely got this when sign in : 

[Moderated]

GuillaumeDerval:
This puts a default timeout of a day for the session, rather than "session", which caused some problems...

Fixes #575.

### DIFF
--- a/inginious/frontend/app.py
+++ b/inginious/frontend/app.py
@@ -5,6 +5,10 @@
 
 """ Starts the webapp """
 import builtins
+import os
+import sys
+from binascii import hexlify
+
 import pymongo
 
 import inginious.frontend.pages.course_admin.utils as course_admin_utils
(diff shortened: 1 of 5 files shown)


inginious/frontend/app.py
```diff
@@ -106,6 +108,30 @@ def _put_configuration_defaults(config):
         config['allowed_file_extensions'] = [".c", ".cpp", ".java", ".oz", ".zip", ".tar.gz", ".tar.bz2", ".txt"]
     if 'max_file_size' not in config:
         config['max_file_size'] = 1024 * 1024
+
+    if 'session_parameters' not in config or 'secret_key' not in config['session_parameters']:
+        print("/!\\ WARNING /!\\", file=sys.stderr)
+        print("You did not define the 'session_parameters' entry in your INGInious configuration file.", file=sys.stderr)
+        print("A default secret key has been used. This is potentially dangerous.", file=sys.stderr)
```
anthonygego, line 115: I think we should simply not allow to run the app without the secret key set in the configuration file instead of using a default one. Unless my memory fails, I think this is the Flask behaviour.

inginious/frontend/installer.py
```diff
@@ -512,12 +520,33 @@ def configure_authentication(self, database):
...
+                self._display_warning("Please consult the documentation to add this plugin.")
+            else:
+                continue
+
+        options["session_parameters"] = {
+            "timeout": 86400,  # 24 * 60 * 60, # 24 hours in seconds
+            "ignore_change_ip": False,
+            "secret_key": "fLjUfxqXtfNoIldA0A0G",
+            "secure": False
+        }
```
anthonygego, line 538: Is this necessary as the parameters are asked below with default values ? This is a bit confusing.

inginious/frontend/installer.py
```diff
@@ -512,12 +520,33 @@ def configure_authentication(self, database):
...
+            self._display_info("- 2. SAML auth plugin. This plugin allows to connect to a distant SAML Identity Provider.")
 
             plugin = self._ask_with_default("Enter the corresponding number to your choice", '1')
-            if plugin not in ['1']:
-                continue
-            elif plugin == '1':
+            if plugin == '1':
                 options["plugins"].append(self.ldap_plugin())
+            elif plugin == '2':
+                self._display_warning("Please consult the documentation to add this plugin.")
```
anthonygego, line 529: This message could even be displayed for all others plugins, actually.

inginious/frontend/installer.py
```diff
@@ -512,12 +520,33 @@ def configure_authentication(self, database):
...
+
+        options["session_parameters"]['timeout'] = self._ask_integer("How much time should a user stay connected, "
+                                                                     "in seconds? The default is 86400, one day.", 86400)
+        options["session_parameters"]['ignore_change_ip'] = not self._ask_boolean("Should user be disconnected when "
+                                                                                  "their IP changes? It may prevent "
+                                                                                  "cookie stealing.",
+                                                                                  True)
+        options["session_parameters"]['secure'] = self._ask_boolean("Do you plan to serve your INGInious instance only"
+                                                                    " in HTTPS?", False)
+        options["session_parameters"]['secret_key'] = hexlify(os.urandom(32))
```
anthonygego, line 548: Is the hexlify necessary ?

GuillaumeDerval: Yup, otherwise it's a binary string

anthonygego:
LGTM

Issue #569, opened 2020-09-15
Course administration - Allow to control course access based on email domain


vvandenschrieck:
I'm using free registration on my inginious instance, but I would like to force my students to use their school adresses.  
I believe that I can configure a control access list with all my student's adresses, but it's kind of overkill for my needs.  Is there a way to force the user to register with an email address from a given domain?

PR #597, opened 2020-11-18
Regex for the ACLs

GuillaumeDerval:
This started as a way to solve @vvandenschrieck 's #569 and then I discovered that most of the calls to `is_registration_possible` actually did not work correctly. So I fixed everything, created a unified data structure for user info (named UserInfo in a sudden burst of creativity), updated the code mostly everywhere, and implemented a (deadly simple) regex system for the ACLs.

And documented it. TL;DR: '*' is a wildcard now.


anthonygego:
LGTM

Issue #601, opened 2020-11-25
Check that feedback is actually a string on submission


jadinm:
A faulty grading script can set a list instead of a string for feedback. For instance with the following code:

```python
feedback.set_problem_feedback(["a", "problematic", "feedback"], "problem-id")
```
Student submissions made with this grading script cannot be viewed later on. The web interface produces the following stacktrace instead of showing the submission:

![Screenshot_2020-11-25  class 'AttributeError' at submission](https://user-images.githubusercontent.com/4190063/100261973-cfc85f00-2f4b-11eb-9cd0-e6c7101df5aa.png)

Of course, the grading script could have been correct in the first place but that would be cool to have a standard "Internal error" and to be able to visualize the student code instead of a stack trace :)

Drumor:
Solved by #603

PR #603, opened 2020-12-02
verifying feedback is str

Drumor:
issue #601

### DIFF
--- a/base-containers/base/inginious_container_api/feedback.py
+++ b/base-containers/base/inginious_container_api/feedback.py
@@ -47,19 +47,21 @@ def save_feedback(rdict):
 # Doing the real stuff
 def set_global_result(result):
     """ Set global result value """
-    rdict = _load_feedback()
-    rdict['result'] = result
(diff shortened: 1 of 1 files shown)


anthonygego:
This check should also be done at the agent side because anyone is able to change the feedback file manually.

anthonygego:
I'm still not satisfied with this. There should really be a check to ensure the feedback inserted in db in a string. And this should be warned after the container is run, so the check should be either in agent-docker or in backend files. 

You can debug and trigger the issue by simply putting a dummy array list in the container code instead of the actual feedback string.

GuillaumeDerval:
Moreover, this fixes actually make the feedback api fail silently (when giving anything else than a str). You should make it throw an exception or something like that.

anthonygego:
What abour this ?

Drumor:
Still have to rework this.

Drumor:
I think this one is good, any review is appreciate :)

Drumor:
Rebase done :)

Issue #173, opened 2016-11-28
Submit code with Ctrl+Enter


shlomihod:
Pressing Ctrl+Enter in Task page, when the keyboard cursor is in a textbox will submit the code.

PR #604, opened 2020-12-03
Press ctrl+enter in CodeMirror to submit

AlexandreDubray:
The long-awaited feature in #173


anthonygego:
I really am skeptical about this. It makes submission quite implicit and this will be only working with CodeMirror problems. This could be confusing for users who might argue that they didn't want to submit due to wrong key strokes when submission limitations or penalties are applied. I'd rather have a confirmation or a scroll to the submit button instead of directly submitting.

anthonygego:
LGTM.

PR #605, opened 2020-12-04
Task dispensers

anthonygego:
This introduces a mechanism to change the task displaying scheme. It is heavily inspired from the work that was done for problem types and, as it is a quite a raw refactor, is subject to evoluate as tasks dispensers will be added.

A Task Dispenser is provided with the task factory listing and a free data structure storing its configuration. 
It currently provides:
- A rendering function of the task listing for the student view, including as many JS needed. (Equivalent of subproblem render)
- A rendering function of the task listing for the admin view, including as many JS needed. (Equivalent of subproblem render_edit)
- A check function to check data from the HTML form and return a pythonic version to store in YAML file (Equivalent of subproblem check)
- A filtering function that specifies if a task can be accessed by the students according to its username. This check was currently not done, only the task file accessibility parameters were taken into account: so a student could access a task missing from the table of contents using its URL.
- A function returning the serialized/flattened ordered list of tasks. This is used for lists such as the ones in submission pages.
- A function returning the position of a task in the serialized list. This is used for sorting a subset of the course tasks, in some specific pages.
- A javascript function ``dispenser_structure_<dispenser_id>`` to pregenerate the "course_structure" that will be POSTed in the edit page, and given to the check function mentioned above.

It is also given two javascript functions ``dispenser_wipe_task(taskid)`` and ``dispenser_delete_task(taskid)`` in order to implement these database-affecting operations.

The current scheme was defined as the _Table of Contents_ task dispenser and __set as default__ (it is therefore retrocompatible). HTML parts for sections were moved into an utility functions package in order to be used with other task dispenser. The task dispenser can be switched from the task list edition page (a button is located next to the page title, to the right).

This mechanism is defined with many degrees of freedom as it allows to completely rewrite both the pages. This was first of all done to avoid any premature optimization that would have made this refactor more tricky. The utility functions with all the (sub-)sections rendering features might be reused in other task dispensers. However, these degrees of freedom allows to support a lot of things which may not based on a task listing, such as adapative testing.

A new task dispenser for designing "combinatory tests" (several tasks randomly picked among a bucket) is also provided and heavily rely on the existing (and improved) sections mechanism.

Coming next:
- A basic documentation


Drumor (commented):
LGTM except some tiny problems i had while testing and that we already talk about.

PR #607, opened 2020-12-14
Optimize task distribution and grade computation

anthonygego:
Pull request #605 has introduced some huge computational problems due to a raw refactor.

Problems:
- Student score computation is drastically reduced due to the ``task_is_visible`` function which was called T*U times, where T is the amount of course tasks and U the amount of user tasks. This function called the task dispenser ``filter_accessibility`` function that typically looped on the task list. In cases U=S*T where S is the amount of students in a course, the student score computation was done with an awful complexity of ST^3
- Visibility has become quite cahotic: the task listing looped on every task and checked if it could be displayed, resulting in another terrible complexity. Specific code had to be added at the task page code to allow admins to access the tasks that are not given by the task dispenser. The code that computed the previous and next task for the navigation was not optimal neither.

This PR solves those issues. More precisely :
- Task visibility has a clearer definition throughout the code:
  - A task that is visible for the user is a task that has to be taken into account for score computation, and is served by the task dispenser. The course task list is displayed the same way for all users. _Hidden tasks are not listed anymore for admins._
  - Hidden tasks, that are not served by the task dispenser, should remain accessible from the admin pages to the admins to allow them to test their tasks
- A ``get_user_task_list`` function is introduced at the task dispenser side to return the specific task lists, that will be used for displaying visible tasks and for grade computation. This function has a complexity of S*T and is called once.
- The old ``user_manager.task_is_visible`` function is only used in the task page code and contains the specific admin filter to allow staff to access a hidden task.

### DIFF
--- a/inginious/frontend/pages/course.py
+++ b/inginious/frontend/pages/course.py
@@ -59,30 +59,29 @@ class CoursePage(INGIniousAuthPage):
         else:
             tasks = course.get_tasks()
 
+            user_task_list = course.get_task_dispenser().get_user_task_list([username])[username]
+
             # Get 5 last submissions
             last_submissions = []
-            for submission in self.submission_manager.get_user_last_submissions(5, {"courseid": course.get_id(), "taskid": {"$in": list(tasks.keys())}}):
-                if self.user_manager.task_is_visible_by_user(tasks[submission['taskid']], username, False):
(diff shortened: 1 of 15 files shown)


Drumor (commented):
LGTM

Drumor:
Practical test is good to me :)

PR #608, opened 2020-12-15
Task dispensers : create task file before editing

anthonygego:
As the task file does not exist yet at edition time :
- The code was aware of sections.
- TOC templates made a redirection to the edit page with the target section as a GET argument.

Moreover, code in  #605 was incorrect, as https://github.com/UCL-INGI/INGInious/blob/master/inginious/frontend/pages/course_admin/task_edit.py#L288 tries to use raw data as a ``SectionsList``, resulting in internal errors when saving non-existing tasks.

This PR :

- Introduces the javascript ``dispenser_add_task`` which allows the dispenser to tell what new taskid has to be created
- Adds logic to create the tasks that are passed in arguments, as this is the case for the courses.
- Disallow further edition of non-existing tasks


Drumor (commented):
LGTM

PR #620, opened 2021-01-21
lti autobind - reworked

AugustH:
reworked version, included remarks from #424


anthonygego:
I'll close this PR as this was still not really satisfying as a short circuit. We need to to rewrite some part of the LTI implementation, partly because of the LTI 1.1 deprecation for LTI Advantage 1.3 ( #870 ) and because cookieless session managemet as it is currently implemented right now is some sort of a nightmare. 

We'll try to think of a solution for people that only want one authentication method for the whole platform in the same time but we can't promise anything.

PR #631, opened 2021-02-18
List/dicts are now output as JSON by get_input. JQ is available in al…

GuillaumeDerval:
…l containers to easily process the data.


anthonygego:
LGTM.

PR #638, opened 2021-02-23
teacher doc update

darkyne:
Tried to update the teacher documentation based on what I believe was not up to date anymore:
-Marketplace addition (in notes)
-Task addition user-interface changed in the web page
-Subproblems can be configured from webapp 
-@random added in the getinput API (examples)

When reading the doc, it was unclear when it was talking about a task, a problem, a subproblem...
I believe it's more clear now.
Tasks can be added/removed from the Tasks page where the *Edit task* button allow to modify them on the Edit task page.
Problems/Subproblems (I believe both terms are for the exact same thing) can be configured from the Edit task page.

Maybe removing problem and subproblem notions and using the same word everywhere ?


doc/teacher_doc/course_admin.rst
```diff
@@ -70,30 +70,26 @@ Task edition
...
-write the YAML problem description as described in the task file format.
+Adding and removing subproblems is very easy in the task editor. Go to the *Subproblems* tab and add a new 
+problem-id (alphanumerical characters) and a problem type. You can configure the problem context from this page, 
+even finalize the entire problem for mcq or match problems.
 
 When editing a multiple choice problem, you're asked if the student is
 shown a multiple-answers- or single-answer-problem and which of the
 possible choices is (are) good answer(s).
 
+Only more complex problems such as code implementation will require more configuration with a grading script in the task files.
```
anthonygego, line 87: It would be better to specify that there currently are two types of grading environment:
- MCQ+ based, which call the internal grading function
- Container-based, which allows to define your own grading script and gives access to the user inputs.

And that among the default subproblems that are shipped with a typical INGInious installation, those two types have an internal grading function:
- MCQ
- Match problems

Maybe this is better to rename or split the section to make this more clear.

anthonygego (changes requested):

doc/teacher_doc/course_admin.rst
darkyne: Probably better now.

anthonygego: Well, it's a bit better, but actually you would have more chance to copy/paste my comment :D The documentation should explicitely tell that those two features are possible for each kind of subproblem, but that in the bundled subproblems type, only mcq and match problems support both. This is important because some other available subproblems (math, network-trace,...) support both.

anthonygego (approved):
LGTM

PR #639, opened 2021-02-24
Displays version on footer

anthonygego:
This relies on setuptools_scm, which displays the version, distance, commit hash and date. This is very useful as our local instance is almost up to date, to know exactly on which version we are.

setuptools_scm tests the pip install, then the local repository, or raise a ValueError, in which case the version string is set by hand.

The version string is displayed in the footer.


Drumor:
LGTM

PR #642, opened 2021-02-26
Fix documentation build

anthonygego:
Some versions of Sphinx have issues inspecting the flask module imports, due to the context-dependent variables ``request``, ``current_app`` and ``session``, and fail when such a variable is imported, as Flask raise an exception if the context is not set. This is particularly the case with current version on ReadtheDocs.

To mitigate this issue in a clean way, all such imports were removed. Lines like:

    from flask import request
    url = request.url
    
were replaced by:

    import flask
    url = flask.request.url


Drumor:
LGTM

PR #648, opened 2021-03-02
add installation guide for Ubuntu+Apache2

erelsgl:
This guide is based on my own experience. Please verify that there are no mistakes.


anthonygego:
Thank you for this. As there is no Ubuntu user in the team, this helps us a lot to keep the procedures up to date.

Unfortunately I can't merge it as is, as it repeats a lot of existing text, sometimes identical, sometimes similar so that it is not even easy to see what changes you've made.
 
 May I ask you to re-work this a bit and try to integrate this and update the existing https://github.com/UCL-INGI/INGInious/blob/master/doc/install_doc/installation.rst ? The install procedure didn't change since v0.6 so changes are safe.
 
 For parts that are similar (Apache setup for instance), you can use the sphinx-tabs directives with a ``RHEL`` and ``Ubuntu`` tab. It is pretty simple to use : https://github.com/executablebooks/sphinx-tabs

erelsgl:
The problem is that somse things should be done in a different order, so it is not only "A or B". But I have another suggestion: I can try to write an installation script for Ubuntu+Apache. Then, in the docs we only have to refer users to this script.

anthonygego:
What steps are in different order ? You can put more than the strict minimum in each tabs. You do not need to remove *all* the duplicates, but try to reduce it, I mean:
- The dependencies : there is already a section for Ubuntu 16, it can be updated to Ubuntu 18
- The pip install part is identical
- If needed, the section for Apache can be split in two tabs with all the text kept in both tabs
- The _Optional apps_ part seems to be identical, and still mentions lighttpd and CentOS, which are therefore irrelevant to Ubuntu+Apache.

If you are willing to, yes, you can write a script and put it in the ``utils`` folder where there already are some misc. scripts, and then refer to it in the docs, it's always welcome !

erelsgl:
I merged the installation instructions for Apache into the main file. Some things I did:

* Added a recommendation to make a folder for inginious, and subfolders for tasks and backup. I think it is useful for any installation.
* Added an explanation that, if you use a web-server, you do not need to manually run the frontend. This is one thing that confused me previously.
* Moved the optional apps - webdav and webterm - to the end of file.

I did not use tabs, since I do not have a compiler for rst files so I cannot check that I did this correctly. But at least the headings should be clear now.

doc/install_doc/installation.rst
```diff
@@ -134,14 +134,27 @@ compatible. This includes Moodle, edX, among many others.
...
 To configure the web app automatically, use the ``inginious-install`` CLI.
 
 ::
 
     $ inginious-install
 
-This will help you create the configuration file in the current directory. For manual configuration and details, see
-:ref:`ConfigReference`.
+This will help you create the configuration file in the current directory. 
+When asked about the tasks folder, enter an absolute folder: /var/www/inginiuos/tasks .
```
anthonygego, line 152: Looks way much better ! There's just a small typo here : ``inginious``. I'll be able to merge it afterwards.

erelsgl: Fixed

anthonygego:
Than you !

PR #655, opened 2021-03-19
Adding the upcoming_tasks plugin

darkyne:
This adds the upcoming_tasks plugin as one of the default plugin in the inginious.frontend.plugins folder.
It is enabled by adding the following line in the configuration.yaml file:
` - plugin_module: "inginious.frontend.plugins.upcoming_tasks" `

It adds a "Upcoming tasks" button in the left menu (from the homepage without selecting a course) that leads to a new page.
This page displays the available unsuccessful tasks that have a deadline for courses the user is registered in.
The tasks are displayed by sections (one section per course).
The sections (courses) are ordered based on their closest deadline then the tasks themselves are ordered based on their deadlines.
A button on the right allows to display only tasks with deadlines in a specific range.

If you find any potential issue please let me know.
The main default I can see for now is that some little parts of the user interface are in english and I don't know how to make them automatically translate to the user language .

Here is a little example:

Student with some work to do:
image

Happy student:
image

### DIFF
--- a/inginious/frontend/plugins/upcoming_tasks/__init__.py
+++ b/inginious/frontend/plugins/upcoming_tasks/__init__.py
@@ -1,0 +1,122 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INGInious. See the LICENSE and the COPYRIGHTS files for
+# more information about the licensing of this file.
+#
+
+""" A plugin that allow students to see their futur work in a single place for all their courses """
+import os
+from collections import OrderedDict
+from datetime import datetime, timedelta
+import flask
+from flask import send_from_directory
+
+from inginious.frontend.pages.utils import INGIniousPage, INGIniousAuthPage
+
+PATH_TO_PLUGIN = os.path.abspath(os.path.dirname(__file__))
+
+def menu(template_helper):
+    """ Displays the link to the board on the main page, if the plugin is activated """
+    return template_helper.render("main_menu.html", template_folder=PATH_TO_PLUGIN + '/templates/')
+
+class StaticMockPage(INGIniousPage):
+    """ MockPage based on auto-evaluation plugin structure """
+
+    def GET(self, path):
+        return send_from_directory(os.path.join(PATH_TO_PLUGIN, "static"), path)
+
+    def POST(self, path):
+        return self.GET(path)
+
+
+class UpComingTasksBoard(INGIniousAuthPage):
+
+    def GET_AUTH(self):
+        """ Called when reaching the page """
+        time_planner = "unlimited"
+        return self.page(time_planner)
+
+    def POST_AUTH(self):
+        """ Called when modifying time planner """
+        user_input = flask.request.form
+        time_planner = user_input.get("time_planner", default="unlimited")
+        return self.page(time_planner)
+
+    def time_planner_conversion(self, string_time_planner):
+        """ Used to convert the time_planner options into int value """
+        if string_time_planner in ["7", "14", "30"]:
+            return int(string_time_planner)
+        return 100000
+
+    def page(self, time_planner):
+        """ General main method called for GET and POST """
+        username = self.user_manager.session_username()
+        all_courses = self.course_factory.get_all_courses()
+        time_planner = self.time_planner_conversion(time_planner)
+
+        # Get the courses id
+        open_courses = {courseid: course for courseid, course in all_courses.items()
+                        if self.user_manager.course_is_open_to_user(course, username, False) and
+                        self.user_manager.course_is_user_registered(course, username)}
+        open_courses = OrderedDict(sorted(iter(open_courses.items()), key=lambda x: x[1].get_name(self.user_manager.session_language())))
+
+        # Get last submissions for left panel
+        last_submissions = self.submission_manager.get_user_last_submissions(5, {"courseid": {"$in": list(open_courses.keys())}})
+        except_free_last_submissions = []
+        for submission in last_submissions:
+            try:
+                submission["task"] = open_courses[submission['courseid']].get_task(submission['taskid'])
+                except_free_last_submissions.append(submission)
+            except:
+                pass
+
+        # Get the courses tasks, remove finished ones and courses that have no available unfinished tasks with upcoming deadline in range
+        tasks_data = {}
+        succeeded_courses = []
+        for courseid, course in open_courses.items():
+            tasks = course.get_tasks()
+            outdated_tasks = [taskid for taskid, task in tasks.items() if (not task.get_accessible_time().is_open()) or ((task.get_accessible_time().get_soft_end_date()) > (datetime.now()+timedelta(days=time_planner)))]
+            new_user_task_list = course.get_task_dispenser().get_user_task_list([username])[username]
+            new_user_task_list = [task_id for task_id in new_user_task_list if task_id not in outdated_tasks]
+            tasks_data[courseid] = {taskid: {"succeeded": False, "grade": 0.0} for taskid in new_user_task_list}
+            user_tasks = self.database.user_tasks.find({"username": username, "courseid": course.get_id(), "taskid": {"$in": new_user_task_list}})
+            for user_task in user_tasks:
+                if not user_task["succeeded"]:
+                    tasks_data[courseid][user_task["taskid"]]["succeeded"] = user_task["succeeded"]
+                    tasks_data[courseid][user_task["taskid"]]["grade"] = user_task["grade"]
+                else:
+                    del tasks_data[courseid][user_task["taskid"]]
+            # Remove courses with no unfinished available tasks with deadline and lti courses
+            if (len(tasks_data[courseid]) == 0) or course.is_lti():
+                succeeded_courses.append(courseid)
+
+        # Remove succeeded courses (including lti courses)
+        for succeeded_course in succeeded_courses:
+            del open_courses[succeeded_course]
+
+        # Sort the courses based on the most urgent task for each course
+        open_courses = OrderedDict( sorted(iter(open_courses.items()), key=lambda x: (sort_by_deadline(x[1], tasks_data[x[0]].keys())[0]).get_accessible_time().get_soft_end_date() ))
+
+        return self.template_helper.render("coming_tasks.html",
+                                           template_folder=PATH_TO_PLUGIN + "/templates/",
+                                           open_courses=open_courses,
+                                           tasks_data=tasks_data,
+                                           sorting_method=sort_by_deadline,
+                                           time_planner=["7", "14", "30", "unlimited"],
+                                           submissions=except_free_last_submissions)
+
+
+def sort_by_deadline(course, user_urgent_task_list):
+    """ Given a course (object) and a list of user urgent tasksid,
+    returns the list of urgent tasks (objects) for that course ordered based on deadline """
+    course_tasks = course.get_tasks()
+    course_user_urgent_task_list = list(set(course_tasks).intersection(user_urgent_task_list))
+    ordered_tasks = sorted(course_user_urgent_task_list, key=lambda x: course.get_task(x).get_accessible_time().get_soft_end_date())
+    return [course_tasks[taskid] for taskid in ordered_tasks]
+
+
+def init(plugin_manager, _, _2, config):
+    """ Init the plugin """
+    plugin_manager.add_page('/coming_tasks', UpComingTasksBoard.as_view("upcomingtasksboardpage"))
+    plugin_manager.add_page('/plugins/coming_tasks/static/<path:path>', StaticMockPage.as_view("upcomingtasksstaticmockpage"))
+    plugin_manager.add_hook('main_menu', menu)
--- a/inginious/frontend/plugins/upcoming_tasks/templates/coming_tasks.html
+++ b/inginious/frontend/plugins/upcoming_tasks/templates/coming_tasks.html
@@ -1,0 +1,111 @@
+{# This file is part of INGInious. See the LICENSE and the COPYRIGHTS files for #}
+{# more information about the licensing of this file. #}
+
+{% extends "layout.html" %}
+{% block title %}{{ _("Urgent tasks") }}{% endblock %}
+
+{% block navbar %}
+    <nav aria-label="breadcrumb">
+        <ol class="breadcrumb">
+            <li class="breadcrumb-item active"><a href="#"><i class="fa fa-th-list"></i> {{ _("Upcoming Tasks") }}
+                <span class="sr-only">{{ _("(current)") }}</span></a>
+            </li>
+        </ol>
+    </nav>
+{% endblock %}
+
+{% block column %}
+    <h3> {{ _("Upcoming tasks") }}</h3>
+    <div class="alert alert-warning" role="alert">
+        {{ _("This page lists the coming tasks ordered on deadlines for courses you are registered in.") }}
+    </div>
+    <div class="list-group mb-3">
+        <a class="list-group-item list-group-item-action list-group-item-info" href="{{ get_homepath() }}/courselist">
+            <i class="fa fa-fw fa-th-list"></i>
+            {{ _("Course list") }}
+        </a>
+    </div>
+
+    <h3>{{ _("Last tried exercises") }}</h3>
+    <div class="list-group">
+        {% if submissions %}
+            {% for submission in submissions %}
+                <a class="list-group-item list-group-item-action
+                    {% if submission['status'] == 'done' and submission['result'] == 'success' %}
+                        list-group-item-success
+                    {% elif submission['status'] == 'done' and submission['result'] == 'save' %}
+                        list-group-item-info
+                    {% elif submission['status'] == 'waiting' %}
+                        list-group-item-warning
+                    {% else %}
+                        list-group-item-danger
+                    {% endif %}"
+                    href="{{ get_homepath() }}/course/{{submission['courseid']}}/{{submission['taskid']}}">
+                    <b>{{ submission["task"].get_course().get_name(user_manager.session_language()) }}</b>: {{ submission["task"].get_name(user_manager.session_language()) }}
+                </a>
+            {% endfor %}
+        {% else %}
+            <a class="list-group-item list-group-item-action disabled submission-empty">{{ _("No submissions") }}</a>
+        {% endif %}
+    </div>
+    {{ template_helper.call('main_menu', template_helper=template_helper) | safe }}
+{% endblock %}
+
+
+{% block content %}
+<h2>{{ _("My Upcoming Tasks") }}
+<div class="pull-right">
+        <div class="btn-group btn-group-sm">
+            <button class="btn btn-warning" data-toggle="modal" data-target="#timeModal">
+                <i class="fa fa-exchange"></i> {{_("Switch time planner")}}
+            </button>
+        </div>
+    </div>
+</h2>
+
+<div class="modal fade" id="timeModal" tabindex="-1" role="dialog" aria-labelledby="timeModalLabel" aria-hidden="true">
+    <div class="modal-dialog" role="document">
+        <form method="post">
+            <div class="modal-content">
+                <div class="modal-header">
+                    <h5 class="modal-title" id="timeModalLabel">{{_("Modify time planner")}}</h5>
+                    <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+                        <span aria-hidden="true">&times;</span>
+                    </button>
+                </div>
+                <div class="modal-body">
+                    <p>{{ _("This allows to display only the tasks whose deadline is in a certain temporal proximity.") }}</p>
+                    <label>{{ _("Number of days:") }}</label>
+                    <select class="form-control" id="new_time_planner" name="time_planner">
+                        {% for timeid in time_planner %}
+                            <option value="{{timeid}}">{{ timeid}}</option>
+                        {% endfor %}
+                    </select>
+                </div>
+                <div class="modal-footer">
+                    <button type="button" class="btn btn-secondary" data-dismiss="modal">{{_("Cancel")}}</button>
+                    <button type="submit" class="btn btn-warning">{{_("Save changes")}}</button>
+                </div>
+            </div>
+        </form>
+    </div>
+</div>
+
+<div class="list-group">
+    {%if open_courses %}
+        {% set username = user_manager.session_username() %}
+        {% for courseid, course in open_courses.items() %}
+                <div class="row">
+                    <div class="col-xs-12 col-md-8">
+                        {{ course.get_name(user_manager.session_language()) }}
+                        {% with course=course, tasks_data=tasks_data[courseid] %}
+                            {% include "upcoming.html" %}
+                        {% endwith %}
+                    </div>
+                </div>
+        {% endfor %}
+    {% else %}
+        <a href="#register" class="list-group-item list-group-item-action disabled">{{ _("You have no upcoming tasks") }}</a>
+    {% endif %}
+</div>
+{% endblock %}
--- a/inginious/frontend/plugins/upcoming_tasks/templates/upcoming.html
+++ b/inginious/frontend/plugins/upcoming_tasks/templates/upcoming.html
@@ -1,0 +1,9 @@
+{# This file is part of INGInious. See the LICENSE and the COPYRIGHTS files for #}
+{# more information about the licensing of this file. #}
+
+{# Recursively print a list of sections and their content #}
+
+{% set level = level|default(3, true) %}
+{% with tasks=sorting_method(course, tasks_data.keys()) %}
+	{% include "upcoming_task_list.html" %}
+{% endwith %}
--- a/inginious/frontend/plugins/upcoming_tasks/templates/upcoming_task_list.html
+++ b/inginious/frontend/plugins/upcoming_tasks/templates/upcoming_task_list.html
@@ -1,0 +1,51 @@
+{# This file is part of INGInious. See the LICENSE and the COPYRIGHTS files for #}
+{# more information about the licensing of this file. #}
+
+{% set registered = user_manager.course_is_user_registered(course) %}
+
+<div id="course" class="section tasks_list card mb-4">
+
+    <div class="content list-group list-group-flush tasks-list">
+        {% for task in tasks %}
+            {% set taskid = task.get_id() %}
+            {% if taskid in tasks_data %}
+                {% set succeeded = tasks_data[taskid]["succeeded"] %}
+                {% set completion = tasks_data[taskid]["grade"] %}
+                <a href="{{ get_homepath() }}/course/{{course.get_id()}}/{{taskid}}" class="list-group-item list-group-item-action
+                    {% if not task.get_accessible_time().is_open() %} disabled {% endif %}">
+                    <div class="row">
+                        <div class="col-xs-12 col-md-7 pl-1">
+                            <i style="color: {% if succeeded %}#468847{% else %}transparent{% endif %};" class="fa fa-check"></i>
+                            {{ task.get_name(user_manager.session_language()) }}
+                            {% if not task.get_accessible_time().is_open() %}
+                                - <b> {{ _("deadline reached") }} </b>
+                            {% endif %}
+                            </div>
+                        <div class="col-xs-12 col-md-5">
+                            {{ "DEADLINE: "}}
+                            <b> {{ task.get_deadline() }} </b>
+                        </div>
+                        <div class="col-xs-12 col-md-5">
+                            {% if registered %}
+                                <div class="progress">
+                                    <div class="progress-bar bg-success" aria-valuenow="{{ completion | int }}" aria-valuemin="0" aria-valuemax="100" style="width: {{ completion | int }}%">
+                                        {% if completion.is_integer() %}{{ completion | int }}{% else %}{{ completion }}{% endif %} %
+                                    </div>
+                                </div>
+                            {% endif %}
+                        </div>
+                    </div>
+					
+                    <div id="tags_of_task" style="display:none;">
+                        {% for category in task.get_categories() %}
+                            {% set tag = tag_list[category] %}
+                            {% if tag.is_visible_for_student() or user_manager.has_staff_rights_on_course(course) %}
+                                <div id="tag" data-tag-name="{{tag.get_id()}}"></div>
+                            {% endif %}
+                        {% endfor %}
+                    </div>
+                </a>
+            {% endif %}
+        {% endfor %}
+    </div>
+</div>
--- a/doc/install_doc/config_reference.rst
+++ b/doc/install_doc/config_reference.rst
@@ -369,6 +369,19 @@ A new configuration page named *Contest* appears on the administration page. To 
 *Enable contest plugin* box on the appropriate course. Please note that the plugin will override the task
 accessibility dates.
 
+
+Upcoming tasks plugin
+`````````````````````
+
+This plugin allows for students to easily visualize their upcoming tasks based on deadlines.
+The new page is available directly from the left main menu.
+
+To enable the plugin, add to your configuration file:
+::
+
+    plugins:
+        - plugin_module: inginious.frontend.plugins.upcoming_tasks
+
 Simple grader plugin
 ````````````````````
 
--- a/doc/teacher_doc/course_admin.rst
+++ b/doc/teacher_doc/course_admin.rst
@@ -76,7 +76,7 @@ When editing a task, you can enter basic informations and parameters in the *Bas
 
 Based on the type of problem you want to put for the task, you can select one of the two available *grading environment* in the *Environment* tab:
 
-- Select **Multiple Choice Question solver** if you only want to add *mcq* or *match* types of problems.
+- Select **Multiple Choice Question solver** if you only want to add *mcq* or *match* types of problems. Note, the *math* problem type from the problems-math plugin also uses this grading environment.
 - Select **Docker container** if you want to add some more complex problems which requires to write a grading script to access the students inputs.
 
 Adding/removing problems
--- a/inginious/frontend/plugins/upcoming_tasks/templates/main_menu.html
+++ b/inginious/frontend/plugins/upcoming_tasks/templates/main_menu.html
@@ -1,0 +1,3 @@
+<div class="list-group mb-3">
+    <a class="list-group-item list-group-item-action list-group-item-info" href="{{get_homepath()}}/coming_tasks"><i class="fa fa-rocket"></i> Upcoming Tasks</a>
+</div>


inginious/frontend/plugins/upcoming_tasks/__init__.py
```diff
@@ -0,0 +1,172 @@
...
+    """ called when modifying time planner"""
+    def POST_AUTH(self): 
+        user_input = flask.request.form
+        time_planner = "unlimited"
+        if "time_planner" in user_input:
+            time_planner = user_input.get("time_planner")
+        return self.page(time_planner)
+
+    """Used to convert the time_planner options into int value"""
+    def time_planner_converstion(self, string_time_planner):
```
anthonygego, line 62: *conversion

inginious/frontend/plugins/upcoming_tasks/__init__.py
```diff
@@ -0,0 +1,172 @@
...
+            new_user_task_list = course.get_task_dispenser().get_user_task_list([username])[username]
+            tasks_data.update({taskid: {"succeeded": False, "grade": 0.0} for taskid in new_user_task_list})
+
+            user_tasks = self.database.user_tasks.find({"username": username, "courseid": course.get_id(), "taskid": {"$in": new_user_task_list}})
+            for user_task in user_tasks:
+                tasks_data[user_task["taskid"]]["succeeded"] = user_task["succeeded"]
+                tasks_data[user_task["taskid"]]["grade"] = user_task["grade"]
+                if ( (tasks_data[user_task["taskid"]]["grade"] == 100.0)): 
+                    tasks_data.pop(user_task["taskid"])
+            "Remove outdated tasks and courses with no unfinished available tasks with deadline"""
```
anthonygego, line 113: I think this comment is badly formatted

inginious/frontend/plugins/upcoming_tasks/__init__.py
```diff
@@ -0,0 +1,172 @@
...
+                if (the_task.get_accessible_time().is_open()==False or ( (the_task.get_accessible_time().get_soft_end_date()) > (datetime.now()+timedelta(days=time_planner)) )): 
+                    outdated_tasks += [task]
+            new_user_task_list = course.get_task_dispenser().get_user_task_list([username])[username]
+            tasks_data.update({taskid: {"succeeded": False, "grade": 0.0} for taskid in new_user_task_list})
+
+            user_tasks = self.database.user_tasks.find({"username": username, "courseid": course.get_id(), "taskid": {"$in": new_user_task_list}})
+            for user_task in user_tasks:
+                tasks_data[user_task["taskid"]]["succeeded"] = user_task["succeeded"]
+                tasks_data[user_task["taskid"]]["grade"] = user_task["grade"]
+                if ( (tasks_data[user_task["taskid"]]["grade"] == 100.0)):
```
anthonygego, line 111: You should rather use the ``succeeded`` flag here. This is why INGInious explicitely makes the difference between.

inginious/frontend/plugins/upcoming_tasks/__init__.py
```diff
@@ -0,0 +1,172 @@
...
+                tasks_data[user_task["taskid"]]["succeeded"] = user_task["succeeded"]
+                tasks_data[user_task["taskid"]]["grade"] = user_task["grade"]
+                if ( (tasks_data[user_task["taskid"]]["grade"] == 100.0)): 
+                    tasks_data.pop(user_task["taskid"])
+            "Remove outdated tasks and courses with no unfinished available tasks with deadline"""
+            for outdated_task in outdated_tasks:
+                if outdated_task in tasks_data:
+                    tasks_data.pop(outdated_task)
+            if (not any(task in tasks for task in tasks_data)):
+                open_courses.pop(course.get_id())
```
anthonygego, line 118: There are multiple dictionary ``pop`` usage here. If you just want to remove the element from the dictionary, you can use ``del``, this is easier to read.

inginious/frontend/plugins/upcoming_tasks/__init__.py
```diff
@@ -0,0 +1,172 @@
...
+                                           open_courses=open_courses,
+                                           tasks_data=tasks_data,
+                                           tasks_list = list(tasks_data.keys()),
+                                           my_render=my_render,
+                                           time_planner=time_planner,
+                                           submissions=except_free_last_submissions)
+
+
+"""Class representing a render which is sent to the html file where it is recursively called with jinja"""
+class Render_Ordered:
```
anthonygego, line 138: ``RenderOrdered`` is more appropriate.

I do not understand well why you need to define this class as you only use it once, and the rendering method does not change.

You can simply use ``{% include %}`` jinja statements.

inginious/frontend/plugins/upcoming_tasks/templates/coming_tasks.html
```diff
@@ -0,0 +1,134 @@
...
+                    </div>
+                    <div class="col-xs-12 col-md-4">
+                        <i class="pull-right">
+                            {% if course.is_lti() %}
+                                {{ _("LTI course") }} -
+                            {% elif not course.is_open_to_non_staff() %}
+                                {{ _("Hidden course") }} -
+                            {% endif %}
+
+                            {% if username in course.get_admins() or user_manager.user_is_superadmin() %}
```
anthonygego, line 116: Not sure those are relevant information here.

inginious/frontend/plugins/upcoming_tasks/templates/coming_tasks.html
```diff
@@ -0,0 +1,134 @@
...
+                    <div class="col-xs-12 col-md-8">
+					{{ course.get_name(user_manager.session_language()) }}
+					{{ my_render.render_upcoming_list(template_helper, course, tasks_data, tasks_list, my_render) | safe }}
+
+
+                    </div>
+                    <div class="col-xs-12 col-md-4">
+                        <i class="pull-right">
+                            {% if course.is_lti() %}
+                                {{ _("LTI course") }} -
```
anthonygego, line 111: If LTI deadlines are displayed, a link to the external platform should be given instead. However, deadlines can and should rather be set at the external platform side. There may also be confusion as the course structure may be different between the two platforms. It can stay here, but I think it has a limited interest.

inginious/frontend/plugins/upcoming_tasks/templates/upcoming_task_list.html
```diff
@@ -0,0 +1,53 @@
...
+
+<div id="course" class="section tasks_list card mb-4">
+
+    <div class="content list-group list-group-flush tasks-list">
+        {% for task in tasks %}
+            {% set taskid = task.get_id() %}
+            {% if taskid in tasks_data %}
+                {% set succeeded = tasks_data[taskid]["succeeded"] %}
+                {% set completion = tasks_data[taskid]["grade"] %}
+                <a href="{{ get_homepath() }}/course/{{course.get_id()}}/{{taskid}}" class="list-group-item list-group-item-action
```
anthonygego, line 14: As I suggested before, the external platform link should be used for LTI exercises as they should be solved from it.

inginious/frontend/plugins/upcoming_tasks/__init__.py
darkyne: Thanks, fixed.

inginious/frontend/plugins/upcoming_tasks/__init__.py
darkyne: Indeed, thanks.

inginious/frontend/plugins/upcoming_tasks/__init__.py
darkyne: Thanks, especially since it's a boolean value...

inginious/frontend/plugins/upcoming_tasks/__init__.py
darkyne: Fine, I use del now instead of pop(). Not sure what's the real difference but at least it does not return the unused removed element now.

inginious/frontend/plugins/upcoming_tasks/__init__.py
darkyne: Name changed.

Actually I probably created this class to counter my inexperience in Jinja...
I had difficulties to correctly setup the path for the html file (for recursion) within Jinja (PATH_TO_PLUGIN is defined in the python file and calling os.path.abspath(os.path.dirname(__file__)) inside Jinja is probably not a good practice either (plus it would require to import os). 
The second reason was also to easily call the specific order() method which is in the class. 
By creating a class, I was able to send an object containing the sorting method together with the plugin_path...
Probably not the bect practice but it was the easier way to implement this for me.

If you believe this is a real issue, let me know.

inginious/frontend/plugins/upcoming_tasks/templates/coming_tasks.html
darkyne: Removed. Indeed this page is mainly useful for students. Showing their status in regard to the course (student / administrator / tutor) is not really relevant.

inginious/frontend/plugins/upcoming_tasks/templates/coming_tasks.html
darkyne: I removed this part together with the fact showing ("Hidden course" on the right side) since anyway courses are filtered before being sent to the html and this scenario will never show up.
This pull-right col was simply copy pased from another html file but had actually no purpose here.

inginious/frontend/plugins/upcoming_tasks/templates/upcoming_task_list.html
darkyne: I'm not sure about this one. In my opinion, this plugin is useful for exercises available directly from INGInious without using other plateforms. Plateforms using INGInious with integration will not push their students to access this page which is available only within INGInious anyway. I mean I don't see what LTI things have to do with this plugin.
The href is just to redirect the student to the actual task, like it is done on the tasks page of a course.

Don't hesitate to tell me if you see things differently.

anthonygego: Therefore LTI course should not be displayed. Either LTI courses are displayed with  the external URL, either there are not. But the usual link won't work for the students if it is a LTI course.

inginious/frontend/plugins/upcoming_tasks/__init__.py
anthonygego: Not an issue but I'm pretty sure this could be done in a easier way. It's ok as it is for me.

inginious/frontend/plugins/upcoming_tasks/templates/upcoming_task_list.html
darkyne: I'm not sure how to correctly handle this. 
Do you suggest to filter the courses so that LTI courses are removed from this page?
If I'm not wrong, LTI is just an option to enable in the couse settings... So this would mean, courses with this option enabled would not be shown anymore on the upcoming tasks page...

I don't know exactly how the LTI things work. I mean, aren't they simple courses with tasks in INGInious etc... But with the option to display the task inside an other website...?

inginious/frontend/plugins/upcoming_tasks/__init__.py
darkyne: Fine, I leave it like this even if it could probably be improved a bit.

inginious/frontend/plugins/upcoming_tasks/templates/upcoming_task_list.html
darkyne: Damn, you are right, LTI is a bit more complexe than I thought.
I will filter the courses so LTI courses are not displayed on this page.
Anyway student from LTI courses are probably not gonna ever use this plugin.

inginious/frontend/plugins/upcoming_tasks/__init__.py
```diff
@@ -0,0 +1,165 @@
+#!/usr/bin/env python
```
anthonygego, line 1: No need for shebang, this is not an executable file. The license header should be the same as for other files:

    # -*- coding: utf-8 -*-
    #
    # This file is part of INGInious. See the LICENSE and the COPYRIGHTS files for
    # more information about the licensing of this file.
    #
 
 You can still mention your name in the file as contributor afterwards , but the CLA (and, in your case, your employee contract :D) requires you to cede the copyright to UCLouvain, so this has to be mentioned.

inginious/frontend/plugins/upcoming_tasks/templates/upcoming_task_list.html
anthonygego: They can use the INGInious platform as well as an external platform. OK if filtered.

inginious/frontend/plugins/upcoming_tasks/__init__.py
darkyne: Well I believe I copy pasted the header from the auto-evaluation plugin actually... That I used as a base for the python file.
I replace it with what you just sent :)

inginious/frontend/plugins/upcoming_tasks/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+                    del tasks_data[user_task["taskid"]]
+            """ Remove outdated tasks and LTi courses or courses with no unfinished available tasks with deadline """
+            for outdated_task in outdated_tasks:
+                if outdated_task in tasks_data:
+                    del tasks_data[outdated_task]
+            if ((not any(task in tasks for task in tasks_data)) or course.is_lti()):
+                del open_courses[course.get_id()]
+        my_render=RenderOrdered(username)
+        time_planner = ["7", "14", "30", "unlimited"]
+        tasks_data_keys = list(tasks_data.keys())
```
anthonygego, line 104: What happens if you have tasks with same taskid in different courses ?

darkyne: Bad things...
Currently resolving this issue.

darkyne: Now tasks are filtered based on courseid and taskid.
Having multiple courses using same tasks id is not an issue anymore.

inginious/frontend/plugins/upcoming_tasks/templates/upcoming_task_list.html
```diff
@@ -0,0 +1,53 @@
...
+                                    <div class="progress-bar bg-success" aria-valuenow="{{ completion | int }}" aria-valuemin="0" aria-valuemax="100" style="width: {{ completion | int }}%">
+                                        {% if completion.is_integer() %}{{ completion | int }}{% else %}{{ completion }}{% endif %} %
+                                    </div>
+                                </div>
+                            {% endif %}
+							
+                        </div>
+                    </div>
+					
+					<div class="col-xs-12 col-md-5">
```
anthonygego, line 36: There tabs instead of spaces here. I'm also wondering why this block alone outside of a ``row`` block. I think the content of this block should be put in the block at line 17 : https://github.com/UCL-INGI/INGInious/pull/655/files#diff-48e9489117a9d34504271681302813d0e5d50ea01158dc6a9209054e2e7da72eR17

darkyne: Well sounds obvious I'm not very used to html haha :D
I've put the block in the previous row block as you advised.
I can't see any tabs from my pycharm interface... ?
It's looking like this on the page now:

darkyne:
Are there still things I may be able to improve here?

darkyne:
Are they still things I may be able to improve here? :)

inginious/frontend/plugins/upcoming_tasks/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+        all_courses = self.course_factory.get_all_courses()
+        time_planner = self.time_planner_conversion(time_planner)
+
+        # Get the courses id
+        open_courses = {courseid: course for courseid, course in all_courses.items()
+                        if self.user_manager.course_is_open_to_user(course, username, False) and
+                        self.user_manager.course_is_user_registered(course, username)}
+        open_courses = OrderedDict(sorted(iter(open_courses.items()), key=lambda x: x[1].get_name(self.user_manager.session_language())))
+
+        courses = {self.course_factory.get_course(courseid): course for courseid, course in all_courses.items()
```
anthonygego, line 66: Here you're using the Course object as dictionary key. I'd avoid that. In worst case it computes the hash based on a reference that can change, in best case it computes it somehow magically based on the complex data structure. 

I'm not even sure you need a dictionary because you don't seem to use key based setters and getters anywhere in the code.

Moreover, this line looks very similar to the ``open_courses`` one.

inginious/frontend/plugins/upcoming_tasks/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+        for submission in last_submissions:
+            try:
+                submission["task"] = open_courses[submission['courseid']].get_task(submission['taskid'])
+                except_free_last_submissions.append(submission)
+            except:
+                pass
+
+        # Get the courses tasks, remove finished ones and courses that have no available unfinished tasks with upcoming deadline in range
+        tasks_data = {}
+        outdated_tasks= {}
```
anthonygego, line 82: You don't seem to use this variable outside of the for loop. Therefore this is useless to keep all the data in a dictionary.

inginious/frontend/plugins/upcoming_tasks/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+        for course in courses:
+            courseid = course.get_id()
+            outdated_tasks[courseid] = []
+            tasks_data[courseid] = {}
+            tasks = course.get_tasks()
+            for task in tasks:
+                the_task = course.get_task(task)
```
anthonygego, line 89: ``course.get_tasks()`` already returns a dictionary with taskid as key and the task object as item.

inginious/frontend/plugins/upcoming_tasks/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+                if (not the_task.get_accessible_time().is_open()) or ((the_task.get_accessible_time().get_soft_end_date()) > (datetime.now()+timedelta(days=time_planner))):
+                    outdated_tasks[courseid] += [task]
+            new_user_task_list = course.get_task_dispenser().get_user_task_list([username])[username]
+            tasks_data[courseid].update({taskid: {"succeeded": False, "grade": 0.0} for taskid in new_user_task_list})
```
anthonygego, line 93: As this is the first time you're filling the dict, you can simply define it with a unique comprehension instead of updating it.

inginious/frontend/plugins/upcoming_tasks/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+            user_tasks = self.database.user_tasks.find({"username": username, "courseid": course.get_id(), "taskid": {"$in": new_user_task_list}})
+            for user_task in user_tasks:
+                tasks_data[courseid][user_task["taskid"]]["succeeded"] = user_task["succeeded"]
+                tasks_data[courseid][user_task["taskid"]]["grade"] = user_task["grade"]
+                if tasks_data[courseid][user_task["taskid"]]["succeeded"]:
```
anthonygego, line 98: If you put this check right after the for statement, you won't need to upgrade the content of a deleted item.

inginious/frontend/plugins/upcoming_tasks/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+                    del tasks_data[courseid][user_task["taskid"]]
+            # Remove outdated tasks and LTi courses or courses with no unfinished available tasks with deadline
+            for outdated_task in outdated_tasks[courseid]:
+                if outdated_task in tasks_data[courseid]:
+                    del tasks_data[courseid][outdated_task]
```
anthonygego, line 103: A list comprehension with a filter is probably more elegant and efficient than those three lines.

inginious/frontend/plugins/upcoming_tasks/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+                                      tasks_data=tasks_data, render=render)
+
+def sort_by_deadline(course, user_urgent_task_list):
+    """ Given a course (object) and a list of user urgent tasksid,
+    returns the list of urgent tasks (objects) for that course ordered based on deadline """
+    course_tasks = course.get_tasks()
+    course_user_urgent_task_list = list(set(course_tasks).intersection(user_urgent_task_list))
+    ordered_tasks = sorted(course_user_urgent_task_list, key=lambda x: course.get_task(x).get_accessible_time().get_soft_end_date())
+    ordered_tasks = map(course.get_task, ordered_tasks)
+    return list(ordered_tasks)
```
anthonygego, line 148: You can probably replace the two last lines by:

    return [course_tasks[taskid] for taskid in ordered_tasks]

inginious/frontend/plugins/upcoming_tasks/templates/coming_tasks.html
```diff
@@ -0,0 +1,126 @@
...
+                </div>
+            </a>
+        {% endfor %}
+    {% else %}
+        <a href="#register" class="list-group-item list-group-item-action disabled">{{ _("You have no upcoming tasks") }}</a>
+    {% endif %}
+</div>
+
+{% endblock %}
+
```
anthonygego, line 119: Too many blank lines here.

anthonygego (changes requested):

inginious/frontend/plugins/upcoming_tasks/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+
+    def GET_AUTH(self):
+        """ Called when reaching the page """
+        time_planner = "unlimited"
+        return self.page(time_planner)
+
+    def POST_AUTH(self):
+        """ Called when modifying time planner """
+        user_input = flask.request.form
+        time_planner = "unlimited"
```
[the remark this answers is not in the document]
darkyne: Actually I noticed, simply using time_planner = user_input.get("time_planner") is enough since it's in the POST and we only enter here if the user opens the time_planer modal then validates it, so there is 100% always a selected value (the cancel button does not proc the POST).
This removes these 3 lines:

         time_planner = "unlimited"
         if "time_planner" in user_input:
               time_planner = user_input.get("time_planner", "unlimited")

into just: 

     time_planner = user_input.get("time_planner")

inginious/frontend/plugins/upcoming_tasks/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+        if "time_planner" in user_input:
+            time_planner = user_input.get("time_planner")
+        return self.page(time_planner)
+
+    def time_planner_conversion(self, string_time_planner):
+        """ Used to convert the time_planner options into int value """
+        if string_time_planner == "unlimited":
+            return 100000
+        return int(string_time_planner)
```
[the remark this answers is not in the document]
darkyne: Here is the related form:
image

So the value is always "7", "14", "30" or "unlimited".
In other words, there is 100% no bug here in the current version but indeed if it was to be adapted or to allow any value then it might require some checks.

Do you want me to change this or should I let it be like this ?

inginious/frontend/plugins/upcoming_tasks/__init__.py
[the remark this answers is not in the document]
anthonygego: You can still receive badly formatted or malicious requests, so you must have a fallback for all unexpected inputs.

inginious/frontend/plugins/upcoming_tasks/__init__.py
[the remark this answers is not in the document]
anthonygego: I can perform a post request that will raise an exception.

inginious/frontend/plugins/upcoming_tasks/__init__.py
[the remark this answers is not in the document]
darkyne: oh, ok. I didn't think about requests not coming from the user interface actually.
From the line you suggested, I suppose you meant the "unlimited" to act as a default value but I'm a bit surprised that the get method from the ImmutableMultiDict (the form) acts like that.

inginious/frontend/plugins/upcoming_tasks/__init__.py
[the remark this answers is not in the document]
darkyne: Well, clearly I didn't take into account manual requests.
Right now, they would raise an exception how do you suggest to handle this?
I thought it they raised an exception it would actually be a good thing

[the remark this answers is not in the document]
darkyne: Is this better ?    

def time_planner_conversion(self, string_time_planner):
        """ Used to convert the time_planner options into int value """
        if string_time_planner in ["7", "14", "30"]:
            return int(string_time_planner)
        return 100000

inginious/frontend/plugins/upcoming_tasks/__init__.py
darkyne: Indeed.

inginious/frontend/plugins/upcoming_tasks/__init__.py
darkyne: Indeed, it was actually possible to manage everything only with open_courses without requiring a second dictonnary + it was, as you mentioned, using course as key instead of courseid.

The only little limitation I had with a single dictionnary is that I can not iterate on it with a for, while deleting items from it in the loop itself but I simply deleted items after the iterations and it's fine now.

inginious/frontend/plugins/upcoming_tasks/__init__.py
darkyne: You are right.

inginious/frontend/plugins/upcoming_tasks/__init__.py
darkyne: Now using taskid and task from the dictionnary without calling course.get_task everytime.

inginious/frontend/plugins/upcoming_tasks/__init__.py
darkyne: Indeed, works fine.

inginious/frontend/plugins/upcoming_tasks/__init__.py
darkyne: Is this line better ?
```python
tasks_data[courseid] = {k: v for k, v in tasks_data[courseid].items() if k not in outdated_tasks}
```

inginious/frontend/plugins/upcoming_tasks/__init__.py
```diff
@@ -0,0 +1,146 @@
...
+            tasks = course.get_tasks()
+            for taskid in tasks:
+                the_task = tasks[taskid]
+                if (not the_task.get_accessible_time().is_open()) or ((the_task.get_accessible_time().get_soft_end_date()) > (datetime.now()+timedelta(days=time_planner))):
+                    outdated_tasks += [taskid]
+            new_user_task_list = course.get_task_dispenser().get_user_task_list([username])[username]
+            tasks_data[courseid] = {taskid: {"succeeded": False, "grade": 0.0} for taskid in new_user_task_list}
+            user_tasks = self.database.user_tasks.find({"username": username, "courseid": course.get_id(), "taskid": {"$in": new_user_task_list}})
+            for user_task in user_tasks:
+                if not tasks_data[courseid][user_task["taskid"]]["succeeded"]:
```
anthonygego, line 90: You don"t need to remove the succeeded ones anymore from the list ?

darkyne: I don't put the information in the tasks_data if it is succeeded.
Before I was putting information in tasks_data for all the tasks then removing the succeeded one, now I only add the non-succeeded ones :)

darkyne: Well, you are right, there is an issue here:
I am checking that tasks_data... is not succeed before even defining its content.
I should check based on user_data not on tasks_data

darkyne: I believe going back to the initial solution for this particular part:
```python
            for user_task in user_tasks:
                tasks_data[courseid][user_task["taskid"]]["succeeded"] = user_task["succeeded"]
                tasks_data[courseid][user_task["taskid"]]["grade"] = user_task["grade"]
                if tasks_data[courseid][user_task["taskid"]]["succeeded"]:
                    del tasks_data[courseid][user_task["taskid"]]
```python

might end up being the most simple solution.
To avoid setting grade and succeed the code would be:

```python
            for user_task in user_tasks:
                if not user_task["succeeded"]:
                    tasks_data[courseid][user_task["taskid"]]["succeeded"] = user_task["succeeded"]
                    tasks_data[courseid][user_task["taskid"]]["grade"] = user_task["grade"]
                else:
                    del tasks_data[courseid][user_task["taskid"]]
```

To me, it's as you prefer.

darkyne: well, you're right, the second option is probably better :)

inginious/frontend/plugins/upcoming_tasks/__init__.py
```diff
@@ -0,0 +1,148 @@
...
+            tasks_data[courseid] = {taskid: {"succeeded": False, "grade": 0.0} for taskid in new_user_task_list}
+            user_tasks = self.database.user_tasks.find({"username": username, "courseid": course.get_id(), "taskid": {"$in": new_user_task_list}})
+            for user_task in user_tasks:
+                if not user_task["succeeded"]:
+                    tasks_data[courseid][user_task["taskid"]]["succeeded"] = user_task["succeeded"]
+                    tasks_data[courseid][user_task["taskid"]]["grade"] = user_task["grade"]
+                else:
+                    del tasks_data[courseid][user_task["taskid"]]
+            # Remove outdated tasks and LTi courses or courses with no unfinished available tasks with deadline
+            tasks_data[courseid] = {k: v for k, v in tasks_data[courseid].items() if k not in outdated_tasks}
```
anthonygego, line 96: It would probably be more efficient and clean to filter the ``outdated_tasks`` directly when ``new_user_task_list`` is defined, to avoid creating entries for them an searching for them in database.

inginious/frontend/plugins/upcoming_tasks/__init__.py
```diff
@@ -0,0 +1,148 @@
...
+            try:
+                submission["task"] = open_courses[submission['courseid']].get_task(submission['taskid'])
+                except_free_last_submissions.append(submission)
+            except:
+                pass
+
+        # Get the courses tasks, remove finished ones and courses that have no available unfinished tasks with upcoming deadline in range
+        tasks_data = {}
+        succeeded_courses = []
+        for courseid in open_courses:
```
anthonygego, line 77: As this snippet is quite beefy, I'd rather have ``for courseid, course in open_courses.tems()`` too.

inginious/frontend/plugins/upcoming_tasks/__init__.py
```diff
@@ -0,0 +1,148 @@
...
+            course = open_courses[courseid]
+            outdated_tasks = []
+            tasks_data[courseid] = {}
+            tasks = course.get_tasks()
+            for taskid in tasks:
+                the_task = tasks[taskid]
+                if (not the_task.get_accessible_time().is_open()) or ((the_task.get_accessible_time().get_soft_end_date()) > (datetime.now()+timedelta(days=time_planner))):
+                    outdated_tasks += [taskid]
```
anthonygego, line 85: I'm pretty sur this can be done with a comprehension. Maybe put the whole if logic in a lambda to express it cleaner.

    is_outdated = lambda thetask: (not the_task....time_planner)))
    outdated_tasks = [taskid for taskid, the_task in tasks.items() if is_outdated(the_task)]

inginious/frontend/plugins/upcoming_tasks/__init__.py
darkyne: Now new_user_task_list is directly filtered to only take non-outdated tasks.
This mean, as you suggested, tasks_data only contains available tasks with deadline and only search in database for these.

inginious/frontend/plugins/upcoming_tasks/__init__.py
darkyne: Yep, nice little shortcut :)

inginious/frontend/plugins/upcoming_tasks/__init__.py
darkyne: Now done in a single line with list comprehension.

inginious/frontend/plugins/upcoming_tasks/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+                except_free_last_submissions.append(submission)
+            except:
+                pass
+
+        # Get the courses tasks, remove finished ones and courses that have no available unfinished tasks with upcoming deadline in range
+        tasks_data = {}
+        succeeded_courses = []
+        for courseid, course in open_courses.items():
+            tasks_data[courseid] = {}
+            tasks = course.get_tasks()
```
anthonygego, line 78: Those can be fetched outside of the for loop.

inginious/frontend/plugins/upcoming_tasks/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+                submission["task"] = open_courses[submission['courseid']].get_task(submission['taskid'])
+                except_free_last_submissions.append(submission)
+            except:
+                pass
+
+        # Get the courses tasks, remove finished ones and courses that have no available unfinished tasks with upcoming deadline in range
+        tasks_data = {}
+        succeeded_courses = []
+        for courseid, course in open_courses.items():
+            tasks_data[courseid] = {}
```
anthonygego, line 77: This line is useless

inginious/frontend/plugins/upcoming_tasks/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+            tasks_data[courseid] = {taskid: {"succeeded": False, "grade": 0.0} for taskid in new_user_task_list}
+            user_tasks = self.database.user_tasks.find({"username": username, "courseid": course.get_id(), "taskid": {"$in": new_user_task_list}})
+            for user_task in user_tasks:
+                if not user_task["succeeded"]:
+                    tasks_data[courseid][user_task["taskid"]]["succeeded"] = user_task["succeeded"]
+                    tasks_data[courseid][user_task["taskid"]]["grade"] = user_task["grade"]
+                else:
+                    del tasks_data[courseid][user_task["taskid"]]
+            # Remove courses with no unfinished available tasks with deadline and lti courses
+            if (not any(task in tasks for task in tasks_data[courseid])) or course.is_lti():
```
anthonygego, line 91: any(task in tasks for task in tasks_data[courseid])
    
   Are you actually checking for a non-zero length ?

anthonygego (commented):
Keep strong, it's getting better.

inginious/frontend/plugins/upcoming_tasks/__init__.py
darkyne: Removed

inginious/frontend/plugins/upcoming_tasks/__init__.py
darkyne: I'm not sure about what you mean...
I treat every course (course where the user is registered) one by one in the loop...
And I fetch the tasks to filter them there...
I don't get why you would want course.get_tasks() to be out the loop.
For example fetching all the tasks for all the courses all at once before ?
I believe it is simpler how it is right now, no ?

anthonygego: My bad, I had the typical page setting in mind.

inginious/frontend/plugins/upcoming_tasks/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+
+        my_render = RenderOrdered()
+        # Sort the courses based on the most urgent task for each course
+        open_courses = OrderedDict( sorted(iter(open_courses.items()), key=lambda x: (sort_by_deadline(x[1], tasks_data[x[0]].keys())[0]).get_accessible_time().get_soft_end_date() ))
+
+        return self.template_helper.render("coming_tasks.html",
+                                           template_folder=PATH_TO_PLUGIN + "/templates/",
+                                           open_courses=open_courses,
+                                           tasks_data=tasks_data,
+                                           my_render=my_render,
```
anthonygego, line 104: 1. Here you can pass the `sort_by_deadline` function in argument instead of `my_render`

inginious/frontend/plugins/upcoming_tasks/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+                                           time_planner=["7", "14", "30", "unlimited"],
+                                           submissions=except_free_last_submissions)
+
+
+class RenderOrdered:
```
anthonygego, line 109: 2. You can remove this whole class

inginious/frontend/plugins/upcoming_tasks/templates/coming_tasks.html
```diff
@@ -0,0 +1,111 @@
...
+</div>
+
+<div class="list-group">
+	{%if open_courses %}
+        {% set username = user_manager.session_username() %}
+        {% for courseid, course in open_courses.items() %}
+                <div class="row">
+                    <div class="col-xs-12 col-md-8">
+					{{ course.get_name(user_manager.session_language()) }}
+					{{ my_render.render_upcoming_list(template_helper, course, tasks_data[courseid], my_render) | safe }}
```
anthonygego, line 101: 3. Replace by:

        {% with course=course, tasks_data=tasks_data[courseid] %}
            {% include "upcoming.html" %}
        {% endwith %}

anthonygego (commented):
You should be able to remove the `RenderOrdered` class by following the steps in comments.

inginious/frontend/plugins/upcoming_tasks/templates/upcoming.html
```diff
@@ -0,0 +1,9 @@
+{# This file is part of INGInious. See the LICENSE and the COPYRIGHTS files for #}
+{# more information about the licensing of this file. #}
+
+{# Recursively print a list of sections and their content #}
+
+{% set level = level|default(3, true) %}
+{% with tasks=render.order(course, tasks_data.keys()) %}
```
anthonygego, line 7: 4. Replace by `sort_by_deadline`. The comment was not posted by github with the review...

anthonygego:
LGTM. Can you add a mention in the documentation ?

Issue #656, opened 2021-03-27
Suggestion : Remove Protocol from get_homepath functions 


pchaow:
I setup an inginious at http://local_ip:8000 by running inginious-webapp.
Then I setup nginx reverse proxy from https://public_domain and point to the inginious server.

All of static link files, such as CSS, JS, etc., it point to http://public_domain/...... 

I suggest to remove protocol from request.url_root in function get_homepath ( /inginious/frontend/app.py )

from http://x.x.x.x to //x.x.x.x

pchaow:
close using https://werkzeug.palletsprojects.com/en/1.0.x/middleware/proxy_fix/ instead

PR #657, opened 2021-03-27
Remove Protocol from get_homepath functions

pchaow:
Link to issue #656 

I create this pull request to show crude solution to remove protocol from get_homepath functions

### DIFF
--- a/inginious/frontend/app.py
+++ b/inginious/frontend/app.py
@@ -106,12 +106,16 @@ def get_homepath(ignore_session=False, force_cookieless=False):
     """
     session = flask.session
     request = flask.request
+
+
+
     if not ignore_session and session.sid is not None and session.cookieless:
         return request.url_root[:-1] + "/@" + session.sid + "@"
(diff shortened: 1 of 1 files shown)


Drumor:
Hello,

Thank u for this contribution. Could you squash your commits into one? Thx!

pchaow:
Ok, I’ll try another solution

anthonygego:
To be clearer, the application itself (and thus the wsgi context) should be aware of the protocol used by the client because several features rely on this : url generation (that should be replaced by the more idiomatic ``url_for``in the future btw), LTI and the social apps auth method that relies on OAuth 1 or 2. Those protocols use URLs to sign their messages.

Therefore the ProxyFix middleware seems to be the best solution to solve this issue properly. As it is not recommended to enable this by default, I suggest to include it as an opt-in in ``inginious-webapp`` file, and to fetch the function arguments via the configuration file.

pchaow:
Thanks, I've tried proxyfix and it's work for me. I'll add some configuration

PR #659, opened 2021-03-29
add reverse-proxy configuration

pchaow:
This is a crude solution for "X-Forwarded-For Proxy Fix" using using https://werkzeug.palletsprojects.com/en/1.0.x/middleware/proxy_fix/

### DIFF
--- a/inginious-webapp
+++ b/inginious-webapp
@@ -14,6 +14,7 @@ import signal
 import sys
 
 from werkzeug.serving import run_simple
+from werkzeug.middleware.proxy_fix import ProxyFix
 
 # Apache wsgi module only set env vars when a request is done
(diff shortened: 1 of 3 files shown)


inginious-webapp
```diff
@@ -89,4 +90,15 @@ if __name__ == "__main__":
 
     # Launch the app
     if __name__ == "__main__":
+
+        reverse_proxy_config = config.get('reverse-proxy-config',{})
```
anthonygego, line 94: Looks good, but I'd rather put all this before the if statement, so that this is also applied if the app is launched through a wsgi server such as uwsgi for instance.

anthonygego (commented):
Could you also add some description of the new parameters in : https://github.com/UCL-INGI/INGInious/blob/master/doc/install_doc/config_reference.rst ?

Thanks !

inginious-webapp
pchaow: Ok, Got it

PR #662, opened 2021-04-02
Fix LDAP Injection

jorgectf:
Properly encoding `login` while formatting `ldap_request` will avoid a LDAP Injection which could allow an attacker to leak information and even bypass the authorization measures.

https://github.com/UCL-INGI/INGInious/blob/4d1a46f1e1df8b608a7bba715faa0b86aa0c18c1/inginious/frontend/plugins/auth/ldap_auth.py#L101


anthonygego:
Wow, it has been there for a while ! Thanks !

jorgectf:
Thank you for the merge! :)

PR #663, opened 2021-04-14
Update task.html

Drumor:
Adding a return to course list shortcut with an arrow up representation.


anthonygego:
I'm not sure this is necessary is it complexifies the user interface by adding more buttons in this group. The breadcrumbs are not really useful if navigation is entirely done through buttons. Moreover, the tooltip misses a capitalized letter.

PR #670, opened 2021-04-26
task_id checker

darkyne:
This adds a checker on the task_id when creating a new task.

It is exactly based on the same code as the checker already implemented on the subproblem_id as you can see:

Code for subproblem task id:
image

Code for task_id:
image

When the id is not valid, an alert pops and the task is not created.

### DIFF
--- a/inginious/frontend/task_dispensers/util.py
+++ b/inginious/frontend/task_dispensers/util.py
@@ -21,7 +21,7 @@ class SectionsList(object):
         for section in sorted(structure,key=lambda k: k['rank']):
             if "sections_list" in section:
                 self._sections.append(NonTerminalSection(section))
-            elif "tasks_list" in section and all(id_checker(id) for id in list(section.get("tasks_list"))):
+            elif "tasks_list" in section:
                 self._sections.append(TerminalSection(section))
             else:
                 raise InvalidTocException(_("One section don't contain a sections list nor a tasks list"))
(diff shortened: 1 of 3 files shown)


inginious/frontend/task_dispensers/util.py
```diff
@@ -20,7 +21,7 @@ def __init__(self, structure):
         for section in sorted(structure,key=lambda k: k['rank']):
             if "sections_list" in section:
                 self._sections.append(NonTerminalSection(section))
-            elif "tasks_list" in section:
+            elif "tasks_list" in section and all(id_checker(id) for id in list(section.get("tasks_list"))):
```
anthonygego, line 24: I'd rather put this check in the ``TerminalSection`` code.

inginious/frontend/task_dispensers/util.py
```diff
@@ -50,6 +51,8 @@ def add_task(self, taskid, sectionid):
         :param sectionid: the section id of the section
         :return: True is the task has been added false otherwise
         """
+        if not id_checker(taskid):
```
anthonygego, line 54: Probably redundant here as it is implemented in ``TerminalSection.add_task``.

inginious/frontend/task_dispensers/util.py
```diff
@@ -137,6 +140,8 @@ def add_task(self, taskid, sectionid):
...
```
anthonygego, line 143: Probably redundant here as it is implemented in ``TerminalSection.add_task``.

anthonygego (approved):
LGTM

PR #680, opened 2021-06-03
ssh_student for Docker

darkyne:
This adds the ssh_student functionality for docker environment (not working with Kata).
Basically, when the grading_script hits the ssh_student function, it starts a student_container which will start a ssh server and send to the front-end the commands for the student to connect into the student_container. When the student leaves the ssh session, the grading_script continues to run.

A very simple example is the following:
image

Notes:
- If no one connects to the ssh server, it times out after 2min and the grading script continues.
- The files modified in the student directory can be used for evaluation in the grading container

Still to improve: (marked with "TODO" in the files)
- If no one connects, it might be better to send an error message (and not to continue the script)
- If the teacher configures the task with a very small time-out (for example less than 5 minutes), it might be useful to send a warning message to tell the teacher to put a longer time limit.
- The student-container (when launched via ssh_student) is using "bridge" as network mode. Maybe it is not the best configuration.
- There are some duplicated code between INGInious and _run_student_intern (setLimits, execute_process, launch sshd and wait for user). It should be possible to put this in a file (example utils.py) and to import it instead of duplications. I had a quick try but didn't work.
- Some communication between parts are sometimes not perfectly designed (but working), example [this](https://github.com/darkyne/INGInious/blob/aac99d98a255e04ae508b897bde2b565cc1508de/base-containers/base/inginious_container_api/run_student.py#L116).
-  Block this functionality if the teacher did not hit "allow internet access" in the task configuration. Indeed, the student_container to allow ssh fonctionalities would require a bridge network mode.
- Verify the behaviour if a student connects with ssh and kills the sshd process.

Technical explanations:
- "ssh_student" calls "run_student" with the parameter "ssh" set to True.
- The student_container is launched with _run_student_intern and receives a message which was sent [here](https://github.com/darkyne/INGInious/blob/aac99d98a255e04ae508b897bde2b565cc1508de/base-containers/base/inginious_container_api/run_student.py#L105). This was originaly the way how run_student sent the cmd to the student_container. Now it also sends the "ssh" boolean.
- When receiving this message, the student_container starts a ssh server only if the "ssh" boolean is True [(here)](https://github.com/darkyne/INGInious/blob/aac99d98a255e04ae508b897bde2b565cc1508de/base-containers/base/bin/_run_student_intern#L115). Otherwise, it runs the cmd as usually.
- After the student_container started the ssh server, it sends a message [(here)](https://github.com/darkyne/INGInious/blob/aac99d98a255e04ae508b897bde2b565cc1508de/base-containers/base/bin/_run_student_intern#L137) with id and password.
- This message is received by the grading_container [(here)](https://github.com/darkyne/INGInious/blob/aac99d98a255e04ae508b897bde2b565cc1508de/base-containers/base/inginious_container_api/run_student.py#L116) and sent to the agent [(here)](https://github.com/darkyne/INGInious/blob/aac99d98a255e04ae508b897bde2b565cc1508de/base-containers/base/inginious_container_api/run_student.py#L122).
- The agent received this message [(here)](https://github.com/darkyne/INGInious/blob/aac99d98a255e04ae508b897bde2b565cc1508de/inginious/agent/docker_agent/__init__.py#L576). It adds the correct external port to reach the concerned student_container (based on the container_id) and sends this to the front-end allowing the student to connect to the student_container.

Probably many little improvements to do but I hope no huge issue.

### DIFF
--- a/base-containers/base/bin/_run_student_intern
+++ b/base-containers/base/bin/_run_student_intern
@@ -14,8 +14,6 @@ import struct
 from inginious_container_api.utils import set_limits_user, start_ssh_server, ssh_wait
 
 
-
-
 def handle_signals(subprocess, socket):
     """ Handles signals given by run_student on the socket """
     while True:
@@ -64,6 +62,10 @@ while start_cmd is None:
 print("Unpacked start cmd")
 user = start_cmd["user"]  # Know if you are worker (on docker) or root (on Kata)
 
+if user=="root" and os.path.exists("/.__input/__shared_kernel"):
+    print("You can not run as root on docker")
+    exit(251)
+
 # Add some elements to /etc/hosts and /etc/resolv.conf if needed
 system_files = {"hosts": ("/etc/hosts", True), "resolv.conf": ("/etc/resolv.conf", False)}
 for name, (spath, append) in system_files.items():
@@ -103,6 +105,7 @@ if start_cmd["ssh"]:
     client.send(message)
     # Wait for user to connect and leave
     retval = ssh_wait(ssh_user)
+
     print("Exiting")
     exit(retval)
 else:
--- a/base-containers/base/bin/_run_student_intern
+++ b/base-containers/base/bin/_run_student_intern
@@ -10,6 +10,7 @@ import subprocess
 import threading
 import msgpack
 import array
+import struct
 from inginious_container_api.utils import set_limits_user, start_ssh_server, ssh_wait, User
 
 
@@ -98,6 +99,8 @@ if start_cmd["ssh"]:
     # Send ssh information
     ssh_user_bin = User[ssh_user].value  # 0 for root, 1 for worker
     message = msgpack.dumps({"type": "ssh_student", "ssh_user": ssh_user_bin, "password": password})  # constant size
+    message_size = struct.pack('I', len(message))
+    client.send(message_size)
     client.send(message)
     # Wait for user to connect and leave
     retval = ssh_wait(ssh_user)
--- a/base-containers/base/bin/_run_student_intern
+++ b/base-containers/base/bin/_run_student_intern
@@ -11,7 +11,7 @@ import threading
 import msgpack
 import array
 import struct
-from inginious_container_api.utils import set_limits_user, start_ssh_server, ssh_wait, User
+from inginious_container_api.utils import set_limits_user, start_ssh_server, ssh_wait
 
 
 
@@ -80,7 +80,7 @@ print("Popen")
 print(fds)
 
 set_limits = lambda: set_limits_user(user)  # To know if the command should be executed as root or worker
-if start_cmd["command"] is None or start_cmd["command"] == "":  # Avoid to ron an empty command
+if start_cmd["command"] is None or start_cmd["command"] == "":  # Avoid to run an empty command
     start_cmd["command"] = "echo 'info: student container started with no command set' "
 
 # Run the command
@@ -97,8 +97,7 @@ if start_cmd["ssh"]:
     # Start ssh server
     ssh_user, password = start_ssh_server(user)
     # Send ssh information
-    ssh_user_bin = User[ssh_user].value  # 0 for root, 1 for worker
-    message = msgpack.dumps({"type": "ssh_student", "ssh_user": ssh_user_bin, "password": password})  # constant size
+    message = msgpack.dumps({"type": "ssh_student", "ssh_user": ssh_user, "password": password})  # constant size
     message_size = struct.pack('I', len(message))
     client.send(message_size)
     client.send(message)
--- a/base-containers/base/bin/_run_student_intern
+++ b/base-containers/base/bin/_run_student_intern
@@ -98,7 +98,7 @@ if start_cmd["ssh"]:
     ssh_user, password = start_ssh_server(user)
     # Send ssh information
     message = msgpack.dumps({"type": "ssh_student", "ssh_user": ssh_user, "password": password})  # constant size
-    message_size = struct.pack('I', len(message))
+    message_size = struct.pack('!I', len(message))
     client.send(message_size)
     client.send(message)
     # Wait for user to connect and leave
--- a/base-containers/base/bin/_run_student_intern
+++ b/base-containers/base/bin/_run_student_intern
@@ -119,11 +119,12 @@ if start_cmd["ssh"]:
     # Wait for user to connect and leave
     retval = ssh_wait(ssh_user)
 
-if start_cmd["closure_script"] != "":
-    # Run the closure_script if there is one
-    p2 = subprocess.Popen(shlex.split(start_cmd["closure_script"]), preexec_fn=set_limits, stdin=fds[0], stdout=fds[1],
+if start_cmd["teardown_script"] != "":
+    # Run the teardown_script if there is one
+    p2 = subprocess.Popen(shlex.split(start_cmd["teardown_script"]), preexec_fn=set_limits, stdin=fds[0], stdout=fds[1],
                          stderr=fds[2])
-    p2.wait() #Waot for closure_script to finish before exiting the container
+    teardown_retval = p2.wait() #Waot for teardown_script to finish before exiting the container
+    print(teardown_retval)  # For debugging purpose
 
 print("Exiting")
 exit(retval)
--- a/base-containers/base/bin/_run_student_intern
+++ b/base-containers/base/bin/_run_student_intern
@@ -98,7 +98,7 @@ set_limits = lambda: set_limits_user(user)  # To know if the command should be e
 if start_cmd["command"] is None or start_cmd["command"] == "":  # Avoid to run an empty command
     start_cmd["command"] = "echo 'info: student container started with no command set' "
 
-# Run the command
+# Run the command, which in the case of ssh, acts like a setup_script
 p = subprocess.Popen(shlex.split(start_cmd["command"]), preexec_fn=set_limits, stdin=fds[0], stdout=fds[1], stderr=fds[2])
 # Handle the signals
 print("Signal thread")
@@ -119,8 +119,12 @@ if start_cmd["ssh"]:
     # Wait for user to connect and leave
     retval = ssh_wait(ssh_user)
 
-    print("Exiting")
-    exit(retval)
-else:
-    print("Exiting")
-    exit(retval)
+if start_cmd["closure_script"] != "":
+    # Run the closure_script if there is one
+    p2 = subprocess.Popen(shlex.split(start_cmd["closure_script"]), preexec_fn=set_limits, stdin=fds[0], stdout=fds[1],
+                         stderr=fds[2])
+    p2.wait() #Waot for closure_script to finish before exiting the container
+
+print("Exiting")
+exit(retval)
+
--- a/base-containers/base/bin/_run_student_intern
+++ b/base-containers/base/bin/_run_student_intern
@@ -8,11 +8,11 @@ import shlex
 import socket
 import subprocess
 import threading
-
 import msgpack
 import array
+from inginious_container_api.utils import set_limits_user, start_ssh_server, ssh_wait, User
 
-import resource
+
 
 
 def handle_signals(subprocess, socket):
@@ -24,6 +24,7 @@ def handle_signals(subprocess, socket):
         print("received %s" % signal.decode("utf8"))
         subprocess.send_signal(int(signal.decode('utf8')))
 
+
 def recv_fds(sock, msglen, maxfds):
     """ Receive FDs from the unix socket. Copy-pasted from the Python doc """
     fds = array.array("i")  # Array of ints
@@ -34,10 +35,6 @@ def recv_fds(sock, msglen, maxfds):
             fds.fromstring(cmsg_data[:len(cmsg_data) - (len(cmsg_data) % fds.itemsize)])
     return msg, list(fds)
 
-def setlimits():
-    os.setgid(4242)
-    os.setuid(4242)
-    resource.setrlimit(resource.RLIMIT_NPROC, (1000, 1000))
 
 # Connect to the socket
 client = socket.socket(socket.AF_UNIX)  # , socket.SOCK_CLOEXEC) # for linux only
@@ -64,6 +61,7 @@ while start_cmd is None:
     for obj in unpacker:
         start_cmd = obj
 print("Unpacked start cmd")
+user = start_cmd["user"]  # Know if you are worker (on docker) or root (on Kata)
 
 # Add some elements to /etc/hosts and /etc/resolv.conf if needed
 system_files = {"hosts": ("/etc/hosts", True), "resolv.conf": ("/etc/resolv.conf", False)}
@@ -79,17 +77,32 @@ print("Chdir")
 os.chdir(start_cmd["working_dir"])
 print("Popen")
 print(fds)
-p = subprocess.Popen(shlex.split(start_cmd["command"]), preexec_fn=setlimits, stdin=fds[0], stdout=fds[1], stderr=fds[2])
 
+set_limits = lambda: set_limits_user(user)  # To know if the command should be executed as root or worker
+if start_cmd["command"] is None or start_cmd["command"] == "":  # Avoid to ron an empty command
+    start_cmd["command"] = "echo 'info: student container started with no command set' "
+
+# Run the command
+p = subprocess.Popen(shlex.split(start_cmd["command"]), preexec_fn=set_limits, stdin=fds[0], stdout=fds[1], stderr=fds[2])
 # Handle the signals
 print("Signal thread")
 thread = threading.Thread(target=lambda: handle_signals(p, client), daemon=True)
 thread.start()
-
 # Wait until the subprocess exit
 print("Waiting for process to end")
 retval = p.wait()
 
-# Exit with the same return value than the process (it will be given to docker and given back to run_student)
-print("Exiting")
-exit(retval)
+if start_cmd["ssh"]:
+    # Start ssh server
+    ssh_user, password = start_ssh_server(user)
+    # Send ssh information
+    ssh_user_bin = User[ssh_user].value  # 0 for root, 1 for worker
+    message = msgpack.dumps({"type": "ssh_student", "ssh_user": ssh_user_bin, "password": password})  # constant size
+    client.send(message)
+    # Wait for user to connect and leave
+    retval = ssh_wait(ssh_user)
+    print("Exiting")
+    exit(retval)
+else:
+    print("Exiting")
+    exit(retval)
--- a/base-containers/base/bin/_run_student_intern
+++ b/base-containers/base/bin/_run_student_intern
@@ -11,6 +11,7 @@ import threading
 import msgpack
 import array
 import struct
+import sys
 from inginious_container_api.utils import set_limits_user, start_ssh_server, ssh_wait
 
 
@@ -33,6 +34,17 @@ def recv_fds(sock, msglen, maxfds):
             # Append data, ignoring any truncated integers at the end.
             fds.fromstring(cmsg_data[:len(cmsg_data) - (len(cmsg_data) % fds.itemsize)])
     return msg, list(fds)
+
+
+# Check the runtime
+runtime = sys.argv[1]
+if runtime == "kata-runtime":
+    shared_kernel = False
+else:
+    shared_kernel = True
+    os.mkdir("/.__input")
+    shared_kernel_file = open("/.__input/__shared_kernel", "w")
+    shared_kernel_file.close()
 
 
 # Connect to the socket
@@ -62,9 +74,10 @@ while start_cmd is None:
 print("Unpacked start cmd")
 user = start_cmd["user"]  # Know if you are worker (on docker) or root (on Kata)
 
-if user=="root" and os.path.exists("/.__input/__shared_kernel"):
-    print("You can not run as root on docker")
+if user == "root" and shared_kernel:
+    print("You can not run as root on docker runtime")
     exit(251)
+
 
 # Add some elements to /etc/hosts and /etc/resolv.conf if needed
 system_files = {"hosts": ("/etc/hosts", True), "resolv.conf": ("/etc/resolv.conf", False)}
--- a/base-containers/base/bin/ssh_student
+++ b/base-containers/base/bin/ssh_student
@@ -10,7 +10,8 @@ from inginious_container_api import ssh_student
 
 parser = argparse.ArgumentParser(formatter_class=argparse.RawTextHelpFormatter, description="Gives to the student a ssh access to a safe environment.\n\
 If a command was specified, runs it in the student container before starting the ssh server. \n\
-This command will proxy stdin, stdout, stderr, most signals and the return value. \n\n\
+This command will proxy stdin, stdout, stderr, most signals and the return value. \n\
+Notice the ssh_student feature requires to allow internet connection in the environment configuration tab. \n\
 There are special return values: \n\
 * 252 means that the command was killed due to an out-of-memory \n\
 * 253 means that the command timed out \n\
@@ -20,15 +21,12 @@ parser.add_argument('--time', help="Timeout (in CPU time) for the container. The
 parser.add_argument('--hard-time', help="Hard timeout for the container (in real time). The default is three times the value indicated for --time",
                     type=int, default=0)
 parser.add_argument('--memory', help="Maximum memory for the container, in Mo. The default is the same as the current container", type=int, default=0)
-parser.add_argument('--share-network', help="Share the network stack of the grading container with the student container. This is not the case by "
-                                            "default. If the container container has network access, this will also be the case for the student!",
-                    action='store_true', dest='share_network')
 parser.add_argument('--run-as-root', help="Tries to give root access to the student if possible. This feature is in beta and should not be used yet", action='store_true')
-parser.add_argument('cmd', help="Command to be run on the remote container before opening ssh connection", default=None, nargs=argparse.REMAINDER)
+parser.add_argument('cmd', help="Command to be run on the remote container before opening ssh connection. If the cmd is a script, it is allowed to start new subprocess"
+                                "If you want subprocess to run in background while the student has ssh access, pay attention to launch them in a non-blocking way", default=None, nargs=argparse.REMAINDER)
 
 args = parser.parse_args()
 
 exit(ssh_student.ssh_student(cmd=' '.join(args.cmd), container=args.container, time_limit=args.time,
                              hard_time_limit=args.hard_time, memory_limit=args.memory,
-                             share_network=args.share_network,
                              run_as_root=args.run_as_root))
--- a/base-containers/base/bin/ssh_student
+++ b/base-containers/base/bin/ssh_student
@@ -6,9 +6,10 @@
 import argparse
 import sys
 
-from inginious_container_api import run_student
+from inginious_container_api import ssh_student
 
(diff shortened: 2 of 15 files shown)


base-containers/base/inginious_container_api/run_student.py
```diff
@@ -109,7 +112,19 @@ def receive_signal(signum_s):
 
             signal_handler_callback(receive_signal)
 
+        if ssh:  # The student_container will send id and password for ssh connection, transfer it to the agent
+            ssh_id = msgpack.loads(connection.recv(250))  # TODO: Why 250? It works but why ?
```
GuillaumeDerval, line 116: You have to ensure that the message length is < 250 bytes

base-containers/base/bin/_run_student_intern
```diff
@@ -79,17 +111,71 @@ print("Chdir")
 os.chdir(start_cmd["working_dir"])
 print("Popen")
 print(fds)
-p = subprocess.Popen(shlex.split(start_cmd["command"]), preexec_fn=setlimits, stdin=fds[0], stdout=fds[1], stderr=fds[2])
 
-# Handle the signals
-print("Signal thread")
-thread = threading.Thread(target=lambda: handle_signals(p, client), daemon=True)
-thread.start()
+if start_cmd["ssh"]:
```
GuillaumeDerval, line 115: Is it possible for this to be in its own command? It's strange to run `run_student` without cmd and for it to start the ssh server itself.

Three reasons:
1) I think the same could be achieved with `run_student start_the_ssh_server_please`
2) The behavior of run_student would be more consistent
3) It would be nice to parametrize more the way the ssh server works. For exemple, here we always start `bash` when someone connects, but we may want to start other things: `run_student start_the_ssh_server_please --with-shell=zsh`

inginious/agent/docker_agent/__init__.py
```diff
@@ -545,14 +563,26 @@ async def handle_running_container(self, info: DockerRunningJob, future_results)
                                 hard_time_limit = min(msg["hard_time_limit"] or info.hard_time_limit, info.hard_time_limit)
                                 share_network = msg["share_network"]
                                 socket_id = msg["socket_id"]
+                                ssh = msg["ssh"]
                                 assert "/" not in socket_id  # ensure task creator do not try to break the agent :-(
                                 self._create_safe_task(self.create_student_container(info, socket_id, environment, memory_limit,
                                                                                      time_limit, hard_time_limit, share_network,
-                                                                                     write_stream))
-                            elif msg["type"] == "ssh_key":
-                                # send the data to the backend (and client)
```
darkyne, line 553: Woops. I meant "frontend" not "backend"

base-containers/base/bin/_run_student_intern
GuillaumeDerval: I see a problem with this: it doesn't warn the agent that the student container needs a ssh port. We can keep the ssh flag for `run_student`: `run_student --port=22 start_the_ssh_server_please --with-shell=zsh` or something like that.

This would make the system more versatile too.

base-containers/base/inginious_container_api/run_student.py
darkyne: Now:
The message sent is: message = {"type": "ssh_student", "ssh_user": ssh_user_value, "password": password}
It has a constant size of 240 bytes (indeed sys.getsizeof(message)==240)
The password size is of constant size and users are stored as 0 for worker, 1 for root.

Thus, now I use: ssh_id = msgpack.loads(connection.recv(240))
I hope it resolves the potential issues here :)

base-containers/base/bin/_run_student_intern
```diff
@@ -79,17 +76,33 @@ print("Chdir")
...
-# Exit with the same return value than the process (it will be given to docker and given back to run_student)
-print("Exiting")
-exit(retval)
+if start_cmd["ssh"]:
+    # Start ssh server
+    ssh_user, password = start_ssh_server(user)
+    # Send ssh information
+    ssh_user_value = 0 if ssh_user == "worker" else 1  # 0 for worker, 1 for root
+    message = {"type": "ssh_student", "ssh_user": ssh_user_value, "password": password}
+    assert (sys.getsizeof(message)) == 240  # Ensure a constant size of 240 bytes
```
anthonygego, line 100: It is highly probable that ``sys.getsizeof`` is architecture-dependent. There may be a more appropriate msgpack API to compute this. I'm not even sure this check would be useful at this stage.

base-containers/base/inginious_container_api/run_student.py
```diff
@@ -99,17 +105,29 @@ def run_student(cmd, container=None,
...
         # Allow to send signals
         if signal_handler_callback is not None:
             def receive_signal(signum_s):
                 signum_data = str(signum_s).zfill(3).encode("utf8")
                 connection.send(signum_data)
-
             signal_handler_callback(receive_signal)
 
+        if ssh:  # The student_container will send id and password for ssh connection, transfer it to the agent
+            ssh_id = msgpack.loads(connection.recv(240))  # The message is asserted to be 240 bytes before sending
```
anthonygego, line 118: There should be a msgpack API to compute exactually at runtime the expected message size, rather than hardcoding it. This is typically the kind of stuff that the library should abstract.

base-containers/base/bin/ssh_student
```diff
@@ -0,0 +1,35 @@
+#! /bin/python3
+# -*- coding: utf-8 -*-
+#
+# This file is part of INGInious. See the LICENSE and the COPYRIGHTS files for
+# more information about the licensing of this file.
+import argparse
+import sys
+
+from inginious_container_api import run_student
```
anthonygego, line 9: As you have defined `ssh_student`, maybe it would be more coherent to import it here instead of `run_student`.

base-containers/base/inginious_container_api/utils.py
```diff
@@ -0,0 +1,94 @@
...
+    execute_process(["/usr/bin/bash", "-c", "echo '{}:{}' | chpasswd".format(ssh_user, password)],
+                    internal_command=True, user=ssh_user)
+    # generate the host keys
+    execute_process(["/usr/bin/ssh-keygen", "-A"], internal_command=True, user=ssh_user)
+
+    # remove /run/nologin if it exists
+    if os.path.exists("/run/nologin"):
+        os.unlink("/run/nologin")
+
+    PermitRootLogin = "yes" if ssh_user == "root" else "no"
```
anthonygego, line 58: `permit_root_login = ...`

anthonygego (changes requested):

base-containers/base/bin/_run_student_intern
darkyne: Indeed, the way I checked its size was incorrect.
I was way better to tansform the message into bytes with msgpack before checking its len.

base-containers/base/inginious_container_api/run_student.py
darkyne: The exact number of bytes can be computed at the student_container side (it's actually a constant: 54 bytes).
But since the whole goal here is to communicate between these 2 sides (student_container and grading_container) I don't see the issue about hardcoding the number of bytes (since it's a consant)  :/

anthonygego: It is rarely a good practice to hardcode that kind of thing unless there is an explicit standardization that ensures this will never change. While msgpack should ensure the size of the message is constant through different architectures by its design, you never know if the internal message representation won't change some day as it is not standardized.

These kind of changes actually already happend once when msgpack was upgraded to version 1.0. So you should really rely on a msgpack for this.

The typical way msgpack deal with this is through its [Unpacker class](https://github.com/msgpack/msgpack-python/tree/main#streaming-unpacking). And this is already the way it is implemented in INGInious : https://github.com/UCL-INGI/INGInious/blob/b60326b702022ffe41391e7b7b003b3b3bbb8080/base-containers/base/bin/_run_student_intern#L59

And, while it is commented "It's ugly", there is actually no many other manners to do this with msgpack. It is simply not designed to support non-msgpack messages in the same stream. A possible alternative, in this really specific case, is to pack a dummy message, compute its size at runtime, and this should be the expected value.

It is clear the actual way to solve this "It's ugly" thing is to ensure only msgpack messages are sent in the stream, and I'm not requiring it. I just want to avoid really bad surprises if something changes in msgpack implementation.

darkyne: Should be better now, computing the expected size at runtime instead of hardcoding it.

darkyne:
Any news about things still to change here ?
To be honnest, I don't understand why Jenkins tests failed (build step failed) but if anyone can give a clue, it may help :)

anthonygego:
Didn't know you were done. I'll have another check very soon. I think we're still waiting for @GuillaumeDerval concerning the way to unpack msgpack messages. Jenkins fails because we need to update to API key, the current one seems to have expired for security reasons.

darkyne:
Well, when testing everything seems to work correctly and the modifications you requested are done.
About the way the message is unpacked, I went for the byte by byte approach as it was already done previously in the code.

That being said, I'm not especially saying "the work is done" just the modifications requested are done but you might want others things to change, don't hesitate :) I'm working on kata during that time anyway.

base-containers/base/bin/_run_student_intern
```diff
@@ -79,17 +78,34 @@ print("Chdir")
 os.chdir(start_cmd["working_dir"])
 print("Popen")
 print(fds)
-p = subprocess.Popen(shlex.split(start_cmd["command"]), preexec_fn=setlimits, stdin=fds[0], stdout=fds[1], stderr=fds[2])
 
+set_limits = lambda: set_limits_user(user)  # To know if the command should be executed as root or worker
+if start_cmd["command"] is None or start_cmd["command"] == "":  # Avoid to ron an empty command
```
GuillaumeDerval, line 83: "ron" -> "run"

base-containers/base/bin/_run_student_intern
```diff
@@ -79,17 +78,34 @@ print("Chdir")
...
 
-# Exit with the same return value than the process (it will be given to docker and given back to run_student)
-print("Exiting")
-exit(retval)
+if start_cmd["ssh"]:
+    # Start ssh server
+    ssh_user, password = start_ssh_server(user)
+    # Send ssh information
+    ssh_user_bin = User[ssh_user].value  # 0 for root, 1 for worker
+    message = msgpack.dumps({"type": "ssh_student", "ssh_user": ssh_user_bin, "password": password})  # constant size
```
GuillaumeDerval, line 101: The comment is not needed and may become wrong in the future

base-containers/base/bin/_run_student_intern
```diff
@@ -79,17 +78,34 @@ print("Chdir")
...
 retval = p.wait()
 
-# Exit with the same return value than the process (it will be given to docker and given back to run_student)
-print("Exiting")
-exit(retval)
+if start_cmd["ssh"]:
+    # Start ssh server
+    ssh_user, password = start_ssh_server(user)
+    # Send ssh information
+    ssh_user_bin = User[ssh_user].value  # 0 for root, 1 for worker
```
GuillaumeDerval, line 100: Why not using "root" and "worker" instead of 0/1?

base-containers/base/bin/ssh_student
```diff
@@ -0,0 +1,34 @@
...
+# -*- coding: utf-8 -*-
+#
+# This file is part of INGInious. See the LICENSE and the COPYRIGHTS files for
+# more information about the licensing of this file.
+import argparse
+import sys
+
+from inginious_container_api import ssh_student
+
+parser = argparse.ArgumentParser(formatter_class=argparse.RawTextHelpFormatter, description='Gives to the student a ssh access to a safe environment.\n\
```
GuillaumeDerval, line 11: Use """ :-)

base-containers/base/inginious_container_api/run_student.py
```diff
@@ -7,18 +7,22 @@
...
 def run_student(cmd, container=None,
         time_limit=0, hard_time_limit=0,
         memory_limit=0, share_network=False,
         working_dir=None, stdin=None, stdout=None, stderr=None,
-        signal_handler_callback=None):
+        signal_handler_callback=None, ssh=False, run_as_root=False):
     """
     Run a command inside a student container
-    :param cmd: command to be ran (as a string, with parameters)
+    :param cmd: command to be ran (as a string, with parameters). If ssh is set to True, this command will be run before launching the ssh server.
```
GuillaumeDerval, line 25: Is it possible to have a process that continues to work once the SSH server is started?
(that is, to detach a process that would continue to work in background).

It would be nice to document a small example explaining how to do that.

base-containers/base/inginious_container_api/run_student.py
```diff
@@ -37,13 +41,20 @@ def run_student(cmd, container=None,
                                     this function can itself be called with a signal value that will immediately be sent
                                     to the remote process. See the run_student script command for an example, or
                                     the hack_signals function below.
+    :param ssh: If set to True, it starts an ssh server for the student instead of running the command as usual.
+    :param run_as_root: If set to True, it tries to execute the command as root (for ssh, it accepts connection as root)
```
GuillaumeDerval, line 45: Please add a warning about runtimes that do not support root

base-containers/base/inginious_container_api/run_student.py
```diff
@@ -81,13 +93,14 @@ def run_student(cmd, container=None,
...
-                                   "socket_id": socket_id}, use_bin_type=True))
+                                   "socket_id": socket_id, "ssh": ssh}, use_bin_type=True))
 
         # Check if the container was correctly started
         message = msgpack.loads(zmq_socket.recv(), use_list=False, strict_map_key=False)
         assert message["type"] == "run_student_started"
+        student_container_id = message["container_id"]
 
-        # Send a dummy message to ask for retval
+        # Send a dummy message to ask for retval (this should not be removed without caution)
```
GuillaumeDerval, line 103: "The socket can only work in a ping-pong way, so we need to send a dummy message"

base-containers/base/inginious_container_api/ssh_student.py
```diff
@@ -0,0 +1,16 @@
...
+#
+# This file is part of INGInious. See the LICENSE and the COPYRIGHTS files for
+# more information about the licensing of this file.
+import sys
+import inginious_container_api.run_student
+
+
+# Simply runs run_student with no cmd and ssh set to True
+def ssh_student(cmd=None, memory_limit=0, hard_time_limit=0, time_limit=0, container=None, share_network=False, run_as_root=False):
+
```
GuillaumeDerval, line 11: Please insert some documentation here

base-containers/base/inginious_container_api/utils.py
```diff
@@ -0,0 +1,99 @@
...
+# more information about the licensing of this file.
+import os
+import tempfile
+import subprocess
+import resource
+import stat
+import time
+from enum import Enum
+
+class User(Enum):
```
GuillaumeDerval, line 13: I would prefer root="root" and worker="worker". It's easier to read while debugging, it loses some bits while sending packets, but it's not a problem.

inginious/agent/docker_agent/__init__.py
```diff
@@ -545,14 +568,30 @@ async def handle_running_container(self, info: DockerRunningJob, future_results)
...
+                                # send the data to the frontend (and client) to reach grading_container
                                 self._logger.info("%s %s", info.container_id, str(msg))
                                 await self.send_ssh_job_info(info.job_id, self._address_host, info.ports[22], msg["ssh_user"], msg["ssh_key"])
+
+                            elif msg["type"] == "ssh_student":
+                                # send the data to the frontend (and client) to reach student_container
+                                if info.enable_network:
+                                    info_student = None
+                                    if len(self._student_containers_running) > 0:
+                                        info_student = self._student_containers_running[msg["container_id"]]
```
GuillaumeDerval, line 586: Please check that the container_id belongs to one of the student containers of this grading container.

GuillaumeDerval (changes requested):

base-containers/base/bin/_run_student_intern
darkyne: Indeed. Actually encoding the user as a binary value (0 or 1) was to allow a constant size message.
Now since the message size is sent seperately, I will simply use "worker" and "root" instead :)

base-containers/base/inginious_container_api/run_student.py
darkyne: Right now with the curent implementation, the command is started in a thread p and we p.wait() before launching the ssh server...
But actually, it would be easy to add an argument to specify that we want the command to continue in its thread and to stop only when the exercice is finished. Thus in this case, we would not p.wait() directly but instead we would start the ssh server and maybe p.wait or p.kill at the end before exiting the container.

Do you think such an argument would be useful ?
--continue: if you want your command (probably a script) to continue running in background while the student has ssh access.

darkyne:
Just added the option:
--continuous

This allows the command (if there is one) to be run in a background process while the student has ssh access.
When --continuous is not set, the command is completed before starting the ssh server.

As a quick test, I wrote a python script writing something in a file every second and putted the --continuous flag.
When I connected as a student with ssh, I was able to see the file being edited every second.
This background process (running the command) terminates when the student leaves the connection (or when he timesout).

Remark: For the general case of run_student (thus with "ssh" set to False), the new "continuous" argument has no effect (as described it is only useful if "ssh" is also set to True).

I hope this meets the requirements.

anthonygego:
I think that ``--run-in-background``, ``--bg``, or anything that recalls the background process concept would be more appropriate than ``--continuous``. I don't mean to be pedant, but the feature would be more easily understood if it relies on commonly used vocabulary.

darkyne:
Indeed, this terminology is better. Corrected.

inginious/agent/docker_agent/__init__.py
```diff
@@ -437,13 +443,26 @@ async def create_student_container(self, parent_info, socket_id, environment_nam
             environment = self._containers[environment_type][environment_name]["id"]
             runtime = self._containers[environment_type][environment_name]["runtime"]
 
+            if ssh:
```
anthonygego, line 446: As this is simple, you can save 3 lines by writing : ``ports_needed = [22] if ssh else []`` :)

darkyne:
Documentation should be better now.
Also did very few little modifications about code-style.
Hope it's fine now :)

base-containers/base/inginious_container_api/run_student.py
```diff
@@ -99,17 +115,30 @@ def run_student(cmd, container=None,
...
         # Allow to send signals
         if signal_handler_callback is not None:
             def receive_signal(signum_s):
                 signum_data = str(signum_s).zfill(3).encode("utf8")
                 connection.send(signum_data)
-
             signal_handler_callback(receive_signal)
 
+        if ssh:  # The student_container will send id and password for ssh connection, transfer it to the agent
+            s = connection.recv(4)  # First 4 bytes are for the size
```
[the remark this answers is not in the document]
anthonygego: This is still to be discussed :-)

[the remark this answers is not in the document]
darkyne: True. We can have a word about it during our meeting :)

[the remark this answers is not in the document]
darkyne: Added the '!' symbol and noticed no difference when testing, should be fine :)

inginious/agent/docker_agent/__init__.py
```diff
@@ -545,14 +564,30 @@ async def handle_running_container(self, info: DockerRunningJob, future_results)
...
+                                    if len(self._student_containers_running) > 0 and msg["container_id"] in info.student_containers:
+                                        info_student = self._student_containers_running[msg["container_id"]]
+                                    else:
+                                        self._logger.exception("Exception: no linked student_container running.")
+                                    self._logger.info("%s %s", info_student.container_id, str(msg))
+                                    await self.send_ssh_job_info(info.job_id, self._address_host, info_student.ports[22], msg["ssh_user"], msg["ssh_key"])
+                                else:
+                                    await self.send_job_result(info.job_id, "ssh for student requires internet access in the task configuration !")
+                                    self._logger.exception("Exception: ssh for student requires internet access in the task configuration !.")
+                                    await self.kill_job(BackendKillJob(job_id=info.job_id))  # Just to be 100% sure the job is killed
```
anthonygego, line 590: It seems that killing the job here may result in exceptions in ``handle_job_closing``:

    2021-06-30 08:48:54,638 - inginious.agent.docker - ERROR - Exception: ssh for student requires internet access in the task configuration !.
    NoneType: None
    2021-06-30 08:48:54,638 - inginious.backend - INFO - Job ad3d0b58-6368-4a64-97c1-25c14e388a35 finished on agent b'\x00\x19I]\x00'
    2021-06-30 08:48:54,713 - inginious.agent.docker - WARNING - Container 8dce2b6b4432639b526bde7e5acf2c2a378d65b1e9d60e132d64df9dff2d4969 has not given any result
    2021-06-30 08:48:54,723 - inginious.agent.docker - ERROR - Exception in handle_job_closing
      File "/home/anthony/Documents/inginious-dev/INGInious/inginious/agent/docker_agent/__init__.py", line 781, in handle_job_closing
        await self.send_job_result(info.job_id, result, error_msg, grade, problems, tests, custom, state, archive, stdout, stderr)
      File "/home/anthony/Documents/inginious-dev/INGInious/inginious/agent/__init__.py", line 215, in send_job_result
        raise JobNotRunningException()
    inginious.agent.JobNotRunningException
    2021-06-30 08:48:55,328 - inginious.webapp.requests - INFO - 127.0.0.1:48590 - "- POST /course/test/test_ssh" - 200 OK
    127.0.0.1 - - [30/Jun/2021 08:48:55] "POST /course/test/test_ssh HTTP/1.1" 200 -

Moreover, I don't know where this ``NoneType: None`` comes from in my logs. The error message is not displayed in the webapp. I'm only left with *Internal error* :-(

darkyne: Indeed, the error should be correctly handled and the feedback should be more explicit.
The strange NoneType: None origin is surprisingly hard to find :o 
Working on this right now :)

darkyne: About the "Nonetype: None" being logged, it's clearly the self._logger.exception("...") that makes this happen...
But no idea why :/

anthonygego:
LGTM. @GuillaumeDerval could you have a quick look at the last commit ?

anthonygego:
@GuillaumeDerval  could you have a quick look at the last commit so that I can merge this part ?

base-containers/base/bin/_run_student_intern
```diff
@@ -64,6 +62,7 @@ while start_cmd is None:
     for obj in unpacker:
         start_cmd = obj
 print("Unpacked start cmd")
+user = start_cmd["user"]  # Know if you are worker (on docker) or root (on Kata)
```
GuillaumeDerval, line 65: Maybe add a check here that we are indeed allowed to start "root" in this container.
Is it possible for a compromised grading container to start a root student container on Docker?

GuillaumeDerval: (it is indeed currently possible since this message comes from the grading container)

base-containers/base/inginious_container_api/run_student.py
```diff
@@ -37,13 +40,22 @@ def run_student(cmd, container=None,
                                     this function can itself be called with a signal value that will immediately be sent
                                     to the remote process. See the run_student script command for an example, or
                                     the hack_signals function below.
+    :param ssh: If set to True, it starts an ssh server for the student after the command finished.
+    :param run_as_root: If set to True, it tries to execute the command as root (for ssh, it accepts connection as root).
+                        Default is False. This is a Beta feature and should not be used yet.
```
GuillaumeDerval, line 45: Add a warning saying that it depends on the runtime?

inginious/agent/docker_agent/__init__.py
```diff
@@ -437,13 +443,22 @@ async def create_student_container(self, parent_info, socket_id, environment_nam
             environment = self._containers[environment_type][environment_name]["id"]
             runtime = self._containers[environment_type][environment_name]["runtime"]
 
+            ports_needed = [22] if ssh else []
+            ports = {}
+            for p in ports_needed:
+                if len(self._external_ports) == 0:
+                    self._logger.warning("User asked for a port but no one are available")
+                    raise CannotCreateJobException('No ports are available right now. Please retry later.')
```
GuillaumeDerval, line 451: Caution: you should ensure that the ports that were already in `ports` are going back to `self._external_ports`.

GuillaumeDerval (changes requested):

base-containers/base/bin/_run_student_intern
darkyne: Just after getting the user from the grading_container message, adding something like:
`if user=="root" and os.path.exists("/.__input/__shared_kernel"): exit`
would do the trick?

This way even in the case of a compromised grading_container manually sending a message asking to run as root, it would not if student container is using docker... no ?

inginious/agent/docker_agent/__init__.py
darkyne: Indeed, actually I had in mind cases with a single port needed so the specific case you are talking about was not possible (since "ports" would be empty)... But indeed since the student_container may ask for multiple ports, some could be given to him before realising there are no enough ports available !
Well seen, I would not have remarked this myself.

base-containers/base/inginious_container_api/run_student.py
darkyne: Well actually it should not depend on the runtime.
Well right now the entire "run_student" thing is unavailable for kata indeed....
But when run_student will be available for kata, the "run_as_root" parameter has nothing to do with the docker or kata runtime being selected for the task (in other words, a docker grading_container should be able to start a kata root student_container)

In other words, the entire parameter "run_as_root" is useful for now indeed run_student strats by exiting if run_as_root is set to true but it will have sense in a near future.

base-containers/base/inginious_container_api/utils.py
```diff
@@ -62,9 +62,12 @@ def start_ssh_server(ssh_user):
     execute_process(["/usr/sbin/sshd",
                     "-p", "22",
                     "-o", "PermitRootLogin={}".format(permit_root_login),
-                    "-o", "PasswordAuthentication=yes", "-o", "StrictModes=no", "-o",
-                    "AllowUsers={}".format(ssh_user)], internal_command=True, user=ssh_user)
+                    "-o", "PasswordAuthentication=yes", "-o", "StrictModes=no",
+                    "-o", "ForceCommand=echo LOGIN: Good luck !; script -q .ssh_logs; cp .ssh_logs student/.ssh_logs; echo LOGOUT: Good bye!",
```
darkyne, line 66: From the tests I did, this works well.
It allows the student to play within an interactive shell (launched via the "script" command) and closes the ssh connection when he leaves the shell. The log is written within the ssh session with worker permissions so the student is able to delete the .ssh_logs file (in the task directory) but should probably only code in the "student" directory so the chance of a sad rm * deleting it is low. Only when the student leaves the session, the file is copied into the student directory to be reflected within the grading_container.

anthonygego (commented):
"Teardown" is probably a more appropriate and common word than "closure" for this case, which is already used in CS for a different concept.

anthonygego:
LGTM. Waiting for @GuillaumeDerval approval. 
Please do not add any new feature here now.

anthonygego (approved):

GuillaumeDerval:
LGTM

GuillaumeDerval (approved):

anthonygego:
As there is one main commit with several fix commits, I'm going to squash everything.

Issue #625, opened 2021-02-10
MCQs : small improvements


anthonygego:
Feature request:
- Allow not to shuffle MCQs answers
- Allow to check only one correct answer

PR #682, opened 2021-06-16
frontend/mcq : Allowing task creator to (not) shuffle choices.

Drumor:
regarding #625 this PR allows to shuffle or not choices in MCQ

### DIFF
--- a/inginious/common/tasks_problems.py
+++ b/inginious/common/tasks_problems.py
@@ -211,6 +211,7 @@ class MultipleChoiceProblem(Problem):
         super(MultipleChoiceProblem, self).__init__(problemid, content, translations, taskfs)
         self._header = content['header'] if "header" in content else ""
         self._multiple = content.get("multiple", False)
+        self._unshuffle = content.get("unshuffle", False)
         if "choices" not in content or not isinstance(content['choices'], (list, tuple)):
             raise Exception("Multiple choice problem " + problemid + " does not have choices or choices are not an array")
         good_choices = []
@@ -330,7 +331,7 @@ class MultipleChoiceProblem(Problem):
     def parse_problem(self, problem_content):
         problem_content = Problem.parse_problem(problem_content)
         # store boolean fields as booleans
-        for field in ["optional", "multiple", "centralize"]:
+        for field in ["optional", "multiple", "centralize","unshuffle"]:
             if field in problem_content:
                 problem_content[field] = True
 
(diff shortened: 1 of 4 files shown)


anthonygego:
Didn't test it yet. Can you confirm that it allows to reload an old submission correctly ?

Drumor:
I did test by submitting with master branch then getting back to shuffle_mcq and reload and seems to work correctly.

inginious/frontend/task_problems.py
```diff
@@ -185,12 +185,12 @@ def show_input(self, template_helper, language, seed):
                 if entry['valid'] and limit > 0:
                     choices.append(entry)
                     limit = limit - 1
-
-        rand.shuffle(choices)
-
+        if self._shuffle:
+            rand.shuffle(choices)
+        else:
+            choices = sorted(choices, key=lambda k: k['index'])
```
anthonygego, line 191: You short-circuit only one shuffle here, and sort an already randomly generated list. Your list of choice will be displayed in the teacher order, but the choice set will still be randomly generated according to the seed.

I don't know if this is the expected result.

Drumor: As there is a pick of correct answer (multiple or not) just after the first shuffle, it ruins the order in any case. So i'm not sure that a course-circuit is necessary on first shuffle as the order will not be respected anyway.

anthonygego: What would be the point of this feature then ?  The idea is to have the same rendering whatever the random seed parameters are. Currently, you will not have the same rendering if the language is different for instance.

anthonygego:
Can this be reviewed now ?

Drumor:
Yes! This one is good for me :)

inginious/common/tasks_problems.py
```diff
@@ -211,6 +211,7 @@ def __init__(self, problemid, content, translations, taskfs):
         super(MultipleChoiceProblem, self).__init__(problemid, content, translations, taskfs)
         self._header = content['header'] if "header" in content else ""
         self._multiple = content.get("multiple", False)
+        self._shuffle = content.get("shuffle", False)
```
anthonygego, line 214: The current behaviour is that answers are shuffled. Therefore, the default should be ``True``, or the name and effect of the feature inversed.

anthonygego (changes requested):

inginious/common/tasks_problems.py
Drumor: You're right! :)

anthonygego:
This is not the only change to make. The default value should reflect in the studio : the checkbox has to be checked if the field is not set.

anthonygego (approved):

anthonygego:
LGTM

PR #685, opened 2021-07-15
lti/cookieless : correctly handle sessions

anthonygego:
This fixes erroneous code where session fetched was still the cookie one if available and never replaced. This resulted in displaying the same task on multiple-LTI-frame pages or more weird things according to the request sequence.

Moreover, Flask doesn't deal at all with session hotswap and its context stack, therefore the session manager creates the new session when it detects the LTI launch page is requested. Workaround code for some strange behaviours could then be removed.

### DIFF
--- a/inginious/frontend/flask/mongo_sessions.py
+++ b/inginious/frontend/flask/mongo_sessions.py
@@ -19,6 +19,8 @@ from itsdangerous import Signer, BadSignature, want_bytes
 from flask.sessions import SessionMixin
 from werkzeug.datastructures import CallbackDict
 from flask.sessions import SessionInterface
+from werkzeug.exceptions import HTTPException
+from inginious.frontend.pages.lti import LTILaunchPage
 
 
 class MongoDBSession(CallbackDict, SessionMixin):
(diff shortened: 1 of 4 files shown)


anthonygego:
@GuillaumeDerval  could you have a look at this before the start of August ? This is actually quite a critical issue.

inginious/frontend/flask/mongo_sessions.py
```diff
@@ -67,9 +67,13 @@ def _get_signer(self, app):
     def open_session(self, app, request):
         # Check for cookieless session in the path
         path_session = re.match(r"(/@)([a-f0-9A-F_]*)(@)", request.path)
+        is_lti_launch = re.match(r"(/lti)", request.path)
```
GuillaumeDerval, line 70: `request.path.startswith("/lti")` ?

inginious/frontend/pages/lti.py
```diff
@@ -173,9 +173,9 @@ def GET(self, courseid, taskid):
     def POST(self, courseid, taskid):
         (sessionid, loggedin) = self._parse_lti_data(courseid, taskid)
         if loggedin:
-            return redirect(self.app.get_homepath() + "/@{}@/lti/task".format(sessionid))
+            return redirect(self.app.get_homepath() + "/lti/task")
```
GuillaumeDerval, line 176: I'm a bit worried that flask thus creates a session when you reach /lti, then redirects you to /lti/task, that creates another session id?

inginious/frontend/flask/mongo_sessions.py
GuillaumeDerval: or you meant `re.match(r"(/lti/?)", request.path)` maybe?

You should ensure that it's either a path under the /lti/ path of the /lti page:
`request.path == "/lti" or request.path.startswith("/lti/")` is another non-regex equivalent, in order to avoid "using a canon to kill a fly", like we say in french ;-)

anthonygego:
Well seen, seems to work because the ``/lti/task`` page is prefixed with the sessionid when redirected, I will specifically set the launch page url to avoid any issue. I'll try to fetch it from the Flask url map to avoid hardcoding it too.

anthonygego:
Updated.

anthonygego:
What about this ? It gets old but it's still critical :-)

GuillaumeDerval:
LGTM

PR #688, opened 2021-07-23
ssh_student: Add information about setup and teardown

darkyne:
Adding some information, explaining the use of the setup and teardown scripts.
Also adding some information about limitations.


anthonygego:
LGTM.

PR #689, opened 2021-07-27
adding scripts isolation

darkyne:
To ease the creation of setup / teardown scripts, a student/scripts directory is automatically mounted in the container. This directory is isolated from the student during the ssh session, allowing the teacher to put scripts files in a safe directory without worying about the student inspecting them.

Documentation was also updated.

### DIFF
--- a/base-containers/base/inginious_container_api/utils.py
+++ b/base-containers/base/inginious_container_api/utils.py
@@ -96,3 +96,14 @@ def ssh_wait(ssh_user, timeout=None):
     else:
         print("No one connected !")
         return 253  # timeout
+
+def scripts_isolation(isolate):
+    """
+    Make the script directory isolated or not from the student
+    """
+    if isolate:
+        value = "-rwx"
+    else:
+        value = "+rwx"
+    execute_process(["/usr/bin/bash", "-c", "chmod {} student/scripts".format(value)],
+                    internal_command=True, user="root")
--- a/base-containers/base/inginious_container_api/utils.py
+++ b/base-containers/base/inginious_container_api/utils.py
@@ -98,12 +98,8 @@ def ssh_wait(ssh_user, timeout=None):
         return 253  # timeout
 
 def scripts_isolation(isolate):
-    """
-    Make the script directory isolated or not from the student
-    """
+    """ Make the script directory isolated or not from the student """
     if isolate:
-        value = "-rwx"
+        os.chmod("student/scripts", 000)
     else:
-        value = "+rwx"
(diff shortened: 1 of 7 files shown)


base-containers/base/inginious_container_api/utils.py
```diff
@@ -96,3 +96,14 @@ def ssh_wait(ssh_user, timeout=None):
...
+
+def scripts_isolation(isolate):
+    """
+    Make the script directory isolated or not from the student
+    """
+    if isolate:
+        value = "-rwx"
+    else:
+        value = "+rwx"
+    execute_process(["/usr/bin/bash", "-c", "chmod {} student/scripts".format(value)],
```
anthonygego, line 108: Try to use Python code as much as possible here (using the library ``os``).

base-containers/base/inginious_container_api/utils.py
```diff
@@ -96,3 +96,10 @@ def ssh_wait(ssh_user, timeout=None):
     else:
         print("No one connected !")
         return 253  # timeout
+
+def scripts_isolation(isolate):
+    """ Make the script directory isolated or not from the student """
+    if isolate:
+        os.chmod("student/scripts", 000)
```
anthonygego, line 103: Did you check this does not fail if the folder does not exist ?

darkyne: Yes, hopefuly I anticipated this one :)
A "student/scripts" directory is always mounted inside the student_container, even if it's empty (and the initial task had no student/scripts directory).
This was just to make it the most general (every student_container thus has a "student/scripts" directory.

Don't hesitate to tell me if you think it's not the right approach

anthonygego: LGTM then. I suggest to let @GuillaumeDerval have a look at his.

darkyne:
Added 2 little lines to resolve #694 so both small features could be added at once.

anthonygego:
Reminder: please change the relative paths to absolute paths to ensure that the scripts will work from everywhere. It should be good for me after that... as no one else seems to care about this PR.

darkyne:
Changed to absolute paths and tested.

Drumor:
Seems ok to me, still needs review from @GuillaumeDerval

GuillaumeDerval:
A remark on this: a "mere" chmod on the directory will not work if the student is root, and this should be noted explicitly somewhere.

Apart from that, i'm ok with this

darkyne:
Yes, obviously if the student has root access on the student_container, this will not prevent him from accessing the files.
I will add a few lines about it in the doc to make it clear.

PR #691, opened 2021-08-09
hotfix ssh_student

darkyne:
This error made the use of "ssh_student" with a python "run" script not correctly closing the student_container.

Actually it tried to run the teardown-script even if the string was None.
This issue, however, did not impact run_student nor ssh_student when using bash "run" script.

It should be merged ASAP.


Drumor:
LGTM

PR #699, opened 2021-08-20
Send LTI informatoin to the client

jyounggo:
The patch will allow the client to access LTI information within a grading container.
"consumer_key" and "outcome*" fields are excluded for security reason.

### DIFF
--- a/inginious/frontend/submission_manager.py
+++ b/inginious/frontend/submission_manager.py
@@ -287,7 +287,7 @@ class WebAppSubmissionManager:
         # Send LTI information to the client except "consumer_key"
         lti_info = self._user_manager.session_lti_info()
         for key in lti_info:
-            if key == "consumer_key": # Skip "consumer_key"
+            if key == "consumer_key" or key.startswith("outcome"): # Skip "consumer_key" and "outcome*"
                 continue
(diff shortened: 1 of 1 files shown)


inginious/frontend/submission_manager.py
```diff
@@ -284,6 +284,16 @@ def add_job(self, task, inputdata, debug=False):
         inputdata["@random"] = states["random"] if "random" in states else []
         inputdata["@state"] = states["state"] if "state" in states else ""
 
+        # Send LTI information to the client except "consumer_key"
+        lti_info = self._user_manager.session_lti_info()
```
anthonygego, line 288: This may return ``None`` if the session is not identified as an LTI one. Putting ``or {}`` at the end of the line should fix this issue I guess.

anthonygego (changes requested):

inginious/frontend/submission_manager.py
jyounggo: Add "if clause" to check if lti_info is available
```suggestion
        lti_info = self._user_manager.session_lti_info()
        if lti_info:
```

anthonygego:
LGTM. Thank you ! I'll squash the commits as this is a small PR.

PR #700, opened 2021-08-24
run_student and ssh_student for Kata runtime

darkyne:
Well, sorry for the messy code monstrosity !

This allows to use the run_student and ssh_student for kata runtimes, being from the grading_container side (directly within the environment configuration page) or at the student_container side (by calling run_student --run-as-root for example).

In other words, these runtimes combinations are possible for grading_container and student_container:
docker-docker (vanilla)
docker-kata
kata-docker
kata-kata

The features are all working (well according to my own tests):
- giving root ssh access and logging the session (well since the user is running as root he can manually stop the logs recording !)
- running student code and scripts with root privileges 
- redirecting the stdout and stderr (not sure about how to handle the stdin yet)

While everything seems to work fine, I am pretty sure few things might not be up to standards yet so don't hesitate to tell me most important things to improve here !
I believe commits descriptions should be relatively clear enough :)

Here are some super basic examples to show the behaviours:

SSH API:
ssh_student_run_as_root

SSH Student interface:
ssh_student_run_as_root2

Run_student trivial example:
run_student_as_root

Run_student outputs: (error example)
run_student_as_root2


inginious/agent/docker_agent/__init__.py
```diff
@@ -441,7 +441,10 @@ async def create_student_container(self, parent_info, socket_id, environment_nam
                 return
 
             environment = self._containers[environment_type][environment_name]["id"]
-            runtime = self._containers[environment_type][environment_name]["runtime"]
+            if run_as_root:
+                runtime = "kata-runtime"
+            else:
+                runtime = "io.containerd.runc.v2"
```
GuillaumeDerval, line 447: This won't work, as runtime names are per-instance (its in the docker's config).
The task should be able to indicate the name of the runtime directly, rather than "run_as_root"

base-containers/base/inginious_container_api/run_student.py
```diff
@@ -52,15 +53,12 @@ def run_student(cmd, container=None,
...
-        return 251
-
-    elif run_as_root:  # Allowing root access to student is forbidden for now (TODO fiw me)
-        print("run_student as root is not available yet")
-        return 251
-
 
+    #  Checking runtimes
+    shared_kernel = os.path.exists("/.__input/__shared_kernel")
+    only_dockers = shared_kernel and not run_as_root
```
GuillaumeDerval, line 59: shared_kernel and run_as_root should be mutually exclusive, and it should be (re)checked here I think

base-containers/base/bin/_run_student_intern
```diff
@@ -36,96 +77,171 @@ def recv_fds(sock, msglen, maxfds):
...
-
 # Add some elements to /etc/hosts and /etc/resolv.conf if needed
-system_files = {"hosts": ("/etc/hosts", True), "resolv.conf": ("/etc/resolv.conf", False)}
-for name, (spath, append) in system_files.items():
-    if os.path.exists(os.path.join('/task/systemfiles/', name)):
-        try:
-            open(spath, 'ab' if append else 'wb').write(b'\n' + open(os.path.join('/task/systemfiles/', name), 'rb').read())
-        except IOError:
-            exit(254)
+    system_files = {"hosts": ("/etc/hosts", True), "resolv.conf": ("/etc/resolv.conf", False)}
```
GuillaumeDerval, line 221: These indents should probably not be there

inginious/agent/docker_agent/__init__.py
darkyne: Well, it works locally but I don't know exactly the differences when running on the deployed version :/ 
Anyway, indeed being able to indicade the runtime name would be more efficient (no bug if the runtime version name change for example) but I wanted a kind of shortcut to tell to start as docker or root without requiring more knowledge... :/ 
I can simply ask the runtime name as argument instead of "run_as_root" but I suppose it may be less easy to use.

base-containers/base/inginious_container_api/run_student.py
darkyne: Well actually, "shared_kernel" simply means that the container is running on docker. Damn it's true that it's not explicit ! I will rename this.
The "only_dockers" means that both grading_container and student_container are on docker.

So they are not excluse since we can have shared_kernel and only_dockers or not (docker-docker or kata-docker).
Actually the only situation to avoid is shared_kernel and run_as_root, still it's a nice idea to recheck this right there :)

darkyne: Nevermind, I missunderstood your comment, you are right shared_kernel and run_as_root are exclusive and should be checked here :)

base-containers/base/bin/_run_student_intern
darkyne: Indeed, my bad !

inginious/agent/docker_agent/__init__.py
anthonygego: Indeed, hardcoding the runtimes should be avoided as it was designed to support multiple runtimes from the start.

@GuillaumeDerval I actually remember being able to run containers with different runtimes on the same instance as long as they are available, but maybe I smoked (not sure your first remark is about that).

base-containers/base/inginious_container_api/run_student.py
anthonygego: Still not clear for me... could you give us a logic table to describe what happens in each cases ?

darkyne: Here, we are in run_student.py so we are on the grading_container.
shared_kernel simply means that the current container (grading_container) is using a shared_kernel runtime (aka: docker runtime)
run_as_root is the argument passed by the user to run_student and represents the fact we want to start a student_container to run code in it as normal user or as root.

Thus there is no mutual exclusions, here are the different scenarios, remember we are in the grading_container so shared-kernel says nothing about the student_container:

shared_kernel  -  run_as_root            
True                     True                      Docker -> Kata
True                     True                      Docker -> Docker (aka usual scenario)
False                    True                     Kata -> Kata
False                    False                   Kata -> Docker

darkyne: I believe there is no issue here so I resolve the conversation.
The only dangerous scenario was about having a student_container with shared_kernel runtime and run_as_root together which is verified here: https://github.com/darkyne/INGInious/blob/60a915a8ab7fdad457777a2832ee110f44fdf8a1/base-containers/base/bin/_run_student_intern#L216

Don't hesitate if it's still not clear :)

inginious/agent/docker_agent/__init__.py
darkyne: This should be resolved now, just pushed modifications: https://github.com/darkyne/INGInious/blob/496b3dc8faef05d06857b9c1c93d52235964e1e0/inginious/agent/docker_agent/__init__.py#L446

Now when the user manually entered an environment as input, it tries to use it.
If nothing was specified (which is generally the case), if run_as_root then it takes the first available runtime which allows run_as_root (each DockerRuntime object has a run_as_root value) otherwise, if not run_as_root it takes the first available runtime wich has run_as_root set to False.

I hope this solved the problem :)

darkyne:
Last pushes on this PR today changed quite a bit part of the code so I'm updating the PR description to better describe the implementation.

darkyne:
The code changed quite a lot so I suggest @GuillaumeDerval to have a look here since I expect some little things to be unperfect for now.

base-containers/base/inginious_container_api/utils.py
```diff
@@ -96,3 +105,213 @@ def ssh_wait(ssh_user, timeout=None):
...
+    message = msgpack.unpackb(bytes(buf), use_list=False)  # EXCEPTION EXTRA DATA
+    return message
+
+
+async def stdio():
+    """ Create the stdin and stdout streams to communicate with the agent """
+    my_loop = asyncio.get_event_loop()
+    reader = asyncio.StreamReader()
+    reader_protocol = asyncio.StreamReaderProtocol(reader)
+    writer_transport, writer_protocol = await my_loop.connect_write_pipe(asyncio.streams.FlowControlMixin, os.fdopen(1, 'wb'))
```
anthonygego, line 245: Here, on the contrary, you seem to allow binary data in your ``fdopen``

base-containers/base/inginious_container_api/utils.py
```diff
@@ -96,3 +105,213 @@ def ssh_wait(ssh_user, timeout=None):
...
+    return reader, writer
+
+
+def handle_stdin_message(msg, proc_input, proc):
+    """ Process a single message from the agent (stdin message for the student code process or signals messages).
+    Used only when both containers are not on a shared kernel """
+    try:
+        if msg["type"] == "stdin":
+            input_content = msg["message"] + '\n'
+            proc_input.write(input_content.encode("utf8"))
```
anthonygego, line 257: Again, only text supported here I guess

base-containers/base/inginious_container_api/utils.py
```diff
@@ -96,3 +105,213 @@ def ssh_wait(ssh_user, timeout=None):
...
+            return "pipe_closed"
+    except:
+        return
+
+
+async def handle_stdin(reader: asyncio.StreamReader, proc_input, proc):
+    """ Deamon to handle messages from the agent.
+    Used only when both containers are not on a shared kernel"""
+    try:
+        while not reader.at_eof():
```
anthonygego, line 275: This piece of code looks very similar or identical to  https://github.com/UCL-INGI/INGInious/pull/700/files#diff-5d473b86eabb4b30b7cce4737f8ab950243263b85e8a7c7b27ce0446071b8880R229

Maybe it would be cleaner to write a small function.

base-containers/base/inginious_container_api/utils.py
```diff
@@ -96,3 +105,213 @@ def ssh_wait(ssh_user, timeout=None):
...
+            if status == "pipe_closed":
+                return
+    except:  # This task will raise an exception when the loop stops
+        return
+
+
+def handle_outputs_helper(output, socket_id, output_type, lock, event_loop, container_stdout):
+    """ Simple helper to launch the handle_outputs function in its own thread using its own asyncio loop.
+    Used only when both containers are not on a shared kernel """
+    loop = asyncio.new_event_loop()
```
anthonygego, line 298: Here you run a new loop with a single task, that do not really rely on async code, in a thread. Wouldn't it be more simple to run sync code in the main loop executor ? I may be mistaken about ``write_stdout`` function yet.

inginious/agent/docker_agent/__init__.py
```diff
@@ -441,15 +440,27 @@ async def create_student_container(self, parent_info, socket_id, environment_nam
                 return
 
             environment = self._containers[environment_type][environment_name]["id"]
-            runtime = self._containers[environment_type][environment_name]["runtime"]
+
+            # If the user did not enter any environment name as input, it is "default"
+            if environment_name != "default":
```
anthonygego, line 445: I'm quite confused about this.

inginious/agent/docker_agent/__init__.py
```diff
@@ -441,15 +440,27 @@ async def create_student_container(self, parent_info, socket_id, environment_nam
...
+                runtime = self._runtimes[runtime_name].runtime
+            else:
+                runtime_name = {k for k in self._runtimes if not self._runtimes[k].run_as_root}.pop()
+                runtime = self._runtimes[runtime_name].runtime
 
             ports_needed = [22] if ssh else []
             ports = {}
             for p in ports_needed:
-                if len(self._external_ports) == 0:
+                if self._external_ports is None or len(self._external_ports) == 0:
```
anthonygego, line 457: ``if not self._external_ports`` should probably also work for both conditions

inginious/agent/docker_agent/__init__.py
```diff
@@ -505,6 +517,46 @@ async def create_student_container(self, parent_info, socket_id, environment_nam
...
+                    content = await read_stream.readexactly(length)
+                    if outtype == 1:  # stdout
+                        buffer += content
+
+                    if outtype == 2:  # stderr
+                        self._logger.debug("Received stderr from containers:\n%s", content)
+                    # 4 first bytes are the length of the message. If we have a complete message...
+                    while len(buffer) > 4 and len(buffer) >= 4 + struct.unpack('!I', buffer[0:4])[0]:
+                        msg_encoded = buffer[4:4 + struct.unpack('!I', buffer[0:4])[0]]  # ... get it
+                        buffer = buffer[4 + struct.unpack('!I', buffer[0:4])[0]:]  # ... withdraw it from the buffer
```
anthonygego, line 538: ``struct.unpack('!I', buffer[0:4])[0]``seems to be computed multiple times here without ``buffer`` being modified in the meantime. It looks like a length ? Maybe it would be more readable to have it in an intermediate variable.

inginious/agent/docker_agent/__init__.py
```diff
@@ -517,6 +569,37 @@ async def _write_to_container_stdin(self, write_stream, message):
...
+                if length != 0:
+                    content = await student_read_stream.readexactly(length)
+                    if outtype == 1:  # stdout
+                        buffer += content
+                    if outtype == 2:  # stderr
+                        self._logger.debug("Received stderr from containers:\n%s", content)
+
+                    # 4 first bytes are the length of the message. If we have a complete message...
+                    while len(buffer) > 4 and len(buffer) >= 4 + struct.unpack('!I', buffer[0:4])[0]:
+                        msg_encoded = buffer[4:4 + struct.unpack('!I', buffer[0:4])[0]]  # ... get it
```
anthonygego, line 588: Same comment here concerning `` struct.unpack('!I', buffer[0:4])[0]``

inginious/agent/docker_agent/__init__.py
```diff
@@ -517,6 +569,37 @@ async def _write_to_container_stdin(self, write_stream, message):
         write_stream.write(msg)
         await write_stream.drain()
 
+    async def _handle_student_container_outputs(self, student_read_stream, grading_write_stream):
+        """ Receive outputs (stdout and stderr) from student_container and send them to grading_container without decoding """
+        buffer = bytearray()
```
anthonygego, line 574: This code is structurally really similar to https://github.com/UCL-INGI/INGInious/pull/700/files#diff-16e5de0847426dadade80d602aeb5d530ad44d13a8b92a94bb720a14abd45392R522

There is probably some refactor possible here to make both cleaner.

inginious/agent/docker_agent/__init__.py
```diff
@@ -567,14 +652,56 @@ async def handle_running_container(self, info: DockerRunningJob, future_results)
...
+                            elif msg["type"] == "stdin":
+                                if student_write_stream is None:
+                                    student_container_id = msg["student_container_id"]
+                                    student_sock = await self._docker.attach_to_container(student_container_id)
+                                    student_read_stream, student_write_stream = await asyncio.open_connection(
+                                        sock=student_sock.get_socket())
+                                await self._write_to_container_stdin(student_write_stream, msg)
+
+                            elif msg["type"] == "student_signal":  # Transfer the signal to the student_container
+                                if student_write_stream is None:
```
anthonygego, line 697: Same piece of code as above here. Only the forwarded message if different. Both clauses can be merged, and the forwarded message set according to ``msg["type"]``

anthonygego (changes requested):
I would check if stdio forwarding actually work with binary data. Ensure that not too much data are treated at a time to avoid the loop being blocked in a task too much time.

base-containers/base/bin/INGInious
```diff
@@ -196,11 +201,34 @@ class INGIniousMainRunner:
...
+                return False
+            if message["type"] == "student_signal": #signal to transfer to the student_container via the agent
+                await self.write_stdout({"type": "student_signal", "student_container_id": message["student_container_id"], "signal_data": message["signal_data"]})
+                await self.intern.send_multipart([addr, b'', msgpack.dumps({"type": "run_student_ask_retval"}, use_bin_type=True)])  # ping pong answer
                 return False
             if message["type"] == "run_student_ask_retval":
                 # ignore, just a dummy message
                 return False
+            if message["type"] == "stdin":  # grading container might send stdin messages to send to the student container via the agent
+                await self.intern.send_multipart([addr, b'', msgpack.dumps({"type": "run_student_ask_retval"}, use_bin_type=True)])  # ping pong answer
```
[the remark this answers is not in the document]
darkyne: renamed into "dummy_message"

base-containers/base/bin/_run_student_intern
```diff
@@ -5,80 +5,43 @@
...
-    unpacker.feed(s)
-    for obj in unpacker:
-        start_cmd = obj
-print("Unpacked start cmd")
+# Check security
+student_container_id = start_cmd["student_container_id"]
 user = start_cmd["user"]  # Know if you are worker (on docker) or root (on Kata)
-
 if user == "root" and shared_kernel:
-    print("You can not run as root on docker runtime")
```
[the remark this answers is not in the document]
darkyne: Added a logger message.
Note using "print" in the run_student_intern script itself is prohibited now since the print would be received by the agent with an incorrect format.

base-containers/base/bin/_run_student_intern
```diff
@@ -88,44 +51,50 @@ for name, (spath, append) in system_files.items():
...
+if only_dockers:
+    p = subprocess.Popen(shlex.split(start_cmd["command"]), preexec_fn=set_limits, stdin=fds[0], stdout=fds[1],
+                         stderr=fds[2])
+    signal_thread = threading.Thread(target=lambda: handle_signals(p, socket_unix), daemon=True)
+    signal_thread.start()
+    retval = p.wait()
+else:
+    p = subprocess.Popen(shlex.split(start_cmd["command"]), preexec_fn=set_limits, stdin=subprocess.PIPE,
+                         stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    write_lock = threading.Lock()  # Lock to avoid stdout_handler and stderr_handler to write on the stdout_stream at the same time
```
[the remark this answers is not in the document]
darkyne: They are not combined.
stdout is sent via message of type "stdout" and stderr is sent via message of type "stderr".
But the fact is both are sent on the same asyncio stream (communication between agent and student_container).
Thus, in a situation were the student code is printing both on stdout and stderr at the same time, the lock will avoid having both threads pushing on the stream at the same time.

base-containers/base/bin/_run_student_intern
```diff
@@ -88,44 +51,50 @@ for name, (spath, append) in system_files.items():
...
+    stdout_handler = threading.Thread(target=handle_outputs_helper, args=(p.stdout, start_cmd["socket_id"], "stdout", write_lock, event_loop, container_stdout))
+    stderr_handler = threading.Thread(target=handle_outputs_helper, args=(p.stderr, start_cmd["socket_id"], "stderr", write_lock, event_loop, container_stdout))
+    stdout_handler.start()
+    stderr_handler.start()
+    try:
+        event_loop.run_until_complete(handle_stdin(container_stdin, p.stdin, p))
+    except RuntimeError as e:
+        if e == 'Event loop stopped before Future completed.':  # The loop will be stopped by stdout_handler
```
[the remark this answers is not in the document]
darkyne: Actually the entire comparison is useless, removed it.

base-containers/base/inginious_container_api/run_student.py
```diff
@@ -221,6 +154,165 @@ def _hack_signals(receive_signal):
...
+        connection, addr = server.accept()
+
+        # _run_student_intern should say hello
+        datagram = connection.recv(1)
+        assert datagram == b'H'
+
+        # send the fds and the command/workdir directly to student_container
+        connection.sendmsg([b'S'], [(socket.SOL_SOCKET, socket.SCM_RIGHTS, array.array("i", [stdin, stdout, stderr]))])
+        connection.send(msgpack.dumps(
+            {"type": "run_student_command", "student_container_id": student_container_id, "command": cmd,
```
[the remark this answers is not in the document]
darkyne: Indeed, the socket_id is not sent because you are looking at the "both_dockers" part and the socket_id is not required since the communications are direct grading - student without going through the docker agent.

base-containers/base/inginious_container_api/run_student.py
```diff
@@ -221,6 +154,165 @@ def _hack_signals(receive_signal):
...
+    while msg_type != "run_student_retval":
+        zmq_socket.send(msgpack.dumps({"type": "run_student_ask_retval"}, use_bin_type=True))  # ping pong socket
+        message = msgpack.loads(zmq_socket.recv(), use_list=False, strict_map_key=False)
+        msg_type = message["type"]
+
+        if msg_type == "stdout":
+            stdout_file.write(message["message"].rstrip()+'\n')
+            stdout_file.flush()
+
+        if msg_type == "stderr":
```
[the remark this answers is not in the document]
darkyne: I believe it is better to handle both cases separately since both cases do not write in same file.
But don't hesitate to correct me if you have a cleaner code in mind !

base-containers/base/inginious_container_api/run_student.py
```diff
@@ -221,6 +154,165 @@ def _hack_signals(receive_signal):
...
+            stderr_file.flush()
+    return message
+
+
+def handle_stdin(stdin, student_container_id):
+    """ Read the specified stdin and send content to the student container"""
+    my_context = zmq.Context()
+    my_zmq_socket = my_context.socket(zmq.REQ)
+    my_zmq_socket.connect("ipc:///sockets/main.sock")
+    with os.fdopen(stdin, 'r') as stdin_read:
```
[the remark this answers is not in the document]
darkyne: Indeed here, we are talking about the stdin specified in the run_student arguments (FD). Thus I believe it should be a FD for a file containing text (values / commands etc) or the sys.stdin (0) for example if you want direct interactivity.

After this, the INGInious script (grading_container) send it to the agent (everytime, being from grading to agent or student to agent and vice-versa, the messages on the asyncio stream are always encoded as binaries).

Still, it's true that I considered the FD specified as stdin to be related to a text file or sys.stdin... I am not sure we should consider binaries ?

base-containers/base/inginious_container_api/run_student.py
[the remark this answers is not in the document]
anthonygego: My bad, didn't read carefully.

base-containers/base/bin/_run_student_intern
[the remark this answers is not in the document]
anthonygego: OK. Was this necessary to have then multiplexed in the same stream then ?

base-containers/base/inginious_container_api/run_student.py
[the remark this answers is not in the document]
anthonygego: OK this is not required in this case, but the field is always fetched in all the message treatment routines I see with the ``[]`` operator, without any additional conditions. Where's the treatment routine for this specific message then ? Or is the field added by some code in the way ?

base-containers/base/inginious_container_api/run_student.py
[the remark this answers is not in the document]
anthonygego: In the best case, you should be able to pipe a binary file through ``run_student``. I am not sure this will work as is because you consider text line delimeters in your code. 

Moreover, you can't really prevent users to pipe binary. In the worst case, the system should NOT break. Using line delimiters to chunk binary data will probably give some funky results.

base-containers/base/inginious_container_api/run_student.py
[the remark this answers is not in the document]
darkyne: I renamed the message to have a different name between only_dockers communications and not only_dockers ones.
Also added a verification of the message "type" on the student_container side.

base-containers/base/bin/_run_student_intern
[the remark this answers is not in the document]
darkyne: I believe having messages with type "stdout" and messages with type "stderr" sent on the same stream is not a big issue.
Having 2 separated streams is maybe a performance upgrade maybe not (it would mean handle the 2 streams both on the agent side and on the grading_container side).
I suggest @GuillaumeDerval to give us his opinion on this.

base-containers/base/inginious_container_api/utils.py
darkyne: Indeed, some uggly code duplication here, fixed !

base-containers/base/inginious_container_api/utils.py
darkyne: write_stdout requires to drain the asyncio stream after writing. This is why a loop was required (to await the drain).
I improved it a little bit (having a single loop for stdout and stderr instead of one for each).
Still, I believe the perfect implementation would be to have 3 coroutines (handle stdin reading at agent message, handle_stdout and handle_stderr reading at process outputs) all running on the main loop without using threads.

While this might be improved in the future by someone else (sadly my asyncio comprehension is too limited to achieve it) I believe the current implementation is still "good enough" :/

inginious/agent/docker_agent/__init__.py
darkyne: Indeed :)

inginious/agent/docker_agent/__init__.py
darkyne: Well, because these 2 lines have no sense :) 
Removed.

inginious/agent/docker_agent/__init__.py
darkyne: Indeed. Removed many lines :)

inginious/agent/docker_agent/__init__.py
darkyne: Indeed, some code duplication was removed there :)

inginious/agent/docker_agent/__init__.py
darkyne: Using intermediate "length" variable.
Anyway, this hard to read part is now included in a small function making it more explicit.

base-containers/base/bin/_run_student_intern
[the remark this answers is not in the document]
darkyne: Basically this exact same kind of stream (asyncio stream) was already used for different types of messages (such as run_student to tell agent to start container, student_retval to tell it finished etc...). Here the principle is the same, we have different types of messages (type stdout and stderr) transfered on this stream so I believe it should not be an issue.

base-containers/base/inginious_container_api/run_student.py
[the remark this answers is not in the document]
darkyne: Stdin, Stdout and Stderr are now handled as binary streams (both on grading_container and student_container sides)

base-containers/base/inginious_container_api/utils.py
darkyne: Same comment as previous one, now all the streams are binaries.

base-containers/base/inginious_container_api/utils.py
darkyne: Same comment as previous one, now all the streams are binaries.

base-containers/base/bin/INGInious
```diff
@@ -196,11 +201,28 @@ class INGIniousMainRunner:
                                          "memory_limit": message["memory_limit"],
                                          "share_network": message["share_network"],
                                          "socket_id": message["socket_id"],
-                                         "ssh": message["ssh"]})
+                                         "ssh": message["ssh"],
+                                         "run_as_root": message["run_as_root"]})
+                return False
+            if message["type"] == "run_student_init":  # This message might be sent from run_student to transfer to the student_container via the agent
```
GuillaumeDerval, line 207: "might"?

base-containers/base/bin/_run_student_intern
```diff
@@ -5,80 +5,44 @@
...
-    for obj in unpacker:
-        start_cmd = obj
-print("Unpacked start cmd")
+# Check security
+student_container_id = start_cmd["student_container_id"]
 user = start_cmd["user"]  # Know if you are worker (on docker) or root (on Kata)
-
 if user == "root" and shared_kernel:
-    print("You can not run as root on docker runtime")
+    logger.info("You can not run as root on docker runtime")
```
GuillaumeDerval, line 43: "on a runtime with a shared kernel, such as Docker"

base-containers/base/inginious_container_api/run_student.py
GuillaumeDerval: Still not ok for me.

`shared_kernel` indicates whether the kernel is shared or not; if it is the case, NOBODY should be allowed to run as root.
`run_as_root` indicates whether the parent container asked to run the student one as root. this can be compromised and should not be trusted. 

Please add a check to verify that, that could be something like:
```
if shared_kernel:
   run_as_root = False
```

And add comments everywhere explaining what are all the variables.

base-containers/base/inginious_container_api/run_student.py
```diff
@@ -69,90 +68,20 @@ def run_student(cmd, container=None,
...
-        try:
-            os.unlink(socket_path)
-            os.unlink(path)
-        except:
-            pass
-
+        server, socket_id, socket_path, path = create_student_socket(only_dockers)
+        zmq_socket, student_container_id = start_student_container(container, time_limit, hard_time_limit, memory_limit, share_network, socket_id, ssh, run_as_root)
+        connection = send_initial_command(socket_id, server, stdin, stdout, stderr, zmq_socket, student_container_id, cmd, teardown_script, working_dir, ssh, user, only_dockers)
+        allow_to_send_signals(signal_handler_callback, connection, student_container_id, only_dockers)
```
GuillaumeDerval, line 76: Have you checked that it works when the command immediatly returns a signal?
Are you sure there are no race conditions here?

inginious/agent/docker_agent/__init__.py
```diff
@@ -441,15 +440,24 @@ async def create_student_container(self, parent_info, socket_id, environment_nam
                 return
 
             environment = self._containers[environment_type][environment_name]["id"]
-            runtime = self._containers[environment_type][environment_name]["runtime"]
+
+            if run_as_root:
+                runtime_name = {k for k in self._runtimes if self._runtimes[k].run_as_root}.pop()
+                runtime = self._runtimes[runtime_name].runtime
```
GuillaumeDerval, line 446: Does this mean that you select a random runtime? Why can't we select one in `run_student` ?

GuillaumeDerval (changes requested):

base-containers/base/inginious_container_api/run_student.py
darkyne: I don't get why shared_kernel and run_as_root is incompatible.
As explain before, shared_kernel does not mean there is a shared_kernel between grading_container and student_container. 
Actually, shared_kernel in run_student means the grading_container is running on docker runtime (or Kata if False) and shared_kernel in run_student_intern means the student_cotnainer is running on docker runtime (or kata if False).

So the scenario where a docker grading_container wants to start a kata student_container, means at the run_student file we have shared_kernel and run_as_root both set to true.

Do you want me to rename "shared_kernel" into "run_on_docker" or something like that?

base-containers/base/bin/INGInious
darkyne: "may" is maybe clearer. 
The idea is to note the fact this message will only be received if we start a student_container on kata runtime.
The comment is inproved right now.

inginious/agent/docker_agent/__init__.py
darkyne: Indeed, this implementation forced to use kata runtime if run_as_root otherwise, it selected randomly a non-root runtime.
This, while working in the situation with only docker and kata runtimes, is indeed an issue.
Now, with last modifications, it forces Kata if run_as_root otherwise, it uses the runtime specified by arguments, as it was the case before this pull request.

base-containers/base/inginious_container_api/run_student.py
darkyne: Directly after sending the initial command (without waiting any answer from the student_container), we setup the handler to transfer signals to the student_container. This is exactly how it was already implemented previously (see https://github.com/UCL-INGI/INGInious/blob/3c8b5b333b817cb9741455dd7f0bf2cfaed2cfed/base-containers/base/inginious_container_api/run_student.py#L121).
I did not encounter any bugs up to now with that but if you have an interesting example in mind, feel free to send it to me :)

GuillaumeDerval:
LGTM

anthonygego:
This. Is. Amazing.

PR #703, opened 2021-08-27
Fix subproblems ordering in task editor

anthonygego:
The Jinja json function ignores OrderedDict key order and prints the dict with its keys sorted alphabetically. This uses Python json function instead to ensures the keys are kept in the right order.


Drumor:
LGTM

PR #705, opened 2021-09-13
course_admin/task_edit: fix HTML escaping issues

anthonygego:
This follows #703 that introduced HTML char escaping issues in Javascript strings (<,>, "). In particular, a closing ``</script>`` markup inside a JS string actually stop JS interpreter.

This way, the dict is dumped into JSON format in the Python code to keep the exercises order in the OrderedDict, and the ``tojson`` Jinja filters applies the appropriate escaping that works with ``JSON.parse``.


Drumor:
LGTM

PR #706, opened 2021-09-13
course_admin/task_edit: allows yaml author field to be empty

anthonygego:
Else it will fail as ``task_data.get('author',[])`` will return ``None``.


Drumor:
LGTM

Issue #702, opened 2021-08-27
UI issue when editing context of MCQ subproblem


qdeconinck:
On inginious.org (INGInious v.0.8.dev5+gaa2e61c), the line numbers overwrite the text of the question.

![image](https://user-images.githubusercontent.com/5878547/131166049-5ec7a8dd-7390-4e16-a99f-b4f231f32557.png)

Drumor:
Hello,

Thx for this issue, I can't saddly reproduce it for the moment.
Could u give me the way you produce this? With a description of your environment (browser,...)

Thx!

qdeconinck:
Indeed, some information is missing :-)

The issue arises as a course administrator, when editing the task. The UI issue is only present when I just created a new problem (clicking on the blue `Add` button) and directly editing the context of this new problem. Interestingly, this issue is not present when editing the context of an already existing task. 

I just reproduced this bug on inginious.org using Brave Version 1.29.81 Chromium: 93.0.4577.82 (Official Build) (64-bit)

Drumor:
Thank you, i can reproduce now :-)

PR #708, opened 2021-09-17
Fix UI issue when editing context of MCQ subproblem

Drumor:
Patch for issue #702 

This improves the non fully working hacky fix.


inginious/frontend/static/js/studio.js
```diff
@@ -17,8 +17,7 @@ function studio_load(data)
     });
 
     // Hacky fix for codemirror in collapsable elements
-    var collapsable = $('#tab_subproblems').find('.collapse');
-    collapsable.on('show.bs.collapse', function()
+    $('.card').on('show.bs.collapse',function()
```
anthonygego, line 20: Wouldn't be more appropriate to keep the ``$("#tab_subproblems")`` selector here, to avoid any redraw when unnecessary ?

inginious/frontend/static/js/studio.js
```diff
@@ -343,6 +345,8 @@ function studio_create_new_subproblem()
 
     studio_create_from_template('#' + new_subproblem_type, new_subproblem_pid);
     studio_init_template(new_subproblem_pid, {"type": new_subproblem_type.substring(11)});
+
+    $('#tab_subproblems .card').on('show.bs.collapse',refresh_codemirror);
```
anthonygego, line 349: Looks like a duplicate  of https://github.com/UCL-INGI/INGInious/pull/708/files#diff-dba29ffb384c39f13ccca8801458355aae23552265a109c2bd298ec22887decaR33 ?

Drumor: It's a double call of the function that refresh codemirror element. One is done on studio loading and another is done when a new sub problem is created.

anthonygego:
Please update minified js

Drumor:
Done

PR #713, opened 2021-09-20
Fix containers version

darkyne:
Previously, there was an issue for people trying to install INGInious via pip install.

Indeed, they had to install for exemple version 0.7 but they had no containers images directly provided.
inginious-install command was directly downloading files and building images from the github branch master (thus incompatible with v0.7) and the inginious-container-update command was downloading the images from docker hub with the tag "latest" which was 7 months old.

Now:
-inginious-install command downloads the files and build the images based on the inginious version installed.
-inginious-container-update downloads the image based on the inginious version installed.

Still, both commands are still compatible with dev environement (inginious not installed via pip), in this case:
-inginious-install command will download the files and build the images based on the github branch master (+ advise to rebuild
them manually if you modified local files)
-inginious-container-update simply detects it is in dev mode (inginious isn't "installed") and ask to build the containers manually instead (since no compatible image avaiable on docker hub).

I hope this is clear enough, anyway the scripts are now very verbose to avoid any confusion.

v0.7 and v0.7.1 were also pushed to docker hub.


darkyne:
As usual, I'm definitively bad at github commits management but only last commit is useful -.-

inginious/frontend/installer.py
```diff
@@ -376,20 +375,54 @@ def _build_container(self, name, folder):
...
             self._display_info("Skipping container building.")
             return
 
+        # Get the local images
+        stock_images = []
+        try:
+            docker_connection = docker.from_env()
+            for image in docker_connection.images.list():
+                for tag in image.attrs["RepoTags"]:
+                    if re.match(r"^ingi/inginious-(c-[a-z]+|agent):v" + __version__, tag):
```
anthonygego, line 390: After consideration, this is quite annoying because they are no functional requirement for the container tags to follow a particular name scheme. Moreover, even if we added this constraint, that would not allow third party containers to be pushed on the hub as the ``ingi/`` tag is reserved for the dev team.

Maybe the version check should only be done based on the docker agent version which is embedded as a label in the container. But there should be a way to compare with remote container data on the hub.

darkyne:
While maybe not perfect, I believe this (as it is right now) would already be a good fix.
- inginious-containers-update only dl base+default from docker hub based on the installed inginious version
- inginious-install dl the container files and build locally based on the installed inginious version (or simply dl from master branch if using it in a dev environment).

The doc was also updated in another PR to make it easier for people to understand the scripts action and the way to manually build if necessary.<

Still, since the other (non mandatory) containers are not updated on docker hub, there are not included in the inginious-containers-update (but are handled by inginious-install anyway)

inginious/frontend/installer.py
GuillaumeDerval: +1

darkyne: Well, we changed our mind on this specific question and decided the scripts to only update the mandatory images (c-base and c-default). Third party containers should be handled manually anyway. And non-mandatory containers (such as java etc) will be correctly handled by inginious-install.

darkyne:
@anthonygego 
I believe this was not backported to previous versions yet, no ?

Issue #622, opened 2021-02-04
"New task" doesn't check taskid validity


anthonygego:
The "new task" feature does not check if the given id is valid. It's possible to add a taskid with spaces or ponctuation, which leads to errors.

darkyne:
Implemented in https://github.com/UCL-INGI/INGInious/pull/670.
Suggest to close this.

Drumor:
I think there is still a problem here...

I try to import a task... i have an alert for the creation.
![Screenshot from 2021-08-24 15-58-57](https://user-images.githubusercontent.com/13468630/130630441-6536ae93-df25-4a03-a3f3-7b2742b11623.png)

Drumor:
resolved

PR #714, opened 2021-09-21
fix importing tasks

darkyne:
Fix for issue #622
Now only the "create new task" button checks the syntax for the string input.
"add task" button related to already existing files was wrongly checking the string input on top which was not even related to those files.


inginious/frontend/static/js/task_dispensers.js
```diff
@@ -101,9 +101,9 @@ function dispenser_util_click_modal_task(task) {
     input.attr("checked", !input.attr("checked"));
 }
 
-function dispenser_util_add_tasks_to_section(button) {
+function dispenser_util_add_tasks_to_section(button, checker) {
     task_id= $("#new_task_id").val();
-    if(!task_id.match(/^[a-zA-Z0-9_\-]+$/)){
+    if(checker && !task_id.match(/^[a-zA-Z0-9_\-]+$/)){
```
Drumor, line 106: Why not simply use the value of `existing_task` variable to define if you need to do this verification?

Drumor (changes requested):
I think this can be simplify with existing variable.

inginious/frontend/static/js/task_dispensers.js
darkyne: Damn you are probably right!
I will do this instead :)

Drumor:
In fact there is a tiny indentation issue in the if

darkyne:
No more :)

PR #715, opened 2021-09-21
DB status index

darkyne:
Added an index on "status" to make "update_pending_jobs" faster (takes time filtering on the submission status at every inginious restart).

Also changed "ensure_index" to "create_index" since the first one was depricated since pymongo 3.0


Drumor:
LGTM

Issue #615, opened 2021-01-15
Task accesibility soft deadline


ogoletti:
When setting a start and stop deadline for a task through the web interface, one has to provide a softdeadline to validate the task. This shouldn't be the case.

Drumor:
resolved

PR #716, opened 2021-09-21
No soft_deadline required

darkyne:
By default, when putting no soft_deadline, it was encoding a dummy value (in year 9999) and raised an error message since soft_deafline can not be after hard_deadline.

Now, if soft_deadline was not encoded (or is after hard_deadline), it will simply be equal to the hard_deadline.

Implements issue #615


Drumor:
LGTM

PR #717, opened 2021-09-21
building containers doc

darkyne:
Adding a little bit more information about the way to build or download containers images.
It was apparently causing issues to some people, I hope it is more clear now.
It is also updated with the recent corrections to the scripts.


Drumor (commented):
I think "create_container" part should not be in Teacher documentation but in dev doc. For the rest its ok :)

Drumor:
LGTM

PR #719, opened 2021-09-22
Update search_user.py

### LINKED ISSUE #711: Autocomplete tutors field
The field to add tutors in `admin/COURSE_ID/settings` sometimes behaves strangely. For some tutors, the autocomplete feature does not find them somehow (based on firstname or lastname)... If I ask the tutor for her username, then it will find it easily (and it then works for both firstname and lastname). 
I understand this might be difficult to reproduce I can help more on demand.

Drumor:
Fix #711  - Make request case insensitive


anthonygego:
LGTM

PR #721, opened 2021-10-05
agents specific for ssh

darkyne:
Added a checkbox in the environment configuration tab to inform that the task will use ssh.
Such tasks will only be handled by agents with the specific flag ("ssh_allowed").
This means, if multiple ssh tasks are running, it would only block the ssh_allowed agents and not the other ones.

DockerAgent has now a new argument, ssh_allowed: bool

Doc was also slightly updated.

### DIFF
--- a/inginious/backend/topic_priority_queue.py
+++ b/inginious/backend/topic_priority_queue.py
@@ -36,10 +36,11 @@ class TopicPriorityQueue:
         :return: the smallest elements that fits in one of the topics
         :raises: queue.Empty exception if the queue has no elements that fits in any of the topics
         """
+        agent_ssh = topics.ssh_allowed
         best_topic = None
         best_elem = None
-        for topic in topics:
-            if topic in self.queues and len(self.queues[topic]) != 0 and (best_elem is None or best_elem > self.queues[topic][0]):
+        for topic in topics.environments:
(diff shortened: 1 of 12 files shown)


GuillaumeDerval:
I don't see in the code where you define "allow_ssh" to true for the agent that starts with the webapp when you have a default installation. Did you ensure that it is the case?

darkyne:
The default start with inginious-webapp uses the function create_arch that will start the agent with ssh_allowed set to true.
Obviously this is for local deployment and testings. When they have to start manually the agents, they will have to add the --ssh option for some of the agents if they want to use ssh features.

GuillaumeDerval:
LGTM then

inginious/backend/topic_priority_queue.py
```diff
@@ -36,10 +36,11 @@ def get(self, topics=None):
         :return: the smallest elements that fits in one of the topics
         :raises: queue.Empty exception if the queue has no elements that fits in any of the topics
         """
+        agent_ssh = topics.ssh_allowed
```
GuillaumeDerval, line 39: TopicPriorityQueue is a reusable component, that could even be moved to the `common` package; it shouldn't be aware of the content of the objects given to it.

It's a bad smell, these exception should be handled elsewhere.

PR #725, opened 2021-10-08
Administration

Drumor:
This is a first implem of administration part. This lets admins create/delete/activate users. It's also possible to remove bindings for specific users.


anthonygego:
Performance for +2k users are still terrible. A pagination system should be implemented. Maybe some parts of the submissions pagination mechanism can be refactored for other pages.

inginious/frontend/pages/admin/admin.py
```diff
@@ -0,0 +1,80 @@
...
+        return self.show_page()
+
+    def show_page(self):
+        """Display page"""
+        all_users = self.user_manager.get_users()
+        return self.template_helper.render("admin/admin_users.html", all_users=all_users)
+
+
+class AdministrationUserActionPage(INGIniousAdministratorPage):
+    """Action on User Admin page."""
```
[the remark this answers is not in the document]
anthonygego: There should be no need to define the POST methods if the classes inherit from ``INGIniousPage``.

inginious/frontend/pages/admin/admin.py
```diff
@@ -0,0 +1,81 @@
...
+            _, _ = self.user_manager.revoke_binding(username, binding_id)
+        elif action == "add_user":
+            realname = request.form.get("realname")
+            email = request.form.get("email")
+            password = request.form.get("password")
+            feedback = self.user_manager.create_user({
+                "username": username,
+                "realname": realname,
+                "email": email,
+                "password": hashlib.sha512(password.encode("utf-8")).hexdigest(),
```
anthonygego, line 75: We should avoid to copy/paste the password hash function in the code, in case we need to migrate to another cryptographic standard in the future.

anthonygego (changes requested):

inginious/frontend/pages/admin/admin.py
Drumor: I defined a hash method and replace all to simplify edition if a migration is necessary.

inginious/frontend/pages/admin/admin.py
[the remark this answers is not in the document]
Drumor: in fact i have to do that cause of the ajax requests. Am i able to set method argument of an ajax request to "POST_AUTH"?
In case of no redefinition those request lost there way.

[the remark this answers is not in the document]
anthonygego: That's probably because the session is lost. The ajax request should send the cookies normally, as this is done for other such requests in the code (submission progress polling, ...). You should check what's going on when the request is processed by the ``INGIniousAuthPage.POST``method.

Remark 2. implies that anyone can perform the requests, which is a major security and privacy flaw.

inginious/common/base.py
```diff
@@ -55,6 +55,14 @@ def get_json_or_yaml(file_path, content):
         return inginious.common.custom_yaml.dump(content)
 
 
+def hash_content(content):
```
anthonygego, line 58: This should have a more explicit name (``hash_password`` for instance), and be located in the frontend files. This whole file probably shouldn't be modified.

inginious/frontend/pages/admin/admin.py
```diff
@@ -0,0 +1,76 @@
...
+                yield {k: data[k] for k in islice(_it, size)}
+
+        page = int(request.form.get("page")) if request.form.get("page") is not None else 1
+
+        all_users = self.user_manager.get_users()
+        size_users = len(all_users)
+        user_per_page = 10
+        pages = size_users // user_per_page + (size_users % user_per_page > 0)
+        subdicts = list(_chunks(all_users, user_per_page))
+        display_users = subdicts[page - 1]
```
anthonygego, line 40: Although it seems to perform not so bad with thousands of users and a local database, you should not fetch all users, parse the whole set of data and then throw everything but 10 results.

inginious/frontend/user_manager.py
```diff
@@ -384,7 +410,29 @@ def get_user_api_key(self, username, create=True):
...
+        user = self._database.users.find_one({"username": username})
+        return user["activate"] if user and "activate" in user else None
+
+    def activate_user(self, activate_hash):
+        """ Activates user """
+        error = False
+        user = self._database.users.find_one_and_update({"activate": activate_hash}, {"$unset": {"activate": True}})
+        if user is None:
+            error = True
+        return error
```
anthonygego, line 427: ``return user is None``. Add more details to docstring.

inginious/frontend/pages/utils.py
```diff
@@ -229,6 +232,35 @@ def preview_allowed(self, *args, **kwargs):
...
+
+    def GET(self, *args, **kwargs):
+        """
+        Checks if user is superadmin and calls GET_AUTH or performs logout.
+        Otherwise, returns the login template.
+        """
+        username = self.user_manager.session_username()
+        if self.user_manager.session_logged_in():
+            if not self.user_manager.user_is_superadmin(username):
+                return self.template_helper.render("forbidden.html", message=_("You have not eligible right to see this part"))
```
anthonygego, line 248: You can be eligible, but eligible right has no meaning. There should already be such forbidden access messages elsewhere that don't need to be repeated to ease localization.

inginious/frontend/templates/admin/admin_users.html
```diff
@@ -0,0 +1,334 @@
...
+            </div>
+            <div class="modal-footer">
+                <input type="hidden" name="username" id="username">
+                <button type="button" class="btn btn-default" data-dismiss="modal">{{_("Cancel")}}</button>
+            </div>
+        </div>
+    </div>
+</div>
+
+<script type="text/javascript">
```
anthonygego, line 214: That's a lot of JS code. This should go in its own js file, and be included in the minified one. Use function names and the onclick attribute instead of setting the event listener at runtime. This will lessen the risk of regression in case an id changes and be more coherent with the remaining of the code.

inginious/frontend/templates/layout.html
```diff
@@ -122,6 +122,12 @@
                             {# Change to teacher privilege when created #}
                             {% if user_manager.user_is_superadmin() %}
                                 <a class="list-group-item list-group-item-action" href="{{ get_homepath() }}/marketplace"><i class="fa fa-shopping-bag fa-fw"></i> {{ _("Marketplace") }}</a>
+                                <a class="list-group-item list-group-item-action dropdown-toggle" data-toggle="collapse" href="#collapseAdministratorMenu" role="button" aria-expanded="false" aria-controls="collapseAdministratorMenu">
+                                    <i class="fa fa-shield fa-fw"></i> {{ _("Administration") }} <span class="caret"></span>
+                                </a>
+                                <div class="collapse list-group-collapse" id="collapseAdministratorMenu">
+                                    <a class="list-group-item list-group-item-action" href="{{ get_homepath() }}/administrator/users"><i class="fa fa-users fa-fw"></i> {{ _("Users") }}</a>
```
anthonygego, line 129: Marketplace can be merged into this administrator menu.

inginious/frontend/templates/admin/admin_users.html
```diff
@@ -0,0 +1,334 @@
...
+                        {% endif %}
+                    </div>
+                </td>
+            </tr>
+        {% endfor %}
+        </tbody>
+    </table>
+
+    <!-- Pagination -->
+<div class="d-flex justify-content-center">
```
anthonygego, line 95: This does not really display centered.

inginious/frontend/templates/admin/admin_users.html
```diff
@@ -0,0 +1,334 @@
...
+        $.ajax({
+          type: "POST",
+          url: "{{get_homepath()}}/administrator/user_action",
+          data: {
+            "username":username,
+            "action":action,
+            "binding_id": binding_id
+          },
+        }).done(function() {
+          window.location.href = "{{get_homepath()}}/administrator/users";
```
anthonygego, line 294: Is this necessary ? Closing the modal isn't enough ? You will loose the current displayed page information here.

inginious/frontend/templates/admin/admin_users.html
```diff
@@ -0,0 +1,334 @@
...
+        });
+    }
+    function get_bindings(username){
+        var returned_value = {};
+        $.ajax({
```
anthonygego, line 299: =>> ``$.post`` is simpler. You can also remove the useless arguments (async...)

inginious/frontend/templates/admin/admin_users.html
```diff
@@ -0,0 +1,334 @@
...
+          url: "{{get_homepath()}}/administrator/user_action",
+          data: {
+            "username":username,
+            "realname":realname,
+            "email": email,
+            "password":password,
+            "action":action
+          },
+        }).done(function(feedback) {
+            if (feedback == ""){
```
anthonygego, line 274: That's a pretty strange and insecure way to confirm things are done.

inginious/frontend/templates/admin/admin_users.html
```diff
@@ -0,0 +1,334 @@
...
+            $("#"+elem+"-template .binding_revoke").data("binding",elem);
+            $("#"+elem+"-template .binding_identifier").text(bindings[elem][0]);
+            $("#"+elem+"-template .binding_method").text(elem);
+        }
+        $('.binding_revoke').bind("click", revoke_binding);
+    });
+    $('#binding_modal').on('hide.bs.modal', function (event) {
+        $("#binding_content").html('');
+    });
+    $(".delete_user").each(function(index) {
```
anthonygego, line 245: This can stay here as it is very specific.

inginious/frontend/user_manager.py
```diff
@@ -288,15 +290,25 @@ def auth_user(self, username, password):
...
 
         return user if user is not None and self.connect_user(username, user["realname"], user["email"],
-                                                              user["language"], user.get("tos_accepted", False)) else None
+                                                              user["language"],
+                                                              user.get("tos_accepted", False)) else None
+
+    def is_user_activated(self, username):
+        """
+        Verify if user is activated
+        :param username: Username/Login
```
anthonygego, line 307: You should clearly allow for several usernames as you currently perform one request per user in a loop.

inginious/frontend/user_manager.py
```diff
@@ -324,6 +336,19 @@ def disconnect_user(self):
...
         self._destroy_session()
 
+    def get_users(self):
+        """
+        list all users
+        :return A dictionary with info of users
+        """
+        retval = {}
+        infos = self._database.users.find()
+        for info in infos:
```
anthonygego, line 346: See remark above: do not fetch everyone if you don't need it.

inginious/frontend/user_manager.py
```diff
@@ -430,6 +478,54 @@ def bind_user(self, auth_id, user):
...
+    def create_user(self, values):
+        """
+        Create a new user
+        :param values: Dictionary of fields
+        :return: An error message if something went wrong else None
+        """
+        already_exits_user = self._database.users.find_one(
+            {"$or": [{"username": values["username"]}, {"email": values["email"]}]})
+        if already_exits_user is not None:
+            return _("We were not able to create user with given params.")
```
anthonygego, line 520: Use passive tense here.

anthonygego (changes requested):

inginious/frontend/user_manager.py
```diff
@@ -430,6 +479,54 @@ def bind_user(self, auth_id, user):
...
+        """
+        Revoke a binding method for a user
+        :param binding_id: The binding method id
+        :param username: username of the user
+        :return: Boolean if error occurred and message if necessary
+        """
+        user_data = self._database.users.find_one({"username": username})
+        if binding_id not in self.get_auth_methods().keys():
+            error = True
+            msg = _("Incorrect authentication binding.")
```
anthonygego, line 492: It seems error messages are not shown when this happens.

inginious/frontend/pages/course_admin/danger_zone.py
```diff
@@ -215,7 +216,7 @@ def get_backup_list(self, course):
 
     def page(self, course, msg="", error=False):
         """ Get all data and display the page """
-        thehash = hashlib.sha512(str(random.getrandbits(256)).encode("utf-8")).hexdigest()
+        thehash = self.user_manager.hash_password(str(random.getrandbits(256)))
```
anthonygego, line 219: Is the method a class method or not ? Above you use ``UserManager.hash_password``. This is not really needed as the user manager is a singleton, but it would be nice to choose and adapt the code accordingly.

inginious/frontend/pages/utils.py
```diff
@@ -229,6 +232,36 @@ def preview_allowed(self, *args, **kwargs):
...
+    def POST(self, *args, **kwargs):
+        """
+        Checks if user is superadmin and calls POST_AUTH.
+        Otherwise, returns the forbidden template.
+        """
+
+        username = self.user_manager.session_username()
+        if self.user_manager.session_logged_in() and self.user_manager.user_is_superadmin(username):
+            return self.POST_AUTH()
+        return self.template_helper.render("forbidden.html", message=_("You have not eligible right to see this part"))
```
anthonygego, line 262: A right is not eligible.

inginious/frontend/templates/admin/admin_users.html
```diff
@@ -0,0 +1,257 @@
+{# This file is part of INGInious. See the LICENSE and the COPYRIGHTS files for #}
+{# more information about the licensing of this file. #}
+
+{% extends "layout.html" %}
+{% block title %}{{_("Users Administration") }}{% endblock %}
+{% block column %}
+    {% if not user_manager.session_logged_in() %}
+        {% include "signin_button.html" %}
```
anthonygego, line 8: Probably not very useful condition here.

inginious/frontend/user_manager.py
```diff
@@ -451,6 +499,54 @@ def bind_user(self, auth_id, user):
...
+        elif len(user_data.get("bindings", {}).keys()) > 1 or "password" in user_data:
```
anthonygego, line 513: This will throw an exception if ``username`` is not found in database and ``user_data`` is ``None``.

inginious/frontend/pages/admin/admin.py
```diff
@@ -0,0 +1,68 @@
...
+        action = request.form.get("action")
+        feedback = None
+        if action == "activate":
+            activate_hash = self.user_manager.get_user_activate_hash(username)
+            self.user_manager.activate_user(activate_hash)
+        elif action == "delete":
+            self.user_manager.delete_user(username)
+        elif action == "get_bindings":
+            user_info = self.user_manager.get_user_info(username)
+            return jsonify(user_info.bindings)
```
anthonygego, line 51: This will probably throw an exception if ``username`` is not found in database, as there won't be any ``bindings`` attribute accessible.

inginious/frontend/pages/admin/admin.py
```diff
@@ -0,0 +1,68 @@
...
+            feedback = self.user_manager.create_user({
+                "username": username,
+                "realname": realname,
+                "email": email,
+                "password": password,
+                "bindings": {},
+                "language": "en"})
+        if feedback:
+            return jsonify({"error": True, "message": feedback})
+        return jsonify({"message": ""})
```
anthonygego, line 68: For any unmanaged ``action`` passed in the form, this will return successfully.

inginious/frontend/pages/admin/admin.py
```diff
@@ -0,0 +1,68 @@
...
+        """ Display admin users page"""
+        return self.show_page()
+
+    def show_page(self):
+        """Display page"""
+
+        page = int(request.form.get("page")) if request.form.get("page") is not None else 1
+        user_per_page = 10  # TODO probably better to let user define user_per_page
+        all_users, size_users = self.user_manager.get_users(limit=user_per_page, skip=(page-1)*user_per_page)
+        activated_users = self.user_manager.get_activated_users(list(all_users))
```
anthonygego, line 30: This acts like a filter on previous results but still make new requests. It's probabl worth adding an ``activated`` field to the ``UserInfo`` class in order to avoid this inefficient behaviour.

inginious/frontend/user_manager.py
```diff
@@ -940,3 +1037,16 @@ def has_staff_rights_on_course(self, course, username=None, include_superadmins=
...
+
+    @classmethod
+    def hash_password(cls, content):
+        """
+        :param content: a str input
+        :return a hash of str input
+        """
+        return hashlib.sha512(content.encode("utf-8")).hexdigest()
+
+    def get_activated_users(self, usernames):
```
anthonygego, line 1049: See comment above.

inginious/frontend/user_manager.py
```diff
@@ -341,6 +352,20 @@ def disconnect_user(self):
...
+        """
+        list all users
+        :param limit A limit of users requested
+        :param skip A quantity of users to skip
+        :return A dictionary with info of users
+        """
+        retval = {}
+        infos = self._database.users.find().skip(skip).limit(limit)
+        for info in infos:
+            retval[info["username"]] = UserInfo(info["realname"], info["email"], info["username"], info["bindings"],
```
anthonygego, line 365: Maybe this could be done using a dict comprehension.

inginious/frontend/user_manager.py
```diff
@@ -351,7 +376,8 @@ def get_users_info(self, usernames) -> Dict[str, Optional[UserInfo]]:
 
         infos = self._database.users.find({"username": {"$in": remaining_users}})
         for info in infos:
-            retval[info["username"]] = UserInfo(info["realname"], info["email"], info["username"], info["bindings"], info["language"])
+            retval[info["username"]] = UserInfo(info["realname"], info["email"], info["username"], info["bindings"],
```
anthonygego, line 379: BTW, maybe this could also be changed to a dict comprehension. I'm however a bit concerned as this code looks like the new ``get_users`` function you defined. There may be some useless code duplication here.

inginious/frontend/templates/admin/admin_users.html
```diff
@@ -0,0 +1,254 @@
...
+                <span class="sr-only">{{_("(current)")}}</span></a></li>
+        </ol>
+    </nav>
+{% endblock %}
+{% block content %}
+
+
+<h2>{{ _("Users Administration") }}</h2>
+<div>
+    <!--<a id="add_user" class="btn-sm btn btn-info" href="{{ get_homepath() }}/administrator/user_add"><i class="fa fa-user-plus"></i>{{_("Create User")}}</a>-->
```
anthonygego, line 22: Commented code spotted.

inginious/frontend/templates/admin/admin_users.html
```diff
@@ -0,0 +1,254 @@
...
+                    <a href="mailto:{{all_users[user]['email']}}">{{all_users[user]["email"]}}</a>
+                </td>
+                <td id="binding_{{all_users[user]['username']}}"  style="display:none;" data-bindings="{{all_users[user]['bindings']}}">
+                </td>
+                <td>
+                    <div class="btn-group btn-group-sm" role="group">
+                        <button type="button" data-username="{{all_users[user]['username']}}"
+                                    data-type="single" data-toggle="modal"
+                                    data-target="#binding_modal"
+                                    class="btn btn-secondary binding_user" title="Bindings"><i class="fa fa-cloud"></i></button>
```
anthonygego, line 69: ``title``s are not wrapped with gettext here.

inginious/frontend/templates/admin/admin_users.html
```diff
@@ -0,0 +1,254 @@
...
+            <div class="pull-right">
+            <a href='#' class='binding_revoke' data-username="" data-binding="" data-toggle="tooltip" data-placement="bottom" title="{{ _('Revoke access') }}">
+                <i style="color:red;" class="fa fa-trash-o fa-fw"></i>
+            </a>
+            </div>
+        </div>
+        <div class="card-body">
+            <div class="row">
+                <div class="col-md-11">
+                    <p>Identifier :<span class="binding_identifier"></span></p>
```
anthonygego, line 247: gettext wrapping missing here too.

inginious/frontend/static/js/admin.js
```diff
@@ -0,0 +1,77 @@
...
+    $.post({
+      url: "user_action",
+      data: {
+        "username":username,
+        "action":"revoke_binding",
+        "binding_id": binding_id
+      },
+    }).done(function(feedback) {
+        if (!feedback.hasOwnProperty('error')){
+            window.location.href = "users";
```
anthonygego, line 17: Do you actually need this redirection ? Is this working without any prefix (hostname and path prefix) required ?

anthonygego (commented):
Stay strong, this is almost done.

inginious/frontend/static/js/admin.js
Drumor: I used this to close (and refresh page) the modal after hitting submit button. If i remove this, the page just stay on the open modal. Yeah working without prefix :)

anthonygego: If you want to close the modal and refresh the page, maybe you should set the button ``type`` attribute to ``submit``.

Drumor: As we said in private, it displays a feedback in the modal so I cannot use the `type="submit"` solution. I mark this as resolved for now.

inginious/frontend/pages/admin/admin.py
anthonygego: There is still no error message returned for unknown ``action``.

Drumor: Solved :)

anthonygego (changes requested):
When you revoke a binding and get an error message for whatever reason, this error message is not cleaned when the modal box is closed.

Printing it in a red alert box instead of just displaying the message black on white would also be nice to better catch the user attention.

inginious/frontend/static/js/admin.js
```diff
@@ -0,0 +1,79 @@
...
+        }
+        $('#feedback_bindings').text(feedback['message']);
```
anthonygego, line 19: Put this in an ``else`` statement. Otherwise, they are executed before the redirection is effective.

anthonygego (changes requested):
This should be good after this last fix. Please also clean your commits. This is a bit too big to be squashed in a unique commit.

inginious/frontend/user_manager.py
```diff
@@ -451,6 +489,56 @@ def bind_user(self, auth_id, user):
...
+                {"$unset": {"bindings." + binding_id: 1}},
+                return_document=ReturnDocument.AFTER)
+            msg = ""
+            error = False
+        else:
+            error = True
+            msg = _("You must set a password before removing all bindings.")
+        return error, msg
+
+    def delete_user(self, username):
```
anthonygego, line 515: This removes the user from the ``users`` collection but does not removes any other trace from the database, and would allow any new user to get an old submission history by selecting the right username at registration time.

The appropriate piece of code already exists and can be fetched from : https://github.com/UCL-INGI/INGInious/blob/master/inginious/frontend/pages/preferences/delete.py#L17

anthonygego (changes requested):
Sorry but I can't merge this with such a security issue remaining.

inginious/frontend/user_manager.py
```diff
@@ -512,14 +512,24 @@ def revoke_binding(self, username, binding_id):
...
-        return result.deleted_count == 1
+        query = {"username": username, "email": confirmation_email} \
+            if confirmation_email is not None else {"username": username}
+        result = self._database.users.find_one_and_delete(query)
+        if not result:
+            return False
+        else:
+            self._database.submissions.delete_many({"username": username})
+            self._database.user_tasks.delete_many({"username": username})
+            all_courses = self._database.courses.find({})
```
anthonygego, line 530: It would be better to only select the user's courses, even if the ``unregister_user`` method can be run with non-registered users, to avoid additional useless requests ans false logging messages (User xxx unregistered from...)

Drumor: I'm intent to resolve this like that:
`user_courses = [course['_id'] for course in self._database.courses.find({})
                            if username in course["students"]]`
Seems good to you?

anthonygego: You can simply use the username as filter. ``{"students": username}``

anthonygego (approved):
LGTM. Can you squash the current changes and update the minified js file ?

PR #744, opened 2021-11-18
[frontend/task_factory] Add translation files back to cache date watcher

### LINKED ISSUE #741: Task factory does not check translation updates anymore
This prevents displaying translation changes in realtime as the cache is not updated when needed.

anthonygego:
Fix #741.

Regression from course and task factories refactor.
Taken from 9be0653


Drumor:
LGTM

PR #745, opened 2021-11-19
LocalFSProvider: remove slash addition in from_subfolder

nrybowski:
Adding a trailing slash at the end of the parent prefix is not required since the \_\_init\_\_ method of the parent LocalFSProvider already did it.

https://github.com/UCL-INGI/INGInious/blob/8d49326d6e73ce887519dff518735d709c099685/inginious/common/filesystems/provider.py#L35-L39


Drumor:
LGTM

PR #748, opened 2021-11-25
use correct id to set value of task_id

### LINKED ISSUE #742: Reset state button...does not reset state ?
It seems the reset state button actually does not reset the state. The field in the ``user_tasks`` entry remains the same.

Drumor:
fix #742


anthonygego (approved):
LGTM

PR #753, opened 2021-12-08
Update mongo_sessions.py

Drumor:
For some reason this condition handle an ``AttributeError: 'function' object has no attribute 'view_class'`` on task edition (probably not only).
This PR propose a quickfix with an attribute verififcation.


anthonygego:
I guess the actual issue is that the requested endpoint is not registered. This can happen if we request an inexistant resource (=> file not found) but this is strange for the task edition page. Can you investigate this further ?

We shouldn't merge this until "some reason" becomes identified.

anthonygego:
@Drumor you can check this and merge if it's OK.

Drumor:
Looks good but didn't test it

PR #756, opened 2021-12-10
[filesystems] Cosmetic updates

nrybowski:
This PR adds some documentation to the `FileSystemProvider` abstract class and its "local" implementation.

It also moves the abstract class from the `provider.py` file to the `__init__.py` one.
The goal is to reduce the amount of empty files and apply a common package structure.
See: https://github.com/UCL-INGI/INGInious/blob/441885b26ae96765a460bdc2eac6e4e6c8cd664b/inginious/agent/__init__.py#L46

Finally, some cleanup has been applied after a `pylint` run on the files.


doc/conf.py
```diff
@@ -36,7 +36,9 @@
     'sphinx.ext.autodoc',
     'sphinx.ext.viewcode',
     'sphinx_tabs.tabs',
-    'IPython.sphinxext.ipython_console_highlighting'
+    'IPython.sphinxext.ipython_console_highlighting',
+    'sphinx.ext.autodoc',
```
anthonygego, line 40: Already imported at line 36

doc/conf.py
```diff
@@ -36,7 +36,9 @@
...
+    'sphinx_autodoc_typehints'
```
anthonygego, line 41: Is this a builtin extension or should it be added to the setup file ?

inginious/common/filesystems/local.py
```diff
@@ -1,10 +1,11 @@
 # coding=utf-8
+from __future__ import annotations
```
anthonygego, line 2: The current requirements is python 3.6. I think it would be nice to keep this support while EL7 is still supported and ships python 3.6.

nrybowski: As discussed previously, python3.6 reached its EOL (https://www.python.org/dev/peps/pep-0494/#lifespan) and python3.8 is supported by EL7 (https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/developer_guide/setting-up_setup-developing-python) so we will bump the requirements to python3.8.

nrybowski:
Rebase on https://github.com/UCL-INGI/INGInious/commit/262e8d92fc4a2f84e38cd047f6adc17cb4d892a1 and add required fixes.

doc/conf.py
nrybowski: Added to the setup file in https://github.com/UCL-INGI/INGInious/pull/756/commits/9033abcd70f990f1379ce5666df725ca5e0bca0e

doc/conf.py
nrybowski: Duplicate removed in https://github.com/UCL-INGI/INGInious/pull/756/commits/9033abcd70f990f1379ce5666df725ca5e0bca0e.

anthonygego (approved):
LGTM

PR #760, opened 2021-12-17
Fix denial of service by job deletion

anthonygego:
Namedtuple ``_replace`` function returns a new instance. Therefore, ``job.msg`` is never set to ``None`` is the waiting list, resulting in exceptions during the queue update after reserving an agent execution slot, never releasing it. The only way to recover is to restart the backend, or, in the best case, the agent.

This proposes a fix that directly removes the job from the waiting list and check if it's still there after popping the job id from the priority queue.

### DIFF
--- a/inginious/backend/backend.py
+++ b/inginious/backend/backend.py
@@ -173,12 +173,8 @@ class Backend(object):
 
         #jobs_waiting: a list of tuples in the form
         #(job_id, is_current_client_job, info, launcher, max_time)
-        jobs_waiting = list()
-
-        for job in self._waiting_jobs.values():
-            if isinstance(job.msg, ClientNewJob):
-                jobs_waiting.append((job.job_id, job.client_addr == client_addr, job.msg.course_id+"/"+job.msg.task_id, job.msg.launcher,
-                                     self._get_time_limit_estimate(job.msg)))
+        jobs_waiting = [(job.job_id, job.client_addr == client_addr, job.msg.course_id+"/"+job.msg.task_id, job.msg.launcher,
+                                     self._get_time_limit_estimate(job.msg)) for job in self._waiting_jobs.values()]
 
         await ZMQUtils.send_with_addr(self._client_socket, client_addr, BackendGetQueue(jobs_running, jobs_waiting))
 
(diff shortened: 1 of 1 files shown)


inginious/backend/backend.py
```diff
@@ -174,12 +173,8 @@ async def handle_client_get_queue(self, client_addr, _: ClientGetQueue):
 
         #jobs_waiting: a list of tuples in the form
         #(job_id, is_current_client_job, info, launcher, max_time)
-        jobs_waiting = list()
-
-        for job in self._waiting_jobs.values():
-            if isinstance(job.msg, ClientNewJob):
```
GuillaumeDerval, line 180: this check disappeared. Expected?

inginious/backend/backend.py
```diff
@@ -138,17 +138,16 @@ async def handle_client_new_job(self, client_addr, message: ClientNewJob):
         self._logger.info("Adding a new job %s %s to the queue", client_addr, message.job_id)
         job = WaitingJob(message.priority, time.time(), client_addr, message.job_id, message)
         self._waiting_jobs[message.job_id] = job
-        self._waiting_jobs_pq.put((message.environment_type, message.environment, message.environment_parameters["ssh_allowed"]), job)
+        self._waiting_jobs_pq.put((message.environment_type, message.environment, message.environment_parameters["ssh_allowed"]), message.job_id)
```
GuillaumeDerval, line 141: If you only push the job_id, the heap will be ordered by job_id. You should push a tuple (priority, time of arrival in the queue, job_id) or something like that instead.

GuillaumeDerval (changes requested):

inginious/backend/backend.py
anthonygego: Since the message is no more supposed to be replaced by ``None``, the ``job.msg`` type in ``WaitingJob``tuple is always supposed to be a ``ClientNewJob``.

GuillaumeDerval:
LGTM

inginious/backend/backend.py
anthonygego: Fixed

PR #764, opened 2022-01-03
[pymongo] migrate to pymongo 4

anthonygego:
Several major deprecations are made effective from pymongo 4.
These changes are made according to this doc : https://pymongo.readthedocs.io/en/stable/migrate-to-pymongo4.html.

I suggest to restrict the pymongo version of the older branches to <4 to avoid any issue.


Drumor:
LGTM
Note : This will certainly break outside plugins.

PR #775, opened 2022-01-28
[tests/filesystems] Add tests and refactor fixtures

nrybowski:
- Add tests for:
  - move()
  - get_last_modification_time()
  - get()
  - put()
  - get_needed_args()
  - init_from_args()
  - copy_to()

- Refactor fixtures:
  copy_{to,from}() requires files on the disk, outside of the FSProvider prefix.  In order to avoid redundancy in the fixtures, factories are introduced as well as  specialized fixtures.
  Some tests case are also parametrized to avoid test code duplication.

### DIFF
--- a/inginious/common/filesystems/__init__.py
+++ b/inginious/common/filesystems/__init__.py
@@ -136,7 +136,7 @@ class FileSystemProvider(metaclass=ABCMeta):
     @abstractmethod
     def distribute(self, filepath, allow_folders=True):
         """ Give information on how to distribute a file. Provides Zip files of folders. Can return:
-            ("file", mimetype, fileobj) where fileobj is an object-like file (with read()) and mimetype its mime-type.
+            ("local", mimetype, fileobj) where fileobj is an object-like file (with read()) and mimetype its mime-type.
             ("url", None, url) where url is a url to a distant server which possess the file.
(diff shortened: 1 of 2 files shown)


nrybowski:
Issues found by the new tests are fixed in https://github.com/UCL-INGI/INGInious/pull/774.

nrybowski:
Rebase on 7cf81aaf25dd1ed8ffd9d59153a5a89489d23c2e and 100% coverage of LocalFSProvider.

nrybowski:
Rebase on 98b8e3341b3dfdd3614e2ac054c00ef04f51e321.

anthonygego (commented):
Sometimes ``os.path`` is used, sometimes not. 

While the whole codebase is currently aimed at running on Linux, I think it would be best practice to avoid always making this assumption and worsen code compatibility in case someone would someday want to make the code runnable on Windows (In the past, the frontend was running fine on it). 

Therefore, I'd rather rely the standard library ``os.path`` to compute path and basename instead of playing with concatenations of ``/`` and splits.

nrybowski:
The verification of the FUT's sides-effects should be more generic and called for each test case.
I will add that in next commits.

anthonygego:
I'm closing this as this hasn't moved since 3 months. Feel free to re-open it when you are able to work on it.

PR #785, opened 2022-02-16
Audience insertion via CSV file

Drumor:
This PR have for purpose to let the teacher of the course import his audiences through a CSV file.
CSV structure is as follow:
A description, a list of students (username or mail separated by a ';'), a list of tutors (username or mail separated by a ';')

### DIFF
--- a/doc/teacher_doc/course_admin.rst
+++ b/doc/teacher_doc/course_admin.rst
@@ -135,35 +135,23 @@ The student list is entirely managed by drag-and-drop.
 Course structure upload
 ```````````````````````
 
-You can generate the course audience structure with an external tool and then upload
-it on INGInious. This is done with a YAML file, which structure is described below.
-The course structure can be uploaded on the audience list view in the course administration.
+You can generate the course audience structure and populate it with a csv file upload
+on INGInious. The structure of the attented file is described below.
+The file can be uploaded on the audience list view in the course administration.
 
-Audiences YAML structure
-*************************
+CSV file structure
+******************
+
+The file doesn't need any header and is populate based on a person per line structure:
 
 ::
 
-    -    description: Audience 1
-         tutors:
-                 - tutor1
-                 - tutor2
-         students:
-                 - user1
-                 - user2
-    -    description: Audience 2
-         tutors:
-                 - tutor1
-                 - tutor2
-         students:
-                 - user3
-                 - user4
+    user_id | id_field | user_role | description
 
--   *description* is a string and corresponds to your audience description
--   *tutors* is a list of strings representing the usernames of the
-    assigned audience tutors.
--   *students* is a list of strings representing the usernames of the
-    audience students.
+-   *user_id* is a string that identify the user. Based on the id_field this may take several form. Most common is the username but is not the only one. you may, for example, give email to identify the user.
+-   *id_field* is a string that corresponds to the type of user_id. It can be values used on user creation (like username, email, realname) but can also be one of the authentication method bindings (for e.g. : bindings.linkedin).
+-   *user_role* is a string that simply represent the role of the user. it can be ``student`` or ``tutor``
+-   *description* is a string and corresponds to your audience description. Line with the same description will be merged within the same audience.
 
 Groups
 ------
--- a/doc/teacher_doc/course_admin.rst
+++ b/doc/teacher_doc/course_admin.rst
@@ -151,7 +151,7 @@ The file doesn't need any header and is populate based on a person per line stru
 A comma ``,`` is used to separate the items while the quote char is ``"``.
 
 -   *user_id* is a string that identify the user. Based on the id_field this may take several form. Most common is the username but is not the only one. you may, for example, give email to identify the user.
--   *id_field* is a string that corresponds to the type of user_id. It can be values used on user creation (like username, email, realname) but can also be one of the authentication method bindings (for e.g. : bindings.linkedin).
+-   *id_field* is a string that corresponds to the type of user_id. It can be values used on user creation (like username, email).
 -   *user_role* is a string that simply represent the role of the user. it can be ``student`` or ``tutor``
 -   *description* is a string and corresponds to your audience description. Line with the same description will be merged within the same audience.
 
--- a/doc/teacher_doc/course_admin.rst
+++ b/doc/teacher_doc/course_admin.rst
@@ -146,12 +146,16 @@ The file doesn't need any header and is populate based on a person per line stru
 
 ::
 
-    user_id | id_field | user_role | description
+    user_id , id_field , user_role , description
+
+A comma ``,`` is used to separate the items while the quote char is ``"``.
 
 -   *user_id* is a string that identify the user. Based on the id_field this may take several form. Most common is the username but is not the only one. you may, for example, give email to identify the user.
 -   *id_field* is a string that corresponds to the type of user_id. It can be values used on user creation (like username, email, realname) but can also be one of the authentication method bindings (for e.g. : bindings.linkedin).
 -   *user_role* is a string that simply represent the role of the user. it can be ``student`` or ``tutor``
 -   *description* is a string and corresponds to your audience description. Line with the same description will be merged within the same audience.
+
+
 
 Groups
 ------
--- a/inginious/frontend/pages/course_admin/student_list.py
+++ b/inginious/frontend/pages/course_admin/student_list.py
@@ -54,7 +54,6 @@ class CourseStudentListPage(INGIniousAdminPage):
         data["delete"] = flask.request.form.getlist("delete")
         data["groupfile"] = flask.request.files.get("groupfile")
         data["audiencefile"] = flask.request.files.get("audiencefile")
-        data["audiencecreationfile"] = flask.request.files.get("audiencecreationfile")
         error = {}
         msg = {}
         active_tab = "tab_students"
@@ -199,35 +198,45 @@ class CourseStudentListPage(INGIniousAdminPage):
             active_tab = "tab_audiences"
 
         try:
-            if "upload_audiences" in data or "audiences" in data:
-                errored_students = []
-                if "upload_audiences" in data:
+            if "audiencefile" in data and 'upload_audiences_creation' in data:
+                # get the Werkzeug datastructures.FileStorage object.
+                # The stream of this object is the stream body of the uploaded file.
+                # Furthermore, FileStorage.stream seems to inherit ʻio.BufferedIOBase`, so this stream should be boiled.
+                reader = csv.reader(io.TextIOWrapper(data["audiencefile"], encoding='utf-8'))
+                csv_data = list(reader)  # read everything else into a list of rows
+
+                for line in csv_data:
+                    if len(line) != 4:
+                        msg["audiences"] = _("File wrongly formatted.")
+                        error["audiences"] = True
+                if "audiences" not in error or not error["audiences"]:
+                    stud_list, aud_li, oth_stu, u_info = self.get_user_lists(course)
                     self.database.audiences.delete_many({"courseid": course.get_id()})
-                    audiences = custom_yaml.load(data["audiencefile"].read())
-                else:
-                    audiences = json.loads(data["audiences"])
+                    for user_id, field, role, description in csv_data:
+                        user_id = user_id.strip()
+                        field = field.strip()
+                        role = role.strip()
+                        description = description.strip()
+                        query = {"courseid": course.get_id(), "description": description}
+                        if field != "username":
+                            user = self.database.users.find_one({field: user_id})
+                            user_id = user["username"] if user is not None else ""
+                            # TODO - Handle user not found
+                        update = {"$push": {"students": user_id}} if role == "student" else {
+                            "$push": {"tutors": user_id}}
 
-                for index, new_audience in enumerate(audiences):
-                    # In case of file upload, no id specified
-                    new_audience['_id'] = new_audience['_id'] if '_id' in new_audience else 'None'
-
-                    # Update the audience
-                    audience, errors = self.update_audience(course, new_audience['_id'], new_audience)
-
-                    # If file upload was done, get the default audience id
-                    audienceid = audience['_id']
-                    errored_students += errors
-
-                if len(errored_students) > 0:
-                    msg["audiences"] = _("Changes couldn't be applied for following students :") + "<ul>"
-                    for student in errored_students:
-                        msg["audiences"] += "<li>" + student + "</li>"
-                    msg["audiences"] += "</ul>"
-                    error["audiences"] = True
-                elif not error:
-                    msg["audiences"] = _("Audience updated.")
+                        self.database.audiences.update_one(query, update, upsert=True)
+                        self.database.audiences.update_one({"courseid": course.get_id(),
+                                                            "description": description,
+                                                            'students': {'$exists': False}}, {'$set': {'students': []}})
+                        self.database.audiences.update_one({"courseid": course.get_id(),
+                                                            "description": description,
+                                                            'tutors': {'$exists': False}}, {'$set': {'tutors': []}})
+                        if user_id not in stud_list and role == "student":
+                            self.user_manager.course_register_user(course, user_id)
                 active_tab = "tab_audiences"
-        except Exception:
+        except Exception as e:
+            print(e)
             msg["audiences"] = _('An error occurred while parsing the data.')
             error["audiences"] = True
             active_tab = "tab_audiences"
@@ -284,6 +293,7 @@ class CourseStudentListPage(INGIniousAdminPage):
                     msg["groups"] = _("Changes couldn't be applied for following students :") + "<ul>"
                     for student in errored_students:
                         msg["groups"] += "<li>" + student + "</li>"
+
                     msg["groups"] += "</ul>"
                     error["groups"] = True
                 elif not error:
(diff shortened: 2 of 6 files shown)


inginious/frontend/pages/course_admin/student_list.py
```diff
@@ -284,6 +292,35 @@ def post_groups(self, course, data, active_tab, msg, error):
...
+                    regex = r'\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Z|a-z]{2,}\b'
+                    if re.fullmatch(regex, usr):
+                        user = self.database.users.find_one({"email": usr})
+                        user is not None and inserted_content.append(user["username"])
+                    else:
+                        inserted_content.append(usr)
+                return inserted_content
+
+            for row in reader:
+                stud = row[1].split(';') if len(row) > 1 else []
```
anthonygego, line 313: CSV files can be seen as relational tables. I suppose here you consider the file format such as there is one line per audience. The additional value compared to a "database-dump" YAML document will be limited as this is rarely the way spreadsheets are produced. 

Administrators often start from the student list. Therefore, a handy format would be something like this:

    user_id | user_role | audience_description

where ``user_id`` is by default, the email address, which is a commonly available piece of information, ``user_role``, a numerical or textual enumeration defining if the user is a student or a tutor, and ``audience_description`` the description of the new audience, which is considered to be unique per audience at the import time (it would, by the way, actually be accurate to also make this unique in database).

A further improvement would be to allow user identification based on additional fields. For instance:

    user_id | id_field | user_role | audience_description
    
where id_field could be the DB field we want to match (``username``, ` `email``, ``bindings.auth_id/field``). A binding-based match would however require more control at the course level as this could violate GDPR, but the two first methods can be implemented.

inginious/frontend/pages/course_admin/student_list.py
```diff
@@ -284,6 +292,35 @@ def post_groups(self, course, data, active_tab, msg, error):
...
+            self.database.audiences.delete_many({"courseid": course.get_id()})
+            text_stream = io.TextIOWrapper(data["audiencecreationfile"], encoding='utf-8')
+            reader = csv.reader(text_stream)
+            stud_list, _, _, _ = self.get_user_lists(course)
+
+            def _list_checker(users):
+                inserted_content = []
+                for usr in users:
+                    regex = r'\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Z|a-z]{2,}\b'
+                    if re.fullmatch(regex, usr):
```
anthonygego, line 305: This won't be needed anymore if you ask the user to be explicit about this.

anthonygego (changes requested):

inginious/frontend/pages/course_admin/student_list.py
```diff
@@ -284,6 +292,37 @@ def post_groups(self, course, data, active_tab, msg, error):
...
+            if not error["udiences"]:
+                stud_list, _, _, _ = self.get_user_lists(course)
+                self.database.audiences.delete_many({"courseid": course.get_id()})
+                for user_id, field, role, description in reader:
+                    user_id = user_id.strip()
+                    field = field.strip()
+                    role = role.strip()
+                    description = description.strip()
+                    query = {"courseid": course.get_id(), "description": description}
+                    if field is not "username":
```
anthonygego, line 310: ``field`` should be validated before performing the request. Otherwise, any field can be matched, and Mongo operators can be used.

inginious/frontend/pages/course_admin/student_list.py
```diff
@@ -284,6 +292,37 @@ def post_groups(self, course, data, active_tab, msg, error):
...
+                        user = self.database.users.find_one({field: user_id})
+                        user_id = user["username"] if user is not None else ""
+                        #TODO - Handle user not found
+                    update = {"$push": {"students": user_id}} if role == "student" else {"$push": {"tutors": user_id}}
+
+                    self.database.audiences.update_one(query, update, upsert=True)
```
anthonygego, line 316: You're doing 3 requests per student here while you can compute the whole audiences before their insertions in one request per audience.

anthonygego (changes requested):

inginious/frontend/pages/course_admin/student_list.py
```diff
@@ -192,33 +198,48 @@ def post_audiences(self, course, data, active_tab, msg, error):
...
         try:
-            if "upload_audiences" in data or "audiences" in data:
-                errored_students = []
-                if "upload_audiences" in data:
+            if "audiencefile" in data and 'upload_audiences_creation' in data:
+                # get the Werkzeug datastructures.FileStorage object.
+                # The stream of this object is the stream body of the uploaded file.
+                # Furthermore, FileStorage.stream seems to inherit ʻio.BufferedIOBase`, so this stream should be boiled.
+                reader = csv.reader(io.TextIOWrapper(data["audiencefile"], encoding='utf-8'))
+                csv_data = list(reader)  # read everything else into a list of rows
```
anthonygego, line 206: This line is probably not needed as reader is a generator. If this is the case it also worsens performance to cast it into a list.

inginious/frontend/pages/course_admin/student_list.py
```diff
@@ -192,33 +198,48 @@ def post_audiences(self, course, data, active_tab, msg, error):
...
-                elif not error:
-                    msg["audiences"] = _("Audience updated.")
+                    for user_id, field, role, description in csv_data:
+                        user_id = user_id.strip()
+                        field = field.strip()
+                        role = role.strip()
+                        description = description.strip()
+                        if field not in ["username", "email", "realname"] + list(auth_method_ids) \
+                                or role not in ["student", "tutors"]:
+                            continue
```
anthonygego, line 222: Maybe we should warn the user of this. There is much chance they will never know the line wasn't parsed.

inginious/frontend/pages/course_admin/student_list.py
```diff
@@ -192,33 +198,48 @@ def post_audiences(self, course, data, active_tab, msg, error):
...
+                        if field != "username":
+                            user = self.database.users.find_one({field: user_id})
+                            user_id = user["username"] if user is not None else ""
+                            # TODO - Handle user not found
```
anthonygego, line 226: Same, user should be warned.

inginious/frontend/pages/course_admin/student_list.py
```diff
@@ -192,33 +198,48 @@ def post_audiences(self, course, data, active_tab, msg, error):
...
+                        query = {"courseid": course.get_id(), "description": description}
+
+                        existing_audience = self.database.audiences.find_one(query)
```
anthonygego, line 229: You shouldn't perform such a merge (and especially mongo queries) for each student in the file. Parse the file, create the audiences dictionaries and then only update the existing audiences if needed. This way you keep the amount of queries to the minimum (or at least, a reasonable amount).

inginious/frontend/pages/course_admin/student_list.py
```diff
@@ -192,33 +198,48 @@ def post_audiences(self, course, data, active_tab, msg, error):
...
+                                if role == "student" else \
+                                {"courseid": course.get_id(), "description": description,
+                                 "students": [], "tutors": [user_id]}
+                            self.database.audiences.insert_one(insert)
+                        else:
+                            update = {'$push': {'students': user_id}} \
+                                if role == "student" else {'$push': {'tutors': user_id}}
+                            self.database.audiences.update_one(query, update)
+                        if user_id not in stud_list and role == "student":
+                            self.user_manager.course_register_user(course, user_id)
```
anthonygego, line 242: Same remark, you can merge the student list with the list in database in one operation, here you will perform a query for each student.

inginious/frontend/templates/course_admin/student_list.html
```diff
@@ -78,7 +78,7 @@ <h2>{{ _("User management") }}</h2>
             <h2> {{ _("Audiences") }}
                     <div class="pull-right">
                         <div class="btn-group btn-group-sm">
-                            <button type="submit" id="switch" class="btn btn-danger" data-toggle="modal" data-target="#upload_modal_audiences"><i class="fa fa-upload"></i> <div class="d-none d-sm-inline">{{ _("Upload structure") }}</div></button>
+                            <button type="submit" id="create_audiences_f" class="btn btn-warning" data-toggle="modal" data-target="#upload_audiences_creation"><i class="fa fa-upload"></i> <div class="d-none d-sm-inline">{{ _("Upload audiences creation") }}</div></button>
```
anthonygego, line 81: audiences ~creation~

anthonygego (changes requested):

inginious/frontend/pages/course_admin/student_list.py
Drumor: reader is an iterator : https://docs.python.org/3/library/csv.html#csv.reader and as I have to iterate the data twice, it seems faster to cast into a list in place of using tee() for example (https://docs.python.org/3/library/itertools.html#itertools.tee).

inginious/frontend/pages/course_admin/student_list.py
```diff
@@ -192,35 +198,70 @@ def post_audiences(self, course, data, active_tab, msg, error):
...
+                            course_students.append(user_id)
+                        else:
+                            tutors_per_audience.setdefault(description, []).append(user_id)
+                            course_tutors.append(user_id)
+                    for key, value in students_per_audience.items():
+                        audiences.append({"description": key, "courseid": courseid,
+                                  "students": value, "tutors": tutors_per_audience[key] if key in tutors_per_audience else []})
+                    new_students = list(set(stud_list).union(set(course_students)))
+                    new_tutors = list(set(course.get_tutors()).union(set(course_tutors)))
+                    self.database.courses.update_one({"_id": "LFSAB1106"}, {"$set": {"students": new_students,
```
anthonygego, line 250: Hardcoded value here.

inginious/frontend/pages/course_admin/student_list.py
```diff
@@ -192,35 +198,70 @@ def post_audiences(self, course, data, active_tab, msg, error):
...
+                                                                                     "tutors": new_tutors}})
+                    for audience in audiences:
```
anthonygego, line 252: There should be a comment here to specify that this is done to avoid removing the audience id and impact the group audience filter. If anyone forgets about this, they may remove the code.

inginious/frontend/pages/course_admin/student_list.py
```diff
@@ -192,35 +198,72 @@ def post_audiences(self, course, data, active_tab, msg, error):
...
+                    courseid = course.get_id()
+                    self.database.audiences.delete_many({"courseid": courseid})
+                    for user_id, field, role, description in csv_data:
+                        user_id = user_id.strip()
+                        field = field.strip()
+                        role = role.strip()
+                        description = description.strip()
+                        if field not in ["username", "email", "realname"] + list(auth_method_ids) \
+                                or role not in ["student", "tutor"]:
+                            msg["audiences"] = _("Some fields were not recognized:" + field)
```
anthonygego, line 228: You need to extract the dynamic part from the gettext string.

inginious/frontend/pages/course_admin/student_list.py
```diff
@@ -192,35 +198,72 @@ def post_audiences(self, course, data, active_tab, msg, error):
...
+                            error["audiences"] = True
+                            continue
+                        if field != "username":
+                            user = self.database.users.find_one({field: user_id})
+                            if user is not None:
+                                user_id = user["username"]
+                            else:
+                                msg["audiences"] = _("Some users were not found: " + user_id)
```
anthonygego, line 236: You need to extract the dynamic part from the gettext string. Do not use plural form if your code only and always returns one user at a time.

inginious/frontend/pages/course_admin/student_list.py
```diff
@@ -192,35 +198,72 @@ def post_audiences(self, course, data, active_tab, msg, error):
...
+                        error["audiences"] = True
+                if "audiences" not in error or not error["audiences"]:
+                    stud_list, aud_li, oth_stu, u_info = self.get_user_lists(course)
+                    courseid = course.get_id()
+                    self.database.audiences.delete_many({"courseid": courseid})
+                    for user_id, field, role, description in csv_data:
+                        user_id = user_id.strip()
+                        field = field.strip()
+                        role = role.strip()
+                        description = description.strip()
```
anthonygego, line 225: If no description is provided, we probably should ignore the line.

anthonygego (changes requested):
Please provide some documentation on how to format the file.

inginious/frontend/pages/course_admin/student_list.py
```diff
@@ -192,35 +198,77 @@ def post_audiences(self, course, data, active_tab, msg, error):
...
+                        error["audiences"] = True
+                if "audiences" not in error or not error["audiences"]:
+                    stud_list, aud_li, oth_stu, u_info = self.get_user_lists(course)
+                    courseid = course.get_id()
+                    self.database.audiences.delete_many({"courseid": courseid})
+                    for user_id, field, role, description in csv_data:
+                        user_id = user_id.strip()
+                        field = field.strip()
+                        role = role.strip()
+                        print("LUDO")
```
anthonygego, line 225: Debug code here...

inginious/frontend/pages/course_admin/student_list.py
```diff
@@ -192,35 +198,77 @@ def post_audiences(self, course, data, active_tab, msg, error):
...
+                        print(description)
+                        if description != "":
+                            description = description.strip()
+                        else:
+                            print("nothing")
```
anthonygego, line 230: Debug code here

anthonygego (changes requested):
All the conversation are marked resolved but there are still debug prints here !!!

anthonygego (changes requested):
There is still no user documentation anywhere. Users won't read the code to know how to format the CSV file.

The last comments made in this discussion could really be avoided by making a self-review of your work.

doc/teacher_doc/course_admin.rst
```diff
@@ -135,35 +135,23 @@ The student list is entirely managed by drag-and-drop.
...
-                 - user1
-                 - user2
-    -    description: Audience 2
-         tutors:
-                 - tutor1
-                 - tutor2
-         students:
-                 - user3
-                 - user4
+    user_id | id_field | user_role | description
```
anthonygego, line 149: This description makes the user think ``|`` is the separator char, but it is probably not as the default CSV dialect is Excel. You need to specify the separator char and the quote char for this to be complete.

inginious/frontend/pages/course_admin/student_list.py
```diff
@@ -24,19 +28,19 @@ def GET_AUTH(self, courseid):  # pylint: disable=arguments-differ
...
         if "download_audiences" in flask.request.args:
             audiences = [{"description": audience["description"],
-                           "students": audience["students"],
-                           "tutors": audience["tutors"]} for audience in
-                          self.user_manager.get_course_audiences(course)]
+                          "students": audience["students"],
+                          "tutors": audience["tutors"]} for audience in
+                         self.user_manager.get_course_audiences(course)]
             response = Response(response=yaml.dump(audiences), content_type='text/x-yaml')
             response.headers['Content-Disposition'] = 'attachment; filename="audiences.yaml"'
```
anthonygego, line 35: There is actually no much interest in downloading the audiences in the YAML format while the upload format is CSV. Using the same file format could allow the user to simply add some people to the list without regenerating it manually. Either the ``id_field`` can be set to the email, either the user can be asked which field they prefer.

inginious/frontend/pages/course_admin/student_list.py
```diff
@@ -192,35 +198,92 @@ def post_audiences(self, course, data, active_tab, msg, error):
...
+                        if field not in ["username", "email", "realname"] + \
+                                list(self.user_manager.get_auth_methods().keys()):
+                            msg["audiences"] = _("Field was not recognized: ") + field
+                            error["audiences"] = True
+                            continue
+                        if role not in ["student", "tutor"]:
+                            msg["audiences"] = _("Unknown role: ") + role
+                            error["audiences"] = True
+                            continue
+                        if field != "username":
```
anthonygego, line 239: Why ? A non-existing username should not be accepted.

anthonygego (changes requested):

inginious/frontend/pages/course_admin/student_list.py
Drumor: It was to avoid unecessary request to database as I work with username after but u re right that a non-existing user is actually passing through.

inginious/frontend/pages/course_admin/student_list.py
```diff
@@ -23,20 +27,22 @@ def GET_AUTH(self, courseid):  # pylint: disable=arguments-differ
...
-                          self.user_manager.get_course_audiences(course)]
-            response = Response(response=yaml.dump(audiences), content_type='text/x-yaml')
-            response.headers['Content-Disposition'] = 'attachment; filename="audiences.yaml"'
+            audiences = []
+            for audience in self.user_manager.get_course_audiences(course):
+                for student in audience["students"]:
+                    audiences.append(student+",username,student,"+audience["description"])
+                for tutor in audience["tutors"]:
+                    audiences.append(tutor+",username,tutor,"+audience["description"])
+            response = Response(response=audiences, content_type='text/csv')
```
anthonygego, line 36: There is no new line between the entries here. You should better rely on the ``csv.writer`` class as it is probably done for the student table exports.

anthonygego (approved):

inginious/frontend/pages/course_admin/student_list.py
anthonygego: BTW, and once again actually:

> Either the id_field can be set to the email, either the user can be asked which field they prefer.

Users often complain usernames are not reliable to identify someone locally. This feature should be useful in practice.

anthonygego (approved):

anthonygego (changes requested):
See comments above.

inginious/frontend/pages/course_admin/student_list.py
```diff
@@ -25,15 +24,21 @@ class CourseStudentListPage(INGIniousAdminPage):
     def GET_AUTH(self, courseid):  # pylint: disable=arguments-differ
         """ GET request """
         course, __ = self.get_course_and_check_rights(courseid)
-
-        if "download_audiences" in flask.request.args:
+        if "preferred_field" in flask.request.args:
+            preferred_field = flask.request.args["preferred_field"]
```
anthonygego, line 28: We can access any database field here !!!

anthonygego (changes requested):

anthonygego:
Well... looks way better, but we actually can't perform those operations on the user's realname, as this is not a unique field. People can actually have the same name, and there is loss of information, so this feature has to be removed.

anthonygego (approved):

PR #786, opened 2022-02-17
Review responsive

Drumor:
This PR proposes some tiny update to make administration more responsive, especially on smartphone.
Most of tables overflow are now manage.

### DIFF
--- a/inginious/frontend/static/css/INGInious.css
+++ b/inginious/frontend/static/css/INGInious.css
@@ -255,6 +255,11 @@ body {
     transform: rotate(90deg);
   }
 }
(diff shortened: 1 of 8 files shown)


inginious/frontend/static/css/INGInious.css
```diff
@@ -255,6 +255,11 @@ body {
     transform: rotate(90deg);
   }
 }
+@media (max-width: 700px){
```
anthonygego, line 258: The ``.less`` file does not seem to be changed.

anthonygego (changes requested):

PR #796, opened 2022-03-09
add - support for common run

anaHue:


base-containers/base/bin/INGInious
```diff
@@ -20,7 +20,7 @@ import asyncio
 import struct
 import zmq
 import zmq.asyncio
-
+import traceback
```
anthonygego, line 23: This can be removed now

base-containers/base/bin/INGInious
```diff
@@ -367,10 +367,11 @@ class INGIniousMainRunner:
                 try:
                     user = "root" if self.run_as_root else "worker"
                     stdout, stderr = execute_process(run_final_cmd, user=user)
-                except:
+                except Exception as e:
                     inginious_container_api.feedback.set_global_result('crash')
                     inginious_container_api.feedback.set_global_feedback(
-                        "An error occured while running the grading script. It is possible that it is non-executable or made a timeout")
+			"An error occured while running the grading script. It is possible that it is non-executable or made a timeout " + str(traceback.format_exc()))
```
anthonygego, line 373: Same remark, line 280 should probably be the only change in this file.

base-containers/base/inginious_container_api/utils.py
```diff
@@ -27,8 +27,9 @@ def set_limits_user(user):
 
 
 def set_executable(filename):
-    st = os.stat(filename)
-    os.chmod(filename, st.st_mode | stat.S_IEXEC)
+    if "course/common" not in filename:
```
anthonygego, line 30: At this moment, we should add why in comment (because the folder is mounted readonly in the container). A _startswith_ check with the complete prefix, instead of a _contains_ check, would ensure this is only applied to the concerned files.

I would investigate, in a future patch, about moving the whole ``set_executable`` thing to the agent where the temporary folder is created. (This can also be mentioned in the comment to keep track)

configuration.yaml.save
```diff
@@ -0,0 +1,26 @@
+backend: local
```
anthonygego, line 1: This file should be removed from this patch.

inginious/common/filesystems/local.py
```diff
@@ -142,7 +142,10 @@ def _recursive_overwrite(self, src, dest):
                 self._recursive_overwrite(os.path.join(src, file),
                                           os.path.join(dest, file))
         else:
-            shutil.copyfile(src, dest, follow_symlinks=False)
+            dirname = os.path.dirname(dest)
```
anthonygego, line 145: A good way to import someone's changes is to use ``cherry-pick``. This avoids stealing the authorship of @nrybowski in this case :) I'll merge his PR first, so you can remove this from the PR.

anthonygego (changes requested):
Please squash your commits once the changes are done.

base-containers/base/bin/INGInious
```diff
@@ -370,7 +369,8 @@ class INGIniousMainRunner:
                 except:
                     inginious_container_api.feedback.set_global_result('crash')
                     inginious_container_api.feedback.set_global_feedback(
-                        "An error occured while running the grading script. It is possible that it is non-executable or made a timeout")
+			"An error occured while running the grading script. It is possible that it is non-executable or made a timeout")
```
anthonygego, line 372: There are tabs instead of spaces here

Issue #736, opened 2021-11-03
[Feature request] Denylist


AlexandreDubray:
It would be nice to forbid the access to a course for a given set of users using a denylist.

PR #803, opened 2022-03-11
[frontend/settings] Add deny list feature

nrybowski:
Propose an implementation for #736

### DIFF
--- a/inginious/frontend/courses.py
+++ b/inginious/frontend/courses.py
@@ -70,7 +70,7 @@ class Course(object):
             self._registration_ac = self._content.get('registration_ac', None)
             if self._registration_ac not in [None, "username", "binding", "email"]:
                 raise Exception("Course has an invalid value for registration_ac: " + self.get_id())
-            self._registration_ac_type = self._content.get('registration_ac_type', "accept")
+            self._registration_ac_accept = self._content.get('registration_ac_accept', True)
             self._registration_ac_list = self._content.get('registration_ac_list', [])
             self._groups_student_choice = self._content.get("groups_student_choice", False)
             self._allow_unregister = self._content.get('allow_unregister', True)
@@ -171,9 +171,9 @@ class Course(object):
         """ Returns either None, "username", "binding", or "email", depending on the method used to verify that users can register to the course """
         return self._registration_ac
 
-    def get_access_control_type(self):
-        """ Returns either "accept" or "deny", depending on the control type used to verify that users can register to the course """
-        return self._registration_ac_type
+    def get_access_control_accept(self):
+        """ Returns either True (accept) or False (deny), depending on the control type used to verify that users can register to the course """
+        return self._registration_ac_accept
 
     def get_access_control_list(self) -> List[str]:
         """ Returns the list of all users/emails/binding methods/... (see get_access_control_method) allowed by the AC list """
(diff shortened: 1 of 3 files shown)


inginious/frontend/templates/course_admin/settings.html
```diff
@@ -340,6 +367,8 @@ <h2>{{_("Course settings")}}</h2>
         updateLTI();
         $("#lti").click(updateLTI);
         $("#lti").change(updateLTI);
+        hideAcType();
+        $("[name='registration_ac']").change(hideAcType);
```
anthonygego, line 371: I think you can add an id to avoid a complex name-based selector.

inginious/frontend/courses.py
```diff
@@ -170,6 +171,10 @@ def get_access_control_method(self):
         """ Returns either None, "username", "binding", or "email", depending on the method used to verify that users can register to the course """
         return self._registration_ac
 
+    def get_access_control_type(self):
+        """ Returns either "accept" or "deny", depending on the control type used to verify that users can register to the course """
+        return self._registration_ac_type
```
anthonygego, line 176: As there are only 2 possible states (allow or deny), I'd rather have a boolean data type instead of a textual one. This would lower the risk of bugs and ease code reading.

anthonygego (changes requested):

inginious/frontend/templates/course_admin/settings.html
nrybowski: The id is defined only once here
https://github.com/UCL-INGI/INGInious/blob/2b9ea8d0e38f9e6750f4b365612fd56209bf7c11/inginious/frontend/templates/course_admin/settings.html#L262
and the next radios use the same name to reference this id.
Matching the id is not sufficient to detect a change in the radio value (it detects when the radio is clicked but not when another radio is selected).

inginious/frontend/courses.py
nrybowski: Fixed in 49ba9d4f6a511a7785f696823835e68da5184a92.

inginious/frontend/pages/course_admin/settings.py
```diff
@@ -76,6 +76,7 @@ def POST_AUTH(self, courseid):  # pylint: disable=arguments-differ
                 errors.append(_('Invalid ACL value'))
             if course_content['registration_ac'] == "None":
                 course_content['registration_ac'] = None
+            course_content['registration_ac_accept'] = data['registration_ac_accept']
```
anthonygego, line 79: You don't store a boolean here, but you load the value and use it as a boolean in the remaining of the code.

anthonygego (changes requested):

nrybowski:
Fix review and rebase on 15a948f60f6851f81d83b98414cf7f67ca2c4fef.

inginious/frontend/pages/course_admin/settings.py
nrybowski: Fixed in fb574ef83152d1534d3f41a08fbe21415547fcdc.

anthonygego:
LGTM

anthonygego (approved):

anthonygego:
Can you solve the conflict ? I guess it comes from the previous accepted PR :-)

PR #804, opened 2022-03-11
[frontend/settings] Remove useless spaces in ACL

### LINKED ISSUE #802: [frontend/settings] Access control list adds useless blank spaces on save
When we push on the "save changes" button of the course settings page, blank spaces are added at the beginning of the first line and at the end of the last line of the ACL.

Here are the interesting part of the course.yaml:
1. after the initial addition:
```
registration_ac_list:
- test
- test2
```
2. after pushing once on "save changes"
```
registration_ac_list:
- ' test'
- 'test2 '
```
3. pushing once again
```
registration_ac_list:
- '  test'
- 'test2  '
```
4. after adding a new entry
```
registration_ac_list:
- '   test'
- 'test2   '
- test3
```
5. after pushing only on "save changes"
```
registration_ac_list:
- '    test'
- 'test2   '
- 'test3 '
```

nrybowski:
Fix #802

### DIFF
--- a/inginious/frontend/pages/course_admin/settings.py
+++ b/inginious/frontend/pages/course_admin/settings.py
@@ -76,7 +76,7 @@ class CourseSettingsPage(INGIniousAdminPage):
                 errors.append(_('Invalid ACL value'))
             if course_content['registration_ac'] == "None":
                 course_content['registration_ac'] = None
-            course_content['registration_ac_list'] = data['registration_ac_list'].splitlines()
+            course_content['registration_ac_list'] = [line.strip() for line in data['registration_ac_list'].splitlines()]
 
             course_content['is_lti'] = 'lti' in data and data['lti'] == "true"
             course_content['lti_url'] = data.get("lti_url", "")
--- a/inginious/frontend/templates/course_admin/settings.html
+++ b/inginious/frontend/templates/course_admin/settings.html
@@ -291,7 +291,7 @@
     <div class="form-group row no-lti">
         <label for="registration_ac_list" class="col-sm-2 control-label">{{_("Access control list")}}</label>
         <div class="col-sm-10">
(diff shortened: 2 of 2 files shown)


anthonygego:
This is probably not enough to fix #802 as this does not ensure the specified items are trimmed before saving changes to file. It only ensures spaces are not added by the rendering process.

nrybowski:
This commit indeed fixes #802 but I agree that further input sanitation is required.
I will add that in subsequent commits.

nrybowski:
> This is probably not enough to fix #802 as this does not ensure the specified items are trimmed before saving changes to file. It only ensures spaces are not added by the rendering process.

Fixed in e79eddc47c9a9d52f4c85d5d66de6b2dfeab2696.

inginious/frontend/pages/course_admin/settings.py
```diff
@@ -76,7 +76,7 @@ def POST_AUTH(self, courseid):  # pylint: disable=arguments-differ
                 errors.append(_('Invalid ACL value'))
             if course_content['registration_ac'] == "None":
                 course_content['registration_ac'] = None
-            course_content['registration_ac_list'] = data['registration_ac_list'].splitlines()
+            course_content['registration_ac_list'] = [re.sub(' ', '', line) for line in data['registration_ac_list'].splitlines()]
```
anthonygego, line 79: ``line.strip()`` is probably more appropriate.

anthonygego (changes requested):

inginious/frontend/pages/course_admin/settings.py
nrybowski: Do we accept entries such as `foo bar` in the ACL?

nrybowski: I wasn't able to add tabs via the web interface but this is still possible via the yaml file.
I will replace the regex by `strip`.

nrybowski:
Rebased on 15a948f60f6851f81d83b98414cf7f67ca2c4fef.

anthonygego:
LGTM.

anthonygego (approved):

PR #807, opened 2022-03-24
[templates/edit_tabs] Fix radio selection condition

### LINKED ISSUE #806: The submission limits disappear when the task settings are changed
When you edit the basic settings of a task via the front end to change a setting, for example to open the task to students, the previously set submission limit disappears. In fact, none of the radio buttons associated to the submission limit are highlighted anymore.

If you are not careful when you save the settings, there are no longer any submission limits even though they were previously set and are still set in the `task.yaml`.

nrybowski:
Fix #806


anthonygego:
This PR/issue probably also means there are no checks on the data and that negative numbers are also accepted and lead to weird behaviours. Could you add some checks for that too ?

anthonygego:
LGTM

anthonygego (approved):

Issue #808, opened 2022-04-03
[Frontend] Allow RTL


yarons:
Localization will increase the adoption of this system and allow more diverse people to use it.

Thank you.

Drumor:
Hello @yarons,

Thank u for your interest for INGInious project. Actually, Localization already exist and works with weblate : https://weblate.info.ucl.ac.be/engage/inginious/?utm_source=widget

If i missunderstood your issue, could you please explain it. Thank you.

Drumor:
I see from #811 that you successfully add Hebrew. So no need extra explanation :) I think you need to sign CLA btw.

Concerning version we will have a look.

yarons:
Sure, there are small things that bother me, for example, in English the puctuation symbols are written without a space between them and the text, so I would write "For example:", in most strings there's a space before the colon: "For example :".

I can probably try and fix that in the source code but it should be the style for all new strings as well.

Where can I sign the CLA?

Drumor:
Feel certainly free to propose adaptation through a PR :)

You can sign CLA in the  #811 PR :)

yarons:
@Drumor Done :)

I'll see what I can do regarding the original strings, thank you.

Drumor:
I close this issue for now but do not hesitate to open it again if needed.

yarons:
That's fine but the English is still wrong, do you want me to open an additional issue about that?

Drumor:
Oh no,  making a PR with English correction is more than enough :) Thank you.

Drumor:
Thx to @yarons awesome work, we have now Hebrew as available translation.
It should be great to make RTL available on frontend to fix text alignments.

Based on our work, this should be done : 

1. Adding bootstraprtl (https://bootstrap.rtlcss.com/)
2. Rebuild bootwatch theme to fit RTL
3. Add feature to choose rtl or not based on selected language.

Question : Do we have to make modification like putting the menu on the right and so on? @yarons (or anyone else) if u have any advice about that, feel free to comment.

yarons:
UX is a very complicated world, for example, moving the menu to the right is quite intuitive yet the fingers remain at the same length so depending on how you're holding your phone this change might not be very comfortable to everyone, yet in a desktop environment it's pretty natural to just flip everything so assuming it's only desktop we're looking at there shouldn't be any issue.

PR #813, opened 2022-04-22
[Frontend] Adding Hebrew as available translation

Drumor:
related to #811 and issue #808


anthonygego:
OK. We should plan to add RTL support code now to fix text alignments.

PR #817, opened 2022-05-18
[Frontend/Feedback] set up a dynamic length for line limit in docutils

### LINKED ISSUE #776: Docutils line length limit breaks feedbacks
A line length limit is introduced in docutils 0.17 and defaults to 10 000. This causes a parsing error when using the ``rst-image`` command that simply produces a unique line with the base64-encoded image.

Drumor:
Fix #776


anthonygego:
LGTM

PR #825, opened 2022-06-20
[Frontend/Course Admin] Additional fields feature

Drumor:
a) Move tags page
b) add "additional fields" feature

### DIFF
--- a/inginious/frontend/courses.py
+++ b/inginious/frontend/courses.py
@@ -11,6 +11,7 @@ import re
 from typing import List
 from collections import OrderedDict
 
+from inginious.common.additional_field import AdditionalField
 from inginious.common.tags import Tag
 from inginious.frontend.accessible_time import AccessibleTime
 from inginious.frontend.parsable_text import ParsableText
@@ -80,6 +81,10 @@ class Course(object):
             self._lti_keys = self._content.get('lti_keys', {})
             self._lti_send_back_grade = self._content.get('lti_send_back_grade', False)
             self._tags = {key: Tag(key, tag_dict, self.gettext) for key, tag_dict in self._content.get("tags", {}).items()}
+            self._additional_fields = {key: AdditionalField(key,
+                                                            field_dict[key]["description"],
+                                                            field_dict[key]["field_type"])
+                                       for key, field_dict in self._content.get("fields", {}).items()}
             task_dispenser_class = task_dispensers.get(self._content.get('task_dispenser', 'toc'), TableOfContents)
             # Here we use a lambda to encourage the task dispenser to pass by the task_factory to fetch course tasks
             # to avoid them to be cached along with the course object. Passing the task factory as argument
@@ -237,6 +242,9 @@ class Course(object):
 
     def get_tags(self):
         return self._tags
+
+    def get_additional_fields(self):
+        return self._additional_fields
 
     def get_task_dispenser(self):
         """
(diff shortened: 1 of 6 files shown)


anthonygego:
Please separate a) and b) so that the review of b) can be done easier, as a) review should be quite straightforward.
Please also fix the remaining conflict.

inginious/frontend/pages/course_admin/settings.py
```diff
@@ -131,6 +134,28 @@ def define_tags(self, course, data, course_content):
...
+        if not isinstance(fields, dict):
+            # prepare_datas returned an error
+            return fields
+
+        # Repair fields
+        for key, field in fields.items():
+            field["type"] = int(field["type"])
+
+            if (field["id"] == "" and field["type"] != 2):
+                return _("Some fields were missing.")
```
anthonygego, line 149: This is dead code: the ``prepare_datas`` return line filters out any field that has no id, as per documentation, this is the way to remove them from the list (no delete button)

anthonygego (changes requested):

inginious/frontend/pages/course_admin/settings.py
```diff
@@ -106,6 +110,7 @@ def POST_AUTH(self, courseid):  # pylint: disable=arguments-differ
 
     def page(self, course, errors=None, saved=False):
         """ Get all data and display the page """
+        self.template_helper.add_to_template_globals("field_types", FieldTypes)
```
anthonygego, line 113: It is not necessary to put these in template globals. Also check that Jinja2 3.1, which is required to pass the enum, is available in a python3.6, or at least a python3.8 venv.

anthonygego (changes requested):

inginious/frontend/pages/course_admin/settings.py
anthonygego: Either the code has to be changed to support python3.6, either the docs has to be changed for python3.8. With the recent release of RHEL9 and its availability in RHEL8 it is acceptable to upgrade the requirements to python 3.8.

anthonygego (changes requested):
You are never parsing and loading the fields into the course object, neither here or in #832. This will therefore never check YAML validity.

This is done for tags here: https://github.com/UCL-INGI/INGInious/blob/c9f38af3a72e6fa98dcd01004e3075a335483933/inginious/frontend/courses.py#L82

Your are therefore using ``get_course_descriptor_content`` in #832 to access the data which should only be used to populate the dictionary before dumping to YAML file. The OOP principles are already too poorly respected in the code to worsen it.

inginious/frontend/courses.py
```diff
@@ -80,6 +81,10 @@ def __init__(self, courseid, content, course_fs, task_factory, plugin_manager, t
             self._lti_keys = self._content.get('lti_keys', {})
             self._lti_send_back_grade = self._content.get('lti_send_back_grade', False)
             self._tags = {key: Tag(key, tag_dict, self.gettext) for key, tag_dict in self._content.get("tags", {}).items()}
+            self._additional_fields = {key: AdditionalField(key,
+                                                            field_dict["description"],
+                                                            field_dict["type"])
```
anthonygego, line 86: You should check here (or in the `AdditionalField` constructor, as this is probably done for `Tag`) that the `field_dict["type"]` value is actually valid and raise an exception if this is not the case.

Currently those are stored in litteral form in the `course.yaml` file. Maybe it's more relevant to store the integer values instead to keep the code more modulable (it is easier to play with the enum numbers in a case of refactor than with the enum name itself)

anthonygego (approved):

PR #832, opened 2022-07-06
[Frontend/Course] Additional fields for student

Drumor:
Let students set custom values for a dedicated course.


inginious/frontend/pages/course_custom_values.py
```diff
@@ -0,0 +1,40 @@
...
+        custom_values_user = current_user["custom"] if "custom" in current_user else {}
+        return self.show_page(courseid, custom_values_user)
+
+    def POST_AUTH(self, courseid):
+        """ POST request """
+        user_input = flask.request.form
+        username = self.user_manager.session_username()
+        current_user = self.database.users.find_one(
+            {"username": username})
+        custom_values_user = current_user["custom"] if "custom" in current_user else {}
```
anthonygego, line 24: Use ``get("custom", {})``

inginious/frontend/pages/course_custom_values.py
```diff
@@ -0,0 +1,40 @@
...
+        for key in user_input:
+            custom_values_user[key] = user_input[key]
```
anthonygego, line 26: `dict1.update(dict2)`

inginious/frontend/pages/course_custom_values.py
```diff
@@ -0,0 +1,40 @@
...
+        removed_keys = list(set(custom_values_user.keys()) - set(user_input.keys()))
+        for key in removed_keys:
+            custom_values_user[key] = "off"
```
anthonygego, line 29: Why have those fields to be set to ``"off"`` ?

inginious/frontend/pages/course_custom_values.py
```diff
@@ -0,0 +1,40 @@
...
+        self.database.users.update_one({"username": username}, {"$set": {"custom": custom_values_user}})
```
anthonygego, line 30: All those fields are defined for the user/course context. You should add a level to this dictionary not to conflict with other courses...

inginious/frontend/pages/course_custom_values.py
```diff
@@ -0,0 +1,40 @@
...
+
+        return self.show_page(courseid, custom_values_user)
+
+    def show_page(self, courseid, custom_values_user):
+        course = self.course_factory.get_course(courseid)
+        course_content = self.course_factory.get_course_descriptor_content(courseid)
+        custom_fields = course_content["fields"] if "fields" in course_content else {}
```
anthonygego, line 37: Encapsulate this in `Course` class to avoid fetching YAML content here.

inginious/frontend/templates/custom_values.html
```diff
@@ -0,0 +1,90 @@
...
+        </ol>
+    </nav>
+{% endblock %}
+{% block content %}
+    <h2>{{_("Custom values")}}</h2>
+    <form method="post" action="">
+    {% for key in custom_fields %}
+        <div class="form-group row">
+            <label class="col-sm-2 " for="{{ key }}" class="control-label">{{ key }}</label>
+            {% if  custom_fields[key]["type"] == 2 %}
```
anthonygego, line 75: Use your enumeration here.

inginious/frontend/templates/custom_values.html
```diff
@@ -0,0 +1,90 @@
...
+                <label><input class="" type="checkbox" id="{{ key }}" name="{{ key }}" {% if custom_values_user[key]=="on" %} checked {% endif %}>{{ custom_fields[key]["description"] }}</label>
+            {% else %}
+                <input class="form-control col-sm-10 "
+                        {% if custom_fields[key]["type"] == 0 %}
```
anthonygego, line 79: Use your enumeration here.

anthonygego (changes requested):

inginious/frontend/pages/course_custom_values.py
```diff
@@ -0,0 +1,48 @@
+"""
+Module definition for CustomValuePage class
+-*- coding: utf-8 -*-
```
anthonygego, line 3: https://peps.python.org/pep-0263/ : first or second line of file.

inginious/frontend/pages/course_custom_values.py
```diff
@@ -0,0 +1,48 @@
...
+        return self.show_page(courseid, custom_values_user.get(courseid, {}))
+
+    def POST_AUTH(self, courseid):
+        """ POST request """
+        user_input = flask.request.form
+        username = self.user_manager.session_username()
+        current_user = self.database.users.find_one(
+            {"username": username})
+        custom_values_user = current_user.get("custom", {}).get(courseid, {})
+        custom_values_user.update(user_input)
```
anthonygego, line 35: You don't sanitize ``user_input`` here. You should be able to check that only the course-defined keys are submitted in the form.

inginious/frontend/pages/course_custom_values.py
```diff
@@ -0,0 +1,48 @@
...
+        removed_keys = list(set(custom_values_user.keys()) - set(user_input.keys()))
+        for key in removed_keys:
```
anthonygego, line 37: May the ``user_input`` be a subset of the database dictionary ? Can't the submitted form be considered as the whole new data ?

inginious/frontend/pages/course_custom_values.py
```diff
@@ -0,0 +1,48 @@
...
+            del custom_values_user[key]
+        self.database.users.update_one({"username": username}, {"$set": {"custom." + courseid: custom_values_user}})
```
anthonygego, line 39: There is no assertion ``courseid`` is an actual courseid here, as no course is loaded yet.

inginious/frontend/templates/custom_values.html
```diff
@@ -0,0 +1,90 @@
...
+            {% if course.allow_unregister() %}
+                <a href="#" data-toggle="modal" data-target="#unregister_modal" class="list-group-item list-group-item-action list-group-item-info">
+                    <i class="fa fa-user-times fa-fw"></i>&nbsp; {{ _("Unregister from this course") }}
+                </a>
+            {% endif %}
+            <a class="list-group-item list-group-item-action list-group-item-info" href="{{ get_homepath() }}/custom/{{ course.get_id() }}">
+                <i class="fa fa-cogs fa-fw"></i>&nbsp; {{ _("Custom fields") }}
+            </a>
+        </div>
+    {% elif not user_manager.session_logged_in() %}
```
anthonygego, line 41: This page is not meant to be accessible to anonymous users as it is a user settings page... This is probably dead code as the page class inherits from ``INGIniousAuthPage``.

inginious/frontend/templates/custom_values.html
```diff
@@ -0,0 +1,90 @@
...
+        {% include "signin_button.html" %}
+    {% elif course.is_registration_possible(user_info) %}
+        <h3>{{ _("Enroll in the course") }}</h3>
```
anthonygego, line 44: This page should only be accessible once the student is registered to the course as it manages personal settings for the course...

inginious/frontend/templates/custom_values.html
```diff
@@ -0,0 +1,90 @@
...
+        </ol>
+    </nav>
+{% endblock %}
+{% block content %}
+    <h2>{{_("Custom values")}}</h2>
+    <form method="post" action="">
+    {% for key in custom_fields %}
+        <div class="form-group row">
+            <label class="col-sm-2 " for="{{ key }}" class="control-label">{{ key }}</label>
+            {% if  fieldtypes[custom_fields[key].get_type()] == fieldtypes['BOOLEAN'] %}
```
anthonygego, line 75: I guess you actually want to compare if ``custom_fields[key].get_type() ==  'BOOLEAN' `` ?

inginious/frontend/templates/custom_values.html
```diff
@@ -0,0 +1,90 @@
...
+                <label><input class="" type="checkbox" id="{{ key }}" name="{{ key }}" {% if custom_values_user[key]=="on" %} checked {% endif %}>{{ custom_fields[key].get_description() }}</label>
+            {% else %}
+                <input id="{{ key }}" name="{{ key }}" placeholder="{{ custom_fields[key].get_description() }}"
+                        {% if fieldtypes[custom_fields[key].get_type()] == fieldtypes['INTEGER'] %}
```
anthonygego, line 79: Same comment here.

inginious/frontend/templates/custom_values.html
```diff
@@ -0,0 +1,90 @@
...
+    </nav>
+{% endblock %}
+{% block content %}
+    <h2>{{_("Custom values")}}</h2>
+    <form method="post" action="">
+    {% for key in custom_fields %}
+        <div class="form-group row">
+            <label class="col-sm-2 " for="{{ key }}" class="control-label">{{ key }}</label>
+            {% if  fieldtypes[custom_fields[key].get_type()] == fieldtypes['BOOLEAN'] %}
+                <label><input class="" type="checkbox" id="{{ key }}" name="{{ key }}" {% if custom_values_user[key]=="on" %} checked {% endif %}>{{ custom_fields[key].get_description() }}</label>
```
anthonygego, line 76: So a boolean may be expected to be ``"on"``. This single line betrays you: there is actually no data check neither in your code. It should be used for more than rendering, because you won't be able to render correctly if data are not the expected one.

inginious/common/tests/test_additional_field.py
```diff
@@ -0,0 +1,44 @@
...
+# more information about the licensing of this file.
+
+import pytest
+
+from inginious.common.additional_field import AdditionalField
+
+
+@pytest.fixture()
+def init():
+    af = AdditionalField("test", "a description", 1)
```
anthonygego, line 14: This is a little weird as it restricts testing to a very limited number of cases...while the goal of testing is actually to ensures the code works for a wide variety of cases.

anthonygego (changes requested):
I would rather use the term of "course user settings" that may probably be more talkative for both users and developers than a couple of very acontextual and theoritical terms as "additional field" and "custom values".

inginious/common/tests/test_additional_field.py
Drumor: You're right but this will be complete within the dedicated test branch.

inginious/frontend/pages/course_user_settings.py
```diff
@@ -0,0 +1,101 @@
...
+        """ POST request """
+        username = self.user_manager.session_username()
+        if not self._is_accessible(courseid, username):
+            return self.template_helper.render("course_unavailable.html")
+        course_user_settings = self._sanitize_content(flask.request.form, courseid)
+        if course_user_settings == {}:
+            return self.show_page(courseid, {"error": ""})
+        self.database.users.update_one({"username": username}, {"$set": {"course_settings." + courseid: course_user_settings}})
+
+        return self.show_page(courseid, course_user_settings)
```
anthonygego, line 39: Form validation is more implicit than explicit: no success message is shown and no error message indicating what field in invalid neither...

inginious/frontend/templates/course.html
```diff
@@ -68,6 +68,9 @@ <h3> {{_("Last tried exercises") }}</h3>
                     <i class="fa fa-user-times fa-fw"></i>&nbsp; {{ _("Unregister from this course") }}
                 </a>
             {% endif %}
+            <a class="list-group-item list-group-item-action list-group-item-info" href="{{ get_homepath() }}/user_settings/{{ course.get_id() }}">
+                <i class="fa fa-cogs fa-fw"></i>&nbsp; {{ _("Course Settings") }}
```
anthonygego, line 72: Currently as the administrator is still not a course student (we should change that in the future), the link is not shown to the admin. But it should, because if not accessible it won't allow the admin to design and tests the task relying on that feature. I'd rather put this link before the unregister one to avoid misclicks.

inginious/frontend/templates/course_user_settings.html
```diff
@@ -0,0 +1,77 @@
...
+            {% endif %}
+            <li class="breadcrumb-item"><a href="{{get_homepath() }}/course/{{ course.get_id() }}">{{ course.get_name(user_manager.session_language()) }} <span class="sr-only">{{ _("(current)") }}</span></a></li>
+            <li class="breadcrumb-item active"><a href="#"><i class="fa fa-cogs"></i> {{ _("Course settings") }}</a></li>
+        </ol>
+    </nav>
+{% endblock %}
+{% block content %}
+    {%  if "error" in course_user_settings %}
+        <div class="alert alert-danger" role="alert">
+            {{ _("Something went wrong with given data.") }}
```
anthonygego, line 54: This should be a little more detailed.

inginious/frontend/pages/course_user_settings.py
```diff
@@ -0,0 +1,101 @@
...
+    def _sanitize_content(self, data, courseid):
+        """
+            Sanitize received data
+            :param: data  - A dictionary that normally contains user settings for a course.
+            :param: courseid - Id of the course.
+            :return: A sanitized dict, an empty dict if something went wrong.
+        """
+        if not isinstance(data, dict):
+            return {}
+        copied_data = data.copy()
```
anthonygego, line 76: As you pass on each item in the loop below, maybe you can just initialize an empty dict instead. This way you won't have to pop the key from the dictionary in case of error.

anthonygego (changes requested):

inginious/frontend/pages/course_user_settings.py
```diff
@@ -0,0 +1,104 @@
...
+        for setting in data:
+            if setting not in add_fields:
+                # if setting is not expected in the course definition. There is no reason to treat it.
+                return "Unknown field: "+str(setting)
+            try:
+                # try to cast given value to be sure that we match expected type.
+                if add_fields[setting].get_type_name() == "STRING":
+                    casted_value= str(data[setting])
+                elif add_fields[setting].get_type_name() == "INTEGER":
+                    casted_value = int(data[setting])
```
anthonygego, line 98: It's probably useful to allow empty setting, or to define default values for each types.

Indeed, values are not set until student specify them. Therefore, you can send values to the grading environment that wouldn't be accepted by your form, leaving the course administrator to deal with invalid values.

inginious/frontend/pages/course_user_settings.py
```diff
@@ -0,0 +1,104 @@
...
+        course_user_settings = current_user.get("course_settings", {})
+        return self.show_page(courseid, course_user_settings.get(courseid, {}), None)
+
+    def POST_AUTH(self, courseid):
+        """ POST request """
+        username = self.user_manager.session_username()
+        if not self._is_accessible(courseid, username):
+            return self.template_helper.render("course_unavailable.html")
+        course_user_settings = self._sanitize_content(flask.request.form, courseid)
+        if type(course_user_settings)==str:
```
anthonygego, line 35: This is not very Pythonic as you perform a try/except in `_sanitize_content`  that returns a string in the except block, and then compare for the type here. You can simply put the try/except block here around the `_sanitize_content` call, so that the method has clean specs : returns the data, and throws an exception in case of errors.

inginious/frontend/pages/course_user_settings.py
```diff
@@ -0,0 +1,104 @@
...
+    def _is_accessible(self, courseid, username):
+        """
+            Verify if course is accessible and that courseid match an existing course.
+            :param: courseid - The id of the course.
+            :param: username - The username of the logged user.
+            :return: A boolean if the course is accessible for the given user.
+        """
+        try:
+            course = self.course_factory.get_course(courseid)
+            return self.user_manager.course_is_user_registered(course, username)
```
anthonygego, line 73:
```suggestion
            return self.user_manager.course_is_user_registered(course, username) and self.user_manager.course_is_open_to_user(course, lti=False)
```

You can be registered to a closed course.

inginious/frontend/templates/course_user_settings.html
```diff
@@ -0,0 +1,77 @@
...
+                        {% if course_user_setting_fields[key].get_type_name() == 'INTEGER' %}
+                            type="number"
+                        {% else %}
+                            type="text"
+                        {% endif %}
+                       class="form-control col-sm-10 " value="{{ course_user_settings[key] }}">
+            {% endif %}
+        </div>
+    {% endfor %}
+        <button type="submit" class="btn btn-primary">{{ _("Save data") }}</button>
```
anthonygego, line 75: End of page buttons styles are generally `btn-block`

inginious/frontend/templates/course_user_settings.html
```diff
@@ -0,0 +1,77 @@
...
+            {% if  course_user_setting_fields[key].get_type_name() == 'BOOLEAN' %}
+                <label><input class="" type="checkbox" id="{{ key }}" name="{{ key }}" {% if course_user_settings[key]%} checked {% endif %}>{{ course_user_setting_fields[key].get_description() }}</label>
+            {% else %}
+                <input id="{{ key }}" name="{{ key }}" placeholder="{{ course_user_setting_fields[key].get_description() }}"
+                        {% if course_user_setting_fields[key].get_type_name() == 'INTEGER' %}
+                            type="number"
+                        {% else %}
+                            type="text"
+                        {% endif %}
+                       class="form-control col-sm-10 " value="{{ course_user_settings[key] }}">
```
anthonygego, line 71: You should not apply `col-sm-10` with `form-control` but wrap the `<input/>` with a `</div/>`, otherwise the input field goes beyond the margins of its container

inginious/frontend/templates/course_user_settings.html
```diff
@@ -0,0 +1,77 @@
...
+            {{ _(feedback[1] ) }}
+        </div>
+    {% endif %}
+    <h2>{{_("Course settings")}}</h2>
+    <form method="post" action="">
+    {% for key in course_user_setting_fields %}
+        <div class="form-group row">
+            <label class="col-sm-2 " for="{{ key }}" class="control-label">{{ key }}</label>
+            {% if  course_user_setting_fields[key].get_type_name() == 'BOOLEAN' %}
+                <label><input class="" type="checkbox" id="{{ key }}" name="{{ key }}" {% if course_user_settings[key]%} checked {% endif %}>{{ course_user_setting_fields[key].get_description() }}</label>
```
anthonygego, line 63: No need to repeat description next to the checkbox then.

inginious/common/tests/test_course_user_settings.py
```diff
@@ -0,0 +1,44 @@
...
+    """Test for course user settings class"""
+
+    def test_course_user_settings_init(self, init):
+        af = init
+        assert af is not None
+        af = CourseUserSetting(0, "a description", 1)
+        try:
+            af = CourseUserSetting("fail", "a description", 99)
+        except Exception:
+            assert True
```
anthonygego, line 28: `assert True` will always pass. If the code fails to throw an exception, the test itself will pass. You may want to `assert False` right after the constructor instead.

anthonygego (changes requested):
The more you read, the more you see... This PR has still not reached the minimum quality level for acceptance. I mean, we aren't in the "nice-to-have" stage.

inginious/common/tests/test_course_user_settings.py
```diff
@@ -0,0 +1,47 @@
+# -*- coding: utf-8 -*-
```
anthonygego, line 1: @nrybowski can you give your opinion on that file ? it still confuses me. I'd rather see it removed from the PR so that we can finally move on.

inginious/common/course_user_setting.py
```diff
@@ -30,24 +30,38 @@ def __hash__(self):
...
-        :return: type of the additional field
+        :return: type of the setting
         """
         return self._type
+
+    def get_cast_type(self):
+        return FieldTypes(self._type).get_cast_type()
+
+    def render(self,value):
+        content = "<input id='" + self.get_id() + "' name='" + self.get_id() + "'"
```
anthonygego, line 59: There is absolutely no additional value compared to the previous code. How is this piece of code more modular than the previous one ? Actually there even is a value loss as HTML code is formatted in Python.

What was meant was to have an abstract rendering method and a specific implementation for each kind of UserSetting.

inginious/frontend/pages/course_user_settings.py
anthonygego: There is still an error returned in case you leave an integer field empty.

inginious/frontend/pages/course_user_settings.py
```diff
@@ -0,0 +1,104 @@
...
+        copied_data = {}
+        course = self.course_factory.get_course(courseid)
+        add_fields = course.get_course_user_settings()
+        for setting in data:
+            if setting not in add_fields:
+                # if setting is not expected in the course definition. There is no reason to treat it.
+                return "Unknown field: "+str(setting)
+            try:
+                # try to cast given value to be sure that we match expected type.
+                if add_fields[setting].get_type_name() == "STRING":
```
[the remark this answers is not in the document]
anthonygego: Still relevant.

anthonygego (commented):
There still remain issues in this PR...

inginious/common/field_types.py
```diff
@@ -16,3 +16,7 @@ class FieldTypes(Enum):
     INTEGER = 1
     STRING = 2
     BOOLEAN = 3
+
+    def get_cast_type(self):
```
anthonygego, line 20: You should define a default value for these types, so that we know what to send to the backend if there is no value set by the user yet.

Maybe it would be worth replacing the enum values by a namedtuple ``(id, class, default_value)`` and storing ``FieldType.XXXX.value.id`` in yaml, so that the code is less error-prone.

inginious/frontend/pages/course_user_settings.py
```diff
@@ -0,0 +1,105 @@
...
+        course = self.course_factory.get_course(courseid)
+        add_fields = course.get_course_user_settings()
+        for field in add_fields:
+            if field not in data:
+                # setup default value.
+                copied_data[field] = add_fields[field].get_cast_type()()
+            else:
+                try:
+                    value = data[field]
+                    if (value is None or value == "") and add_fields[field].get_type_name()=="integer":
```
anthonygego, line 99: To be removed thus (this was still not modular btw) and replaced by field default value

anthonygego (changes requested):

inginious/frontend/templates/course.html
```diff
@@ -63,6 +64,9 @@ <h3> {{_("Last tried exercises") }}</h3>
                     {{ _("Group management") }}
                 {% endif %}
             </a>
+            <a class="list-group-item list-group-item-action list-group-item-info" href="{{ get_homepath() }}/user_settings/{{ course.get_id() }}">
+                <i class="fa fa-cogs fa-fw"></i>&nbsp; {{ _("Course Settings") }}
```
anthonygego, line 68: The link should also be available to admins, at least for testing purposes.

inginious/frontend/templates/course_user_settings.html
```diff
@@ -0,0 +1,69 @@
...
+    </nav>
+{% endblock %}
+{% block content %}
+    {%  if feedback is not none %}
+        <div class="alert alert-{{ feedback[0] }}" role="alert">
+            {{ _(feedback[1] ) }}
+        </div>
+    {% endif %}
+    <h2>{{_("Course settings")}}</h2>
+    <form method="post" action="">
```
anthonygego, line 58: An ``alert-info`` block telling the user *there is no customizable option for this course* in case no settings in defined would be great. It would also be simpler than hiding the page and restrict access in this case.

anthonygego (changes requested):
Two small cosmetic changes and it will, at last, be good :slightly_smiling_face:

inginious/frontend/templates/course_user_settings.html
```diff
@@ -0,0 +1,73 @@
...
+{% endblock %}
+{% block content %}
+    {%  if feedback is not none %}
+        <div class="alert alert-{{ feedback[0] }}" role="alert">
+            {{ _(feedback[1] ) }}
+        </div>
+    {% endif %}
+    <h2>{{_("Course settings")}}</h2>
+    {% if course_user_setting_fields | length ==0 %}
+        <div class="alert alert-info" role="alert">{{ _("there is no customizable option for this course") }}</div>
```
anthonygego, line 59:
```suggestion
        <div class="alert alert-info" role="alert">{{ _("There is no customizable option for this course") }}</div>
```

inginious/common/course_user_setting.py
```diff
@@ -0,0 +1,71 @@
...
+                return ft.value.cast_class
+
+    def get_default_value(self):
+        for ft in FieldTypes:
+            if ft.value.id == self._type:
+                return ft.value.default_value
+
+    def render(self,template_helper,value):
+        if self.get_type_name() !="":
+            return template_helper.render("user_settings/"+self.get_type_name()+".html",input_id=self.get_id(),value=value)
```
anthonygego, line 70:
```suggestion
            return template_helper.render("user_settings/"+self.get_type_name()+".html",input_id=self.get_id(),value=value or self.get_default_value())
```

inginious/common/course_user_setting.py
```diff
@@ -0,0 +1,71 @@
+"""
```
anthonygego, line 1: This whole thing in a frontend layer that the backend should not be aware of. Please move those files from ``/common`` to ``/frontend/user_settings`` for instance, as you called it for the templates.

PR #836, opened 2022-07-07
[Frontend / Course Admin] Bad student count.

Drumor:
As user_data only contains students, it's unnecessary to subtract admins.

### DIFF
--- a/inginious/frontend/templates/course_admin/student_list.html
+++ b/inginious/frontend/templates/course_admin/student_list.html
@@ -54,7 +54,7 @@
         <div role="tabpanel" class="card-body tab-pane active" id="tab_students">
             {% set len_all_users = user_data|length %}
             {% set len_staff = course.get_staff()|length %}
-            <p>{{ _("Number of students") }} : {{ len_all_users - len_staff }}</p>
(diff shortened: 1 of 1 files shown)


inginious/frontend/templates/course_admin/student_list.html
```diff
@@ -54,7 +54,7 @@ <h2>{{ _("User management") }}</h2>
         <div role="tabpanel" class="card-body tab-pane active" id="tab_students">
             {% set len_all_users = user_data|length %}
             {% set len_staff = course.get_staff()|length %}
-            <p>{{ _("Number of students") }} : {{ len_all_users - len_staff }}</p>
+            <p>{{ _("Number of students") }} : {{ len_all_users }}</p>
```
anthonygego, line 57: Is this correct then ? The next line actually slices user_data by len_staff. Both the values should match...

Drumor: This is not correct. We notify an issue when user_data is empty (data from database) but course.get_staff() returned a list (from course descriptor file). This results in non sense value.

anthonygego:
LGTM

Issue #542, opened 2020-06-10
Tasks in error should be more visible


Drumor:
INGInious give a feedback for task that are not correctly configured. (Which is great)
But this feedback is at the bottom of the task list in the course admin. This is not really visible. (Which is bad).
A quick answer to this problem is to set a message at the top of the admin interface to explain that some tasks are in error and give a link to the section.

PR #837, opened 2022-07-08
fix issue #542

anaHue:
add a div with links to tasks with errors

### DIFF
--- a/inginious/frontend/templates/course_admin/task_list.html
+++ b/inginious/frontend/templates/course_admin/task_list.html
@@ -83,6 +83,11 @@
     </div>
 {% endif %}
 
+<div id="errorLogContainer" class="col-md-12">
+    <p id="errorLog" style="color: white; margin-top: auto; margin-bottom: auto"></p>
+</div>
+<br>
+
 {% if errors %}
     {% for error in errors %}
         <div class="alert alert-danger" role="alert">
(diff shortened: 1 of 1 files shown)


inginious/frontend/templates/course_admin/task_list.html
```diff
@@ -83,6 +83,11 @@ <h5 class="modal-title" id="taskDispenserModalLabel">{{_("Switch task dispenser"
     </div>
 {% endif %}
 
+<div id="errorLogContainer" class="col-md-12">
```
anthonygego, line 86: Please use bootstrap alert boxes here.

inginious/frontend/templates/course_admin/task_list.html
```diff
@@ -113,4 +118,48 @@ <h5 class="modal-title" id="taskDispenserModalLabel">{{_("Switch task dispenser"
         </button>
     </div>
 </div>
+
+<script>
+	const tasksWithErrors = document.getElementsByClassName("bg-danger")
```
anthonygego, line 123: This whole piece of code is vanilla javascript. I'd recommend you to make it more jQuery-idiomatic to remain consistent with other parts of the code and ensure maximum compatibility with supported jQuery browsers/devices.

inginious/frontend/templates/course_admin/task_list.html
```diff
@@ -113,4 +118,48 @@ <h5 class="modal-title" id="taskDispenserModalLabel">{{_("Switch task dispenser"
...
+
+	for(let task of tasksWithErrors){
+		document.getElementById("errorLogContainer").innerHTML += "<br>&nbsp &nbsp &nbsp<a href='#" + task.id + "' style='color: white; text-decoration: underline'>" + task.innerText + "</a><br>"
```
anthonygego, line 126: Avoid writing HTML code by hand in code, jQuery has the adequate models to build HTML objects.

inginious/frontend/templates/course_admin/task_list.html
```diff
@@ -113,4 +118,48 @@ <h5 class="modal-title" id="taskDispenserModalLabel">{{_("Switch task dispenser"
...
+	}
+
+	console.log(tasksWithErrors.length)
+
+	if(tasksWithErrors.length == 1){
+		document.getElementById("errorLog").innerText = "There is a task with some errors";
```
anthonygego, line 132: Messages need to be localized. If you need these messages in JS, please export them from the jinja template using gettext. I'm honestly quite not sure this has to be done in javascript actually, as it is not really dynamic...

inginious/frontend/templates/course_admin/task_list.html
```diff
@@ -113,4 +118,48 @@ <h5 class="modal-title" id="taskDispenserModalLabel">{{_("Switch task dispenser"
...
+	} else {
+		document.getElementById("errorLogContainer").classList.add("d-none");
+	}
+
+	const list_delete_button = document.getElementsByClassName("delete_task")
+	
+	for(let del_button of list_delete_button){
+		const task_id = del_button.parentElement.parentElement.id
+		
+		del_button.addEventListener("click", () => {
```
anthonygego, line 148: You have to be very careful here. Deleting tasks from the list is the responsibility of each task dispensers, and button class can differ from implementations. I think this piece of code can be added in the ``task_dispenser_delete_task`` function which has to be called by each task dispenser when a task in deleted.

anthonygego (changes requested):

PR #840, opened 2022-07-11
fix #390

### LINKED ISSUE #390: Local FSProvider copy_from does not copy metadata of files
https://github.com/UCL-INGI/INGInious/blob/8415b6a1181a55fd30123c7d72bfd9488bdc798e/inginious/common/filesystems/local.py#L146

maybe we should use copy2 instead of copyfile.

anaHue:
use of copy2 instead of copy


anthonygego:
LGTM.

PR #841, opened 2022-07-13
[Frontend/Course Admin] move tags page to settings page

Drumor:


anthonygego:
There should be some HTML broken here as the footer text get pushed into the bootstrap card, at least on Chromium. You should also rather move the apply settings button outside of the card.

inginious/frontend/pages/course_admin/settings.py
```diff
@@ -103,3 +107,41 @@ def POST_AUTH(self, courseid):  # pylint: disable=arguments-differ
...
+    def define_tags(self, course, data, course_content):
+        tags = self.prepare_datas(data, "tags")
+        if type(tags) is not dict:
+            # prepare_datas returned an error
+            return tags
+
+        # Repair tags
+        for key, tag in tags.items():
+            # Since unchecked checkboxes are not present here, we manually add them to avoid later errors
+            print(tag)
```
anthonygego, line 120: Debug statement here...

anthonygego:
LGTM

PR #846, opened 2022-07-15
fix #211

### LINKED ISSUE #211: An indication to the user whether a subproblem is mandatory or not.
In case there are in one problem a mix of mandatory and optional code subproblems, there is no indication which one are from which kind.

anaHue:


anaHue:
The codacy review shows an error but the line indicated by the error message does not have the same content as the one in the file.

anthonygego:
It's true. You should be careful as the ``_optional`` attribute is only defined in ``CodeProblem`` and not any wider. This won't work for any other problems as long as a default ``_optional`` attribute is not set for them.

I'm closing this for now. Feel free to re-open it if you have time to work on it.

PR #847, opened 2022-07-15
fix #542

### LINKED ISSUE #542: Tasks in error should be more visible

anaHue:


inginious/frontend/templates/course_admin/task_list.html
```diff
@@ -113,4 +116,16 @@ <h5 class="modal-title" id="taskDispenserModalLabel">{{_("Switch task dispenser"
         </button>
     </div>
 </div>
+
+<script>
+    const tasksWithErrors = document.getElementsByClassName("bg-danger")
+
+    for(let task of tasksWithErrors){
```
anthonygego, line 123: This still makes no sense to do this with javascript code. This can be done in the Jinja template using bootstrap alerts and relocated on top of the page.

anthonygego:
I'm closing this for now. Feel free to reopen it when you have time to work on it.

PR #849, opened 2022-07-20
Moving the concept of weight into the Task Dispenser

maleclercq:


inginious/frontend/pages/course.py
```diff
@@ -88,6 +88,8 @@ def show_page(self, course):
 
             course_grade = round(tasks_score[0]/tasks_score[1]) if tasks_score[1] > 0 else 0
```
anthonygego, line 89: I guess this is now good to be removed...

inginious/frontend/static/js/task_dispensers.js
```diff
@@ -382,6 +388,32 @@ function dispenser_util_get_section_config(element) {
...
 }
 
+function dispenser_util_get_weights() {
+    const weight_list = {};
+    $(".weight").each(function(index){
+        weight = parseFloat(this.value)
+        if(weight > 1){
+            weight_list[this.id] = weight
+        }
+        else if(isNaN(weight) && this.value !== ""){
```
anthonygego, line 398: Those error messages should better be generated from the Python code, as it must be ensured that anarbitrary POSTed  form will get the same treatment and not inject bad values in database/files. I guess these checks can be moved to the ``check_dispenser_data`` instead.

Furthermore, all positive values can be (and were) tolerated, so values between 0 and 1 should be accepted. I guess one may have the practice to use several weight values that sum to 1.

inginious/frontend/task_dispensers/toc.py
```diff
@@ -26,6 +27,34 @@ def get_name(cls, language):
         """ Returns the localized task dispenser name """
         return _("Table of contents")
 
+    def get_course_grade(self, username, course, user_task_list):
+        tasks = course.get_tasks()
```
anthonygego, line 31: Here you should remove the dependency to course and use the internal ``task_list_func()`` call. You can take the ``get_user_task_list`` implementation as an example, and even, as this would be way cleaner, use that internal method to avoid passing the list, that, in fine, is generated by the task dispenser, as argument.

This will better respect the encapsulation principle. 

This remark is also valid for ``combinatory_test`` code.

inginious/frontend/task_dispensers/__init__.py
```diff
@@ -11,6 +12,16 @@ def __init__(self, task_list_func, dispenser_data, database, course_id):
         """
         pass
 
+    @abstractmethod
+    def get_weight(self, taskid):
```
anthonygego, line 16: I'm not really quite sure this makes sense here, as there may be task dispensers that do not use tasks weights, as yours (adaptative testing). ``get_weight`` should not be part of the interface, and the concept should be encapsulated in the task dispenser.

inginious/frontend/task_factory.py
```diff
@@ -241,6 +241,10 @@ def _get_last_updates(self, course, taskid, task_fs, need_content=False):
         if need_content:
             try:
                 task_content = descriptor_reader.load(task_fs.get(descriptor_name))
+                dispenser_data = course.get_descriptor()["dispenser_data"]
```
anthonygego, line 244: The task should not be aware of the weight anymore. You can break any form of retrocompatibility here as ensuring it will most probably lead to a mess in the future. A migration script should be provided (in the future) to automatically move the task weights from task YAMLs to course YAMLs.

inginious/frontend/tasks.py
```diff
@@ -110,7 +110,11 @@ def __init__(self, course, taskid, content, filesystem, plugin_manager, task_pro
         self._evaluate = self._data.get("evaluate", "best")
 
         # Grade weight
-        self._weight = float(self._data.get("weight", 1.0))
```
anthonygego, line 113: Same remark here: the weight should not be a task concept anymore.

inginious/frontend/templates/course_admin/edit_tabs/basic.html
```diff
@@ -62,12 +62,14 @@
                value="{{ task_data.get('categories', []) | join(',') }}">
     </div>
 </div>
+<!--
```
anthonygego, line 65: This part of the code can be deleted.

inginious/frontend/templates/course_admin/task_dispensers/task_buttons.html
```diff
@@ -2,6 +2,9 @@
 {# more information about the licensing of this file. #}
 
 <div class="ml-auto btn-group btn-group-sm" role="group">
+    <button class="btn btn-primary" data-toggle="modal" data-target="#weightModal_{{taskid}}">
+        Weight
```
anthonygego, line 6: You can use a gear icon or something similar as well as a tooltip to mimick the other buttons and to indicate additional settings.

anthonygego (changes requested):
You got the global idea but there is some changes to make. 

Please also remove the dead code: the copy/pasted code in ``task_dispenser.get_course_grade`` should be removed from where it was before: probably the ``user_manager`` ? 

There are chances this will break at some places because the course grade is not only computed in the student view (` pages/course.py``) but also in the admin view (student_list). You can also check if this is not also the case for the API code.

anthonygego (changes requested):
There still should be some dead code in the admin part (student list where grades are shown, and maybe in the ``/api`` folder).

anthonygego:
LGTM

PR #852, opened 2022-07-27
combinatory_test now return a proper dispenser data

maleclercq:
Prevent from a crash that could occur on: https://github.com/UCL-INGI/INGInious/blob/fc5bde432e8571f247538eae6bbecc493560a16a/inginious/frontend/pages/course_admin/task_edit.py#L109


anthonygego:
LGTM. Temporary mitigation for #853

Issue #743, opened 2021-11-16
Stopping submission wipes the submission state


anthonygego:
How to reproduce:
* Perform some submissions that produce a new submission state
* Launch a new submission and cancel it immediately after
* The submission state is set to an empty string

Probably concerns v0.7 too.

PR #857, opened 2022-08-10
[backend] do not erase previous state in case of errors

anthonygego:
The fix for #743  is done at the agent level, as it is not easy to know if the previous state has to be used at the queue level based on the received information as the state can legitimately be empty.


Drumor (approved):
Make some tests and looks good to me.

Issue #864, opened 2022-09-22
[frontend/plugin_manager] Problem type auto-discovery for plugins and base code


nrybowski:
The current way of loading INGInious plugins requires a call to the `load` method of the `PluginManager` https://github.com/UCL-INGI/INGInious/blob/fac7761f9f1a9992bb83f4ad81ec085a7e3b289f/inginious/frontend/plugin_manager.py#L65 

e.g. https://github.com/UCL-INGI/INGInious/blob/f28ad8daa77f0a724e85180eb4c8dcd61e792e7b/inginious/frontend/app.py#L315

This method then calls the `init`function of the plugin which, in turn, may add pages and hooks to the webapp, e.g. https://github.com/cnp3/INGInious-problems-network-trace/blob/d1c3f0229e06e777def1b87ab1671f71d23615a1/inginious-problems-network-trace/__init__.py#L260. It thus involves the usage of a deployed webapp.

In some use-cases, e.g. https://github.com/UCL-INGI/INGInious/pull/861, one may need to only load the new type(s) of problem(s) proposed by a given plugin in a headless configuration. Plugins should thus expose a method to that end.

In the same idea, INGInious does not propose auto-discovery functions for problem types, hence manual loading is still required, e.g. https://github.com/UCL-INGI/INGInious/blob/f28ad8daa77f0a724e85180eb4c8dcd61e792e7b/inginious/frontend/app.py#L217. A function solving this issue should also be available.

@anthonygego any inputs on this?

PR #861, opened 2022-09-08
[utils/task-tester] Add task tester

nrybowski:
Add the task tester feature. It allows replaying old submissions to ensure that tasks are still valid / functional e.g. after a modification of a task or the update of its environment.

This PR also contains a submission anonymizer. This is a script removing all the fields allowing the identification of the original submission author.

See https://github.com/UCL-INGI/.github/pull/1 for automated deployement in GitHub actions.

#### Task tester
- [x] Replay submissions 
  - [x] From a single task
  - [x] From a subset of the course's tasks
  - [x] From all the course's tasks
- [x] Load additional task types from plugins
- [x] Improve logs
- [ ] Compare stdout and stderr outputs and propose improvements to the task developers without making the tests fail
- [ ] Add thread pool to parallelize tasks and jobs rerun

#### Submissions anonymizer
- [x] Anonymize a set of submission files from a .tar.gz archive
- [x] Put the anonymized submissions directly in the course directory
- [x] Add tests
- [ ] Fix paths for windows
- [ ] Handle case of single submission in the archive

PR #866, opened 2022-09-28
Problem types auto-discovery

nrybowski:
This PR adds the feature described in #864.

### TODO
- [x] Base function to get tupple of problem name and problem class
- [x] Generic function to get all default problem types available
- [x] Get problem types added by a plugin
- [x] Handle displayable problem types

### DIFF
--- a/inginious-autotest
+++ b/inginious-autotest
@@ -24,7 +24,7 @@ from inginious.client.client_sync import ClientSync
 from inginious.frontend.arch_helper import start_asyncio_and_zmq, create_arch
 from yaml import load
 from inginious.common.filesystems.local import LocalFSProvider
-from inginious.frontend.task_problems import *
+from inginious.frontend.task_problems import get_default_displayable_problem_types
 
 
 def import_class(name):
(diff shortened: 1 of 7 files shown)


anthonygego (commented):
The code looks good to me, but I'm not sure to get the real advantage of inspecting the two base modules as the hardcoded list can be returned by the new ``get_default_XXX_types`` functions (the list will remain small).

How are plugins supposed to get advantage of these changes then ?

nrybowski:
Lines https://github.com/UCL-INGI/INGInious/pull/861/files/0cbb49dadaa831dc57d0d6ba982bb300009efd39#diff-1d137a6d9bae298e93e0d36cb3b8e2ef0cc297051516f2c5bd5ea479d9aa257bR239-R246 become something like
```
all_pbl_types = get_default_problem_type()
for plugin in plugins:
    pbl_types = get_problems_type(plugin)
    all_pbl_type.update(pbl_types)
```
and the plugin is immediately loaded.

anthonygego:
OK. Maybe we should take the opportunity to update the code so that the frontend and agents load the problem types this way and make the current ``add_problem_type`` function a dummy function ? This would be more coherent. What do you think ?

nrybowski:
Rebased on 0959be6d8127e5a867a3e190423897069bf2eae9.

The PR is ready for merge.

anthonygego:
Did we conclude anything from my previous comment ? If you feel the same way, do you plan to do this in another PR and let me merge this one in the meanwhile ?

nrybowski:
It seems I lost the commit implementing this during my rebase... I should be able to get it back from a local branch.

nrybowski:
Here is the missing commit 316841981eb03b197f85107be53b9135d0a2ba71. I pushed from the wrong branch yesterday ..

anthonygego:
LGTM.

PR #877, opened 2022-10-06
[Frontend/installer] review container build feature

### LINKED ISSUE #833: [inginious-installer] Fails to install the latest pyjavacpp grading container
The installer throws the following error when one try to install the `pyjavacpp` container:
```
[...]
Indicate the name of a container to build, or press enter to continue []: pyjavacpp
(INFO) Ok, I'll build container pyjavacpp
(INFO) Container pyjavacpp requires container ingi/inginious-c-cpp, I'll build it too.
(ERROR) An error occurred while copying the directory: [Errno 2] No such file or directory: '/tmp/tmp8t9s8wht/grading/ingi/inginious-c-cpp/Dockerfile'
```

The `ingi/inginious-c-cpp` grading container has been previously installed with the same script and is available in docker:
```
$ docker images
REPOSITORY                 TAG       IMAGE ID       CREATED        SIZE
ingi/inginious-c-cpp       latest    dcf99f5b6ca3   3 hours ago    1.71GB
<none>                     <none>    6535507ccb5c   3 hours ago    685MB
ingi/inginious-c-default   latest    ef0f7226a128   3 hours ago    685MB
ingi/inginious-c-base      latest    189d0cdc489b   3 hours ago    685MB
rockylinux                 8         c830f8e8f82b   3 months ago   205MB
```
It might also be interesting to first look if the dependency is available in docker before trying to rebuild it from scratch.

### LINKED ISSUE #871: [frontend/installer] prefixes mess when building containers with dependencies
**Describe the bug**
When building the containers using the installer scripts, containers with dependencies that are not base will fail to build, because their parent container is added to the list with a prefix.

Container without prefix is added to a ``todo`` dictionary when asked to the user, and with prefix when parsed automatically. However, the logic then check the todo lists using both nomenclatures.

https://github.com/UCL-INGI/INGInious/blob/64d71e3a9339a538b29b535f472d1fe301897e78/inginious/frontend/installer.py#L465

The whole code needs to be cleaned.

- Version: 0.7+

**To Reproduce**
1. Launch inginious-install up to container build and answer yes
2. Select a container with dependencies (for install pyjavacpp)
3. It will fail to build searching for non-existing folder

**Expected behavior**
Parsing works and builds the appropriate container if necessary

Drumor:
Fix #871
Fix #833

### DIFF
--- a/inginious/frontend/installer.py
+++ b/inginious/frontend/installer.py
@@ -12,6 +12,7 @@ import tempfile
 import re
 import urllib.request
 from binascii import hexlify
+from collections import OrderedDict
 import docker
 from docker.errors import BuildError
 from gridfs import GridFS
@@ -393,8 +394,8 @@ class Installer:
         self._display_info("done.".format(name))
 
     def select_containers_to_build(self):
-        #If on a dev branch, download from github master branch (then manually rebuild if needed)
-        #If on an pip installed version, download with the correct tag
+        # If on a dev branch, download from github master branch (then manually rebuild if needed)
+        # If on an pip installed version, download with the correct tag
         if not self._ask_boolean("Build the default containers? This is highly recommended, and is required to build other containers.", True):
             self._display_info("Skipping container building.")
             return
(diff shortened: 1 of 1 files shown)


anthonygego (changes requested):
There is still two nomenclatures used in the same data structures here. This is still pretty messy, useless checks are done, and there are high chances the code does not work with multiple levels of inheritence due to that.

Please select one nomenclature and stick to it.

inginious/frontend/installer.py
```diff
@@ -477,29 +485,22 @@ def add_container(container):
...
-                    for container in todo_now:
-                        try:
-                            self._build_container("ingi/inginious-c-{}".format(container),
-                                                  os.path.join(tmpdirname, 'grading', container))
-                        except BuildError:
-                            self._display_error(
-                                "An error occured while building the container. Please retry manually.")
+                        __add_container(answer)
+
+                done = ["ingi/inginious-c-base", "ingi/inginious-c-default"]
```
anthonygego, line 490: This list does not seem to be used used anymore.

anthonygego (changes requested):

anthonygego:
LGTM

PR #902, opened 2022-12-07
File extensions must be separated by commas

tbarbette:
Apparently space do not work, extensions must be separated by commas (which is more usual. Am I the only one to follow the doc? :p)


Drumor:
LGTM. Thx!

PR #906, opened 2022-12-21
Update installation.rst

petitpouletrose:
the documentation targets Ubuntu 22.04 but the installation of Mongo refers to ubuntu focal (20.04).


Drumor:
LGTM

PR #910, opened 2023-02-10
[Frontend] Make RTL available

Drumor:
Based on #808 and as we have a fully translation of Hebrew (thx to @yarons ), this PR proposes a dedicated theme adaptation to RTL languages.

This PR :
1. Add bootstrap rtl for full support of RTL in Bootstrap 4. (https://bootstrap.rtlcss.com/)
2. Adapt theme based on bootswatch-rtl (https://github.com/mrjelveh/bootswatch-rtl)
3. Propose a tiny system to automatically define if the language is RTL or not by defining a specific msgid **__is_rtl** in each support language. Default behaviour remains LTR.


inginious/frontend/templates/layout.html
```diff
@@ -15,14 +15,22 @@
     <title>{% block title %}{% endblock %} | INGInious</title>
 
     <!-- Bootstrap -->
-    <link href="{{get_homepath(True)}}/static/css/bootstrap.min.css" rel="stylesheet">
+    <!-- CSS -->
+    {% if  _("__is_rtl") == "true"%}
+        <link href="{{get_homepath(True)}}/static/css/bootstrap-rtl.min.css" rel="stylesheet">
+        <!-- INGInious Style -->
+        <link href="{{get_homepath(True)}}/static/css/INGInious-rtl.css" rel="stylesheet">
```
anthonygego, line 22: It's maybe worth adding the same version suffix to ease future updates.

utils/minify/Gruntfile-rtl.js
```diff
@@ -0,0 +1,57 @@
+module.exports = function(grunt)
```
anthonygego, line 1: Can't we just have one Gruntfile to generate both the minified files to avoid repetitions ? Looks like the `concat.dist.files` and `uglify.dist.files` options take are dictionaries that would allow specifying both the output files in the same Gruntfile.

anthonygego (changes requested):
Can you split this PR into several commits:

1. Imported files (prebuilt bootstrap)
1. Specific style files + gruntfile for INGInous
1. RTL activation logic
1. Localization file updates

A +1900, -1200 commit including imported and automatically updated code is hard to read in case of issue or later consulting to recap the way a feature was implemented.

utils/minify/Gruntfile-rtl.js
anthonygego: Any news on this ?

inginious/frontend/static/css/bootstrap-rtl.min.css
```diff
@@ -0,0 +1,14 @@
+/*!
+ * Bootswatch v4.5.3
```
anthonygego, line 2: It's probably useful to make both the Bootstrap versions (LTR and RTL) match to minimize unwanted behaviours and benefit from the same feature set. The other one is still v4.1. Both the versions do not, for instance, use the exact same color palette in tabulars.

inginious/frontend/static/css/INGInious.less
```diff
@@ -75,6 +78,7 @@
             color: #333;
             padding-left: 1rem;
             font-size: 1.1em;
+            margin-right: @breadcrumb-a-right-m;
```
anthonygego, line 81: This does not provide the same result when there are several items in the breadcrumbs.
What you should actually do (this is intuitive) is to replace the `padding-left: 1rem`s of both `<span/>` and `<a/>` by `padding-right: 1rem`

anthonygego (changes requested):

anthonygego:
LGTM

PR #915, opened 2023-03-02
[agent_docker] add support for nvidia runtime

anthonygego:
This PRs adds support for the NVIDIA runtime to leverage GPU usage. This implementation as a seperate runtime was prefered over a GPU option in the environment config for the following reasons : 
- ``--gpus``, or its docker-py equivalent 

         device_requests=[docker.types.DeviceRequest(count=-1, capabilities=[['gpu']])]
   automatically selects the nvidia runtime as it provides the capability but will fails with a `500` error in the Docker API if the runtime is not available, exception which is hard to catch using docker-py.
- There is no docker-py API (or even Docker API ?) to know the capabilities of a runtime to prevent such an error.
- The GPU opt-in would have to be shown only for compatible runtimes (Kata, for instance, would not support it)
- It is actually really more lightweight to implement and not less a generic implementation as only NVIDIA GPUs are currently supported by the Docker ``--gpus`` argument.

An additional label is introduced to restrict usage of a Docker image to the nvidia environment type only.

As mentioned in the provided documentation, the `nvidia/cuda:x.y.z-base-rockylinux8` works in replacement of `rockylinux:8` as base for the `ingi/inginious-c-base` container. This was not included in this PR yet as it would require some practical usage tests to evaluate how accurate it would be.

### DIFF
--- a/doc/admin_doc/install_doc/create_container.rst
+++ b/doc/admin_doc/install_doc/create_container.rst
@@ -28,7 +28,7 @@ Here is an example of Dockerfile:
    RUN echo 'run_types["php"] = "/bin/php"' >> /usr/lib/python3.6/site-packages/inginious_container_api/run_types.py
 
 As easily seen, this Dockerfile creates a new container image that can launch PHP scripts.
-The syntax of these Dockerfiles is extensively described on the website of Docker_, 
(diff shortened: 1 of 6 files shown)


Drumor (approved):
LGTM

Issue #433, opened 2019-09-12
Feature request : Hiding of multiple/all exerices of a course at one


kimmens:
Being able to hide a large series, or all, exercises of a course. Typically very useful at the beginning of an academic year when you want to make available only the exercises relevant for the course matter that has already been taught.

nrybowski:
Done with #982

PR #917, opened 2023-03-03
Adding "hidden" config item to Section

### LINKED ISSUE #914: [frontend/course administration] Add a way to hide all tasks from a section / hide a section
Currently, there is no way to hide a section to the students or quickly hide all tasks within a given section.

Add an option on the sections similar to the closed by default to make all tasks and/or the section itself hidden/visible.

Drumor:
This PR proposes an extension of section configuration by adding an "hidden" item to simply hide a section.

It doesn't influence calculation of scores which remains for discussion:
Does this feature should influence score or does it simply hide a section for a user.

Done in this PR:

- Addition of an "hidden" config for a section
- Update of templates to hide a section if the config is set to True.

Once merged, this PR resolves #914 and #433

### DIFF
--- a/inginious/frontend/task_dispensers/toc.py
+++ b/inginious/frontend/task_dispensers/toc.py
@@ -107,11 +107,9 @@ class TableOfContents(TaskDispenser):
                                       course_structure=self._toc, tasks=task_data, config_fields=config_fields,
                                       config_items_funcs=["dispenser_util_get_" + config_item.get_id() for config_item in self.config_items])
 
-    def render(self, template_helper, course, tasks_data, tag_list):
+    def render(self, template_helper, course, tasks_data, tag_list, username):
         """ Returns the formatted task list"""
-        course_req = self._database.courses.find_one({"_id":course.get_id()})
-        people = course_req["students"] if course_req is not None else [] + course.get_staff()
-        accessibilities = course.get_task_dispenser().get_accessibilities(self._task_list_func(),people)
+        accessibilities = course.get_task_dispenser().get_accessibilities(self._task_list_func(),[username])
         return template_helper.render("task_dispensers/toc.html", course=course, tasks=self._task_list_func(),
                                       tasks_data=tasks_data, tag_filter_list=tag_list, sections=self._toc,accessibilities=accessibilities)
 
--- a/inginious/frontend/task_dispensers/toc.py
+++ b/inginious/frontend/task_dispensers/toc.py
@@ -100,7 +100,8 @@ class TableOfContents(TaskDispenser):
     def render_edit(self, template_helper, course, task_data):
         """ Returns the formatted task list edition form """
         config_fields = {
-            "closed": SectionConfigItem(_("Closed by default"), "checkbox", False)
+            "closed": SectionConfigItem(_("Closed by default"), "checkbox", False),
+            "hidden": SectionConfigItem(_("Hidden by default"),"checkbox",False)
         }
         return template_helper.render("course_admin/task_dispensers/toc.html", course=course,
                                       course_structure=self._toc, tasks=task_data, config_fields=config_fields,
--- a/inginious/frontend/task_dispensers/toc.py
+++ b/inginious/frontend/task_dispensers/toc.py
@@ -109,8 +109,10 @@ class TableOfContents(TaskDispenser):
 
     def render(self, template_helper, course, tasks_data, tag_list):
         """ Returns the formatted task list"""
+        people = self._database.courses.find_one({"_id":course.get_id()})["students"] + course.get_staff()
+        accessibilities = course.get_task_dispenser().get_accessibilities(self._task_list_func(),people)
         return template_helper.render("task_dispensers/toc.html", course=course, tasks=self._task_list_func(),
-                                      tasks_data=tasks_data, tag_filter_list=tag_list, sections=self._toc)
+                                      tasks_data=tasks_data, tag_filter_list=tag_list, sections=self._toc,accessibilities=accessibilities)
 
     def check_dispenser_data(self, dispenser_data):
         """ Checks the dispenser data as formatted by the form from render_edit function """
--- a/inginious/frontend/task_dispensers/toc.py
+++ b/inginious/frontend/task_dispensers/toc.py
@@ -101,7 +101,7 @@ class TableOfContents(TaskDispenser):
         """ Returns the formatted task list edition form """
         config_fields = {
             "closed": SectionConfigItem(_("Closed by default"), "checkbox", False),
-            "hidden": SectionConfigItem(_("Hidden if empty"),"checkbox",False)
+            "hidden_if_empty": SectionConfigItem(_("Hidden if empty"),"checkbox",False)
         }
         return template_helper.render("course_admin/task_dispensers/toc.html", course=course,
                                       course_structure=self._toc, tasks=task_data, config_fields=config_fields,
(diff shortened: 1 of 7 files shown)


anthonygego:
This implementation is pretty dangerous as it leads to misunderstanding. Section will be hidden by tasks will still be accessible to the students if their settings allow them to.

The most relevant implementation for current master branch should be "Hidden if empty" and behaves as following : if the section remains empty for a given student, it is simply not shown.

anthonygego:
This still does not do the job. The section must not be shown if it results, for a given user, in an empty section. That means, if all tasks of a section are not accessible.

inginious/frontend/templates/task_dispensers/task_list.html
```diff
@@ -1,57 +1,57 @@
 {# This file is part of INGInious. See the LICENSE and the COPYRIGHTS files for #}
```
anthonygego, line 1: There should be no need to change this file as it does not need to be included in the section is hidden.

inginious/frontend/task_dispensers/util.py
```diff
@@ -144,6 +144,9 @@ def get_config(self):
         :return: the config dict of this section
         """
         return self._config
+    @staticmethod
+    def have_accessible_tasks(tasks):
+        return any([task.get_accessible_time().is_open() for task in tasks])
```
anthonygego, line 149: It's only used at one place. I'd rather put this line there in `inginious/frontend/templates/task_dispensers/toc.html`

Current changes are not better OOP: while it actually concerns a section, it does not have sense to ask for the tasks as they could be contained in the section properties.

inginious/frontend/task_dispensers/toc.py
```diff
@@ -104,7 +104,8 @@ def get_dispenser_data(self):
     def render_edit(self, template_helper, course, task_data):
         """ Returns the formatted task list edition form """
         config_fields = {
-            "closed": SectionConfigItem(_("Closed by default"), "checkbox", False)
+            "closed": SectionConfigItem(_("Closed by default"), "checkbox", False),
+            "hidden": SectionConfigItem(_("Hidden if empty"),"checkbox",False)
```
anthonygego, line 108: `hidden` => `hidden_if_empty`, otherwise confusing.

anthonygego (changes requested):

inginious/frontend/templates/task_dispensers/toc.html
```diff
@@ -6,35 +6,38 @@
...
-        {% endwith %}
-    {% else %}
-        <div id="section_{{ section.get_id() }}" class="section sections-list mb-4">
-            <div class="section_header d-flex justify-content-between divided pr-3" onclick="dropdown(this)">
-                <span class="title h{{level}} mr-3">
-                    {{ _(section.get_title()) }}
-                </span>
-                <span class="divider"></span>
+    {% set ns = namespace(section_tasks=[],have_accessible_tasks=False) %}
+    {% for taskid in section.get_tasks() if taskid in tasks %}
```
anthonygego, line 10: This will compute the accessibility multiple times. Non-terminal sections `get_tasks()` is recursive and will return the tasks of all the terminal sections it contains.

It would be better on highly structured courses to avoid this behaviour. Maybe it's better to wait for #908 to fix this.

Drumor: I solved multiple computation problem but have to call get_tasks for each section... Probably still improvable?

Drumor:
As pair reflexion with @anthonygego, this PR should be fully reworked.

1. To improve complexity of recursive sections journey and accessibility verification, a first journey should be done before template rendering to calculate a per section accessibility.
2. Method is_empty() in Section may probably be deleted as it's probably useless.

inginious/frontend/task_dispensers/toc.py
```diff
@@ -100,16 +100,20 @@ def get_dispenser_data(self):
...
         }
         return template_helper.render("course_admin/task_dispensers/toc.html", course=course,
                                       course_structure=self._toc, tasks=task_data, config_fields=config_fields,
                                       config_items_funcs=["dispenser_util_get_" + config_item.get_id() for config_item in self.config_items])
 
     def render(self, template_helper, course, tasks_data, tag_list):
         """ Returns the formatted task list"""
+        course_req = self._database.courses.find_one({"_id":course.get_id()})
+        people = course_req["students"] if course_req is not None else [] + course.get_staff()
+        accessibilities = course.get_task_dispenser().get_accessibilities(self._task_list_func(),people)
```
anthonygego, line 114: Only the session user accessibilities have to be computed. You can also check and propagate this big dictionary accross all the templates (`task_list.html` for instance) to avoid recomputing accessibilities when printing each task.

inginious/frontend/templates/course_admin/task_dispensers/section.html
```diff
@@ -2,7 +2,7 @@
 {# more information about the licensing of this file. #}
 
 {% for section in sections %}
-    {% if section.is_empty() %}
+    {% if section.get_tasks() | length ==0  %}
```
anthonygego, line 5: I fear that doing so has quite the same result as the one deleted for accessibilities : it will recursively navigate and count tasks for each subsection. Maybe `is_empty` remains the best option here.

inginious/frontend/task_dispensers/toc.py
```diff
@@ -100,16 +100,18 @@ def get_dispenser_data(self):
...
-            "closed": SectionConfigItem(_("Closed by default"), "checkbox", False)
+            "closed": SectionConfigItem(_("Closed by default"), "checkbox", False),
+            "hidden_if_empty": SectionConfigItem(_("Hidden if empty"),"checkbox",False)
         }
         return template_helper.render("course_admin/task_dispensers/toc.html", course=course,
                                       course_structure=self._toc, tasks=task_data, config_fields=config_fields,
                                       config_items_funcs=["dispenser_util_get_" + config_item.get_id() for config_item in self.config_items])
 
-    def render(self, template_helper, course, tasks_data, tag_list):
+    def render(self, template_helper, course, tasks_data, tag_list, username):
```
anthonygego, line 110: You'll need to update the `render` function signature in other task dispensers. I think only `CombinatoryTest` reimplements it.

inginious/frontend/templates/task_dispensers/toc.html
```diff
@@ -4,37 +4,39 @@
...
-        <div class="section sections-list mb-4">
-            <div class="section_header d-flex justify-content-between divided pr-3" onclick="dropdown(this)">
-                <span class="title h{{level}} mr-3">
-                    {{ _(section.get_title()) }}
-                </span>
-                <span class="divider"></span>
+    {% set ns = namespace(section_tasks=[],have_accessible_tasks=False) %}
+    {% for taskid in section.get_tasks() if taskid in tasks %}
+        {% set ns.section_tasks = ns.section_tasks + [tasks[taskid]] %}
+        {% set ns.have_accessible_tasks = ns.have_accessible_tasks or user_manager.user_is_superadmin() or accessibilities[user_manager.session_username()][taskid].is_open()%}
```
anthonygego, line 11:
```suggestion
        {% set ns.have_accessible_tasks = ns.have_accessible_tasks or accessibilities[user_manager.session_username()][taskid].is_open()%}
```
It has been decided with the introduction of task dispensers to make the student view visible as a student to the course administrator. Bypassing the check for administrators would confuse the user in parameterizing the course table of contents.

But bypassing for superadmins only is even more confusing and doesn't seem to make sense ?

anthonygego (changes requested):

inginious/frontend/templates/task_dispensers/toc.html
Drumor: You're right... Forgot the superadmin case...

PR #918, opened 2023-03-06
Adding a centralized feedback feature for match problem

### LINKED ISSUE #909: [problem_types/match] Support silent feedback for match problems

MCQ problems allows to hide feedback in case of errors such that only the general task feedback indicates there are errors in the submission.

Match problem types does not allow such a feature and the student always know if the sub-problem is correct or wrong. 

It would be useful not to show specific feedback for match problems. This implies addition of a config field in the match problem and filtering feedback based on that config fields in the `check_answer` method.

Drumor:
This PR resolves #909 by proposing a new parameter (called "centralize") in match problem configuration. This will let the administrator decide to display or not a specific feedback to each match sub problem.

Done in this PR:
- Adding parameter "centralize"
- Filtering feedback based on the parameter in the `check_answer` method.
- Add edition of the parameter in the studio.


anthonygego:
LGTM

Issue #897, opened 2022-11-29
[ci] INGInious containers packaging and distribution


nrybowski:
**Is your feature request related to a problem? Please describe.**
The containers available on DockerHub are quite outdated because they are replaced by hand.

**Describe the solution you'd like**
The exercise containers and the future services (see https://github.com/UCL-INGI/INGInious/pull/882) should be built and pushed on public registries automatically through the CI.

The same build process should be applied for each supported version (e.g. v0.7 and v0.8) and the development version.

**Additional context**
Automated tools (e.g. https://github.com/UCL-INGI/.github/pull/1) deploying parts of or the whole INGInious project would benefit from that.

nrybowski:
Dockerhub stopped its free plan for Free Organizations. We should migrate on the github registry (https://docs.github.com/en/packages) and / or quay.io (https://quay.io/).

anthonygego:
Indeed, compose containers are not built automatically.

nrybowski:
We should use the same method as https://github.com/INGInious/INGInious/commit/2585840e54b7fedf71916beda7a7929472ca4b78 to build and deploy the core containers.

PR #921, opened 2023-03-16
Update container URL to GitHub registry

nrybowski:
Move containers from Dockerhub to GitHub packages. See #897 for details.

### DIFF
--- a/base-containers/default/Dockerfile
+++ b/base-containers/default/Dockerfile
@@ -2,5 +2,5 @@
 
 #inherit from the base container, which have all the needed script to launch tasks
 ARG   VERSION=latest
-FROM  ingi/inginious-c-base:${VERSION}
+FROM  ghcr.io/inginious/env-base:${VERSION}
 LABEL org.inginious.grading.name="default"


nrybowski:
This PR should be squashed on merge.

.github/workflows/env_containers.yml
```diff
@@ -0,0 +1,19 @@
...
+
+  workflow_dispatch:
+
+jobs:
+  containers_build_and_push:
+    uses: INGInious/.github/.github/workflows/containers.yml@673723b119b1a92695dd5fea1520f5ae28e38abb
+    with:
+      working-directory: base-containers 
+      context-path: context.yml
+      compose-path: compose.yml
```
anthonygego, line 15: Where is this file ?

nrybowski: It is generated by the [reusable workflow](https://github.com/INGInious/.github/blob/main/.github/workflows/containers.yml) called in the `uses` instruction.

anthonygego (approved):

PR #931, opened 2023-04-03
[frontend/theme] script for bootswatch theme compilation

Ananas120:
Script to compile (new) bootswatch theme

Usage : 
- Go to `inginious/frontend/static/css/themes`
- Run `make yeti` to compile the yeti theme


inginious/frontend/static/css/themes/dark_mode.css
anthonygego, line 1: This is the compiled version. The source file should be stored somewhere.

inginious/frontend/static/css/themes/Makefile
```diff
@@ -0,0 +1,16 @@
+BOOTSWATCH_URL	= https://github.com/thomaspark/bootswatch.git
+BASE_THEME	= yeti
+
+compile:
+	if [ ! -d "bootswatch" ]; then make install; fi
+	cd bootswatch; ./node_modules/grunt/bin/grunt swatch:$(theme)
+	mv bootswatch/dist/$(theme)/bootstrap.css $(theme).css
+
+install:
+	git clone $BOOTSWATCH_URL
```
anthonygego, line 10: In this case we do not use a fixed version as discussed before. @nrybowski suggested to use a submodule to include the source at a given commit stage. We can probably add a `third-party` folder at the source root to store the submodule.

Besides, as discussed previously, it would probably be better to store the source files for yeti and the modified yeti dark mode into our repository (keeping their copyright of course) and only keep a submodule to bootstrap.

In all case we need to be able to rebuild the `dark_mode.css` file from the sources.

inginious/frontend/static/css/themes/Makefile
```diff
@@ -0,0 +1,16 @@
+BOOTSWATCH_URL	= https://github.com/thomaspark/bootswatch.git
+BASE_THEME	= yeti
+
+compile:
+	if [ ! -d "bootswatch" ]; then make install; fi
```
anthonygego, line 5: It would be more convenient if all these steps could be carried out by grunt to avoid adding development dependencies. In case the yeti sources are provided in our repo the build configuration file should be provided too and can be modified to place the file in the appropriate locations.

inginious/frontend/templates/layout.html
```diff
@@ -21,7 +21,7 @@
         <!-- INGInious Style -->
         <link href="{{get_homepath(True)}}/static/css/INGInious-rtl.css?v2" rel="stylesheet">
     {% else %}
-        <link href="{{get_homepath(True)}}/static/css/bootstrap.min.css" rel="stylesheet">
+        <link href="{{get_homepath(True)}}/static/css/themes/default.min.css" rel="stylesheet">
         <!-- INGInious Style -->
         <link href="{{get_homepath(True)}}/static/css/INGInious.css?v2" rel="stylesheet">
```
anthonygego, line 26: Maybe we should also provide a specific version of this file for the different themes ?

anthonygego (changes requested):
All the source files should be present so that at least the devs can rebuild the themes. We agreed previously to keep a submodule of bootswatch to be sure we are using a version compatible with our sources. It's probably also possible to keep the all process done by grunt only not to add additional development dependencies.

Ananas120:
I have updated the workflow to remove the need of any third-party module by copying the Gruntfile + build directory from the bootswatch repo (which is enough for theme compilation)
I have also added the "darkly" theme with the "dracula" CodeMirror theme (not fully tested in inginious + theme not deployable yet)

I left the Makefile for simplicty but it is not required

themes/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "release": "grunt release",
+    "test": "npm run stylelint"
+  },
+  "files": [
+    "dist/**/*.{css,scss}"
+  ],
+  "devDependencies": {
+    "@lodder/grunt-postcss": "^3.1.1",
+    "autoprefixer": "^10.4.12",
+    "bootstrap": "^5.2.3",
```
anthonygego, line 29: The current codebase still relies on Bootstrap 4. This is probably the first change to do (the first comment to treat).

themes/Gruntfile.js
```diff
@@ -0,0 +1,266 @@
+'use strict';
+
+const path = require('path');
+const sass = require('sass');
+const autoprefixer = require('autoprefixer');
+const pkg = require('./package.json');
+
+const SWATCHES = [
```
anthonygego, line 8: This can probably be cleaned

themes/build/html/thumbnail.html
```diff
@@ -0,0 +1,125 @@
+<!DOCTYPE html>
```
anthonygego, line 1: We probably don't need to keep any thumbnail rendering feature. I guess ?

themes/test.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
```
anthonygego, line 1: Test file should not be pushed into the repository

themes/Gruntfile.js
```diff
@@ -0,0 +1,266 @@
+'use strict';
```
anthonygego, line 1: We should keep a note here about the original author.

themes/Gruntfile.js
```diff
@@ -0,0 +1,266 @@
...
+    const isValidTheme = grunt.file.exists(path.join(themeDir, '/_variables.scss')) && grunt.file.exists(path.join(themeDir, '/_bootswatch.scss'));
+
+    // cancel the build (without failing) if this directory is not a valid theme
+    if (!isValidTheme) {
+      grunt.log.writeln(`${theme} theme does not exist!`);
+      return;
+    }
+
+    const concatSrc = path.join(BUILD_DIR, '/scss/build.scss');
+    const concatDest = path.join(themeDir, '/build.scss');
```
anthonygego, line 202: Based on my understanding, the two following files are generated and can be removed from the repository.

themes/Gruntfile.js
```diff
@@ -0,0 +1,266 @@
...
+  'superhero',
+  'united',
+  'vapor',
+  'yeti',
+  'zephyr'
+];
+
+const BUILD_DIR = 'build/';
+const DIST_DIR = 'themes/';
+const DOCS_DEST = 'docs/5/';
```
anthonygego, line 38: I think that we can delete all things related to docs as we don't aim at redistributing a full modified bootstrap copy.

.gitignore
```diff
@@ -23,7 +23,12 @@ venv
 inginious/tasks
 # Default backup directory
 inginious/backup
+# The themes files
+themes/node_modules
```
anthonygego, line 27: The `themes` folder can be put into the `frontend` one. This would make sense.

themes/Makefile
```diff
@@ -0,0 +1,16 @@
+BASE_THEME	= yeti
+THEMES_DIR	= themes
+OUTPUT_DIR	= ../inginious/frontend/static/css/themes
+
+compile:
+	if [ ! -d "node_modules" ]; then make install; fi
```
anthonygego, line 6: I actually don't ask this to be modified, but `make` supports dependencies handling. That's one of his core features.

themes/Makefile
```diff
@@ -0,0 +1,16 @@
...
+	if [ ! -d "$(OUTPUT_DIR)/$(theme)" ]; then cp -r $(OUTPUT_DIR)/default $(OUTPUT_DIR)/$(theme); fi
+	./node_modules/grunt/bin/grunt swatch:$(theme)
+	mv $(THEMES_DIR)/$(theme)/bootstrap.min.css $(OUTPUT_DIR)/$(theme)/.
```
anthonygego, line 9: There should clearly be way to modify `Gruntfile.js` in order to compile all themes as the default action, and to copy the `bootstrap.min.css` in the appropriate folder.

This way, this whole Makefile would be kind of useless and compiling the themes would just consists in `npm install` and `grunt`.

themes/Makefile
```diff
@@ -0,0 +1,16 @@
...
+
+install:
+	npm install
+	npm install grunt-cli
```
anthonygego, line 13: You can add `grunt-cli` as a dependency in `package.json`.

anthonygego (changes requested):
I made comments based on reading the code only. Please fetch the files for Bootstrap 4 in the base repo first.

inginious/frontend/app.py
```diff
@@ -175,6 +175,15 @@ def get_app(config):
     available_languages = {"en": "English"}
     available_languages.update(available_translations)
 
+    available_themes = [
+        theme_dir for theme_dir in os.listdir(os.path.join('inginious', 'frontend', 'static', 'css', 'themes'))
+        if theme_dir not in ('yeti', 'codemirror') and not theme_dir.startswith(('.', '_'))
```
anthonygego, line 180: Not sure `yeti` is required as an exception here.

inginious/frontend/app.py
```diff
@@ -175,6 +175,15 @@ def get_app(config):
     available_languages = {"en": "English"}
     available_languages.update(available_translations)
 
+    available_themes = [
+        theme_dir for theme_dir in os.listdir(os.path.join('inginious', 'frontend', 'static', 'css', 'themes'))
```
anthonygego, line 179: You can use ``inginious.get_root_path()`` here. This will avoid issues with the current working directory environment variable. Typically in a production setup where the `inginious-webapp` script is not in the same directory as the `inginious` package.

inginious/frontend/app.py
```diff
@@ -175,6 +175,15 @@ def get_app(config):
...
+        if theme_dir not in ('yeti', 'codemirror') and not theme_dir.startswith(('.', '_'))
+    ]
+    available_codemirror_themes= [
+        file.split('.')[0] for file in os.listdir(os.path.join('inginious', 'frontend', 'static', 'css', 'themes', 'codemirror'))
```
anthonygego, line 183: Same comment here for the path.

inginious/frontend/parsable_text.py
```diff
@@ -35,6 +35,8 @@ def run(self):
         if not self.content:
             translation = _get_inginious_translation()
             self.content = [translation.gettext("[no content]")]
+        
+        self.options.setdefault('classes', []).append('theme={}'.format(flask.current_app.user_manager.get_codemirror_theme()))
```
anthonygego, line 39: That should be OK but maybe it's more careful and idiomatic (it seems to be the de-facto rule) to use `-` instead of `=` to separate the prefix of your CSS class. 

In case of bad HTML, browsers typically try to fix the error by themselves. A `=` is part of the HTML tag syntax and may be badly interpreted as an attribute in case a correction is needed. Of course we aim at providing valid HTML to the client, but those things can still happen.

inginious/frontend/static/js/common.js
```diff
@@ -46,14 +46,18 @@ function colorizeStaticCode()
...
+        var theme = undefined;
         $.each(classes, function(idx, elem)
         {
             if(elem != "code" && elem != "literal-block")
             {
                 var nmode = CodeMirror.findModeByName(elem);
-                if(nmode != undefined)
+                if(nmode != undefined) {
                     mode = nmode;
+                } else if (elem.startsWith('theme=')) {
```
anthonygego, line 58: ... continuation of my comment about `-`and `=`.

inginious/frontend/user_manager.py
```diff
@@ -33,6 +33,14 @@ class AuthInvalidMethodException(Exception):
 
 class AuthMethod(object, metaclass=ABCMeta):
 
+    @abstractmethod
```
anthonygego, line 36: Those methods do not have to be defined in `AuthMethod` class as there should be no dependance between them.

inginious/frontend/user_manager.py
```diff
@@ -107,6 +115,12 @@ def sanitize_email(cls, email: str) -> str:
     #           User session management          #
     ##############################################
 
+    def get_theme(self):
```
anthonygego, line 118: I'd rather call them `session_theme` and `session_codemirror_theme` to use the same nomenclature and recall throughout the code that those values are fetched from the user session.

inginious/frontend/templates/tasks/code.html
```diff
@@ -7,6 +7,7 @@
 
 <textarea name="{{inputId}}"
           class="code-editor form-control {% if '/' in inputId %} single {% endif %}"
+          theme="dracula"
```
anthonygego, line 10: Hardcoded value :-)

inginious/frontend/themes/Gruntfile.js
```diff
@@ -0,0 +1,170 @@
+'use strict';
```
anthonygego, line 1: You can rename `themes/themes` into `themes/src` to be less confusing.

You should probably add an initial comment such as 
   
    Based on Bootswatch from Thomas Park (https://github.com/thomaspark/bootswatch)
    
in the beginning of the imported files

anthonygego (changes requested):
Do you think you can split this PR in :

- One with the `themes` and `inginious/frontend/static/css` changes. (that is, essentially the imported files and building logic)
- One with the remaining changes (that is, the effective integration into INGInious)

That would allow me to make sure my review is complete as I don't want to review all the imported CSS files, the review listing is a bit messy right now :-)

**Do not delete this one now**

Ananas120:
I still update this branch PR to be sure, I will now try to create a new one on the tasksets branch...

Ananas120:
All the commands you told me have (in theory) been successfully executed, and conflicts solved ! Hope it is correct now
Note that for the "theme=" comment, I have changed to "theme:" instead of "theme-" because some codemirror themes have the "-" symbol, which may raise errors when splitting it

anthonygego:
I'll close this pull-request as it has many conflicts with the current `main` branch and is no more a priority. Feel free to rebase and re-open it in the future.

PR #933, opened 2023-04-05
Remove social share feature

Drumor:
This PR removes mostly dead code used for the share feature on social media.
This PR is done in 2 commits:

1. The code removal
2. The update of minified files.


inginious/frontend/pages/social.py
```diff
@@ -43,14 +43,6 @@ def process_callback(self, auth_id):
         if user and auth_storage.get("method", "") == "signin":
```
anthonygego, line 43: There should be no more need to store the method in `auth_storage` and there probably is some additional code that can be removed.

PR #937, opened 2023-04-13
adding a lighten download csv button

### LINKED ISSUE #936: [course_admin/submissions] lighten CSV outputs
When you filter submissions in the submissions view and then download the CSV, you get all the submission info, feedback included. The later may be very large and not useful in a CSV download, especially if you only want to draw charts from the data.

We should be able to select or filter out rows (that could be in a modal for instance, and the checked rows can be saved in cookie, session, or whatever), or to have a lighter download options.

Drumor:
This "dummy" solution fix #936 by adding a lighten csv download button.

### DIFF
--- a/inginious/frontend/pages/course_admin/utils.py
+++ b/inginious/frontend/pages/course_admin/utils.py
@@ -276,7 +276,7 @@ class UnicodeWriter(object):
             self.writerow(row)
 
 
-def make_csv(data, lighten=False):
+def make_csv(data, lighten=False, fields =[]):
     """ Returns the content of a CSV file with the data of the dict/list data """
     # Convert sub-dicts to news cols
     for entry in data:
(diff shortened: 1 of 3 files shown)


inginious/frontend/pages/course_admin/utils.py
```diff
@@ -296,19 +296,22 @@ def make_csv(data):
...
     # Convert everything to CSV
-    columns = set()
     output = [[]]
-    if isinstance(data, dict):
-        output[0].append("id")
-        for entry in data:
-            for col in data[entry]:
-                columns.add(col)
+    if lighten:
+        columns = ["_id","best","courseid","grade","result","status","submitted_on","taskid","username"]
```
anthonygego, line 301: This goes against the principles of the function code, which remains content agnostic. This is one of the things we should be careful doing Python code. Things hardcoded here will clearly be missed in case of refactor (changing field names, ...).

A more robust solution should be dynamic. Something even more convenient would be to select the displayable columns in the frontend UI and send the columns ids from the form. 

Another valid and more straightforward solution, as what is called here "lightened csv" is to provide the current visible columns only, is simply to make the form send back the shown columns ids.

anthonygego (changes requested):

anthonygego:
I suggest to close this PR as this is better to generate the available fields on the fly based on what's inside the submission dict, and to propagate it in HTML and then in Python, instead of hardcoding the fields in the code.

Issue #887, opened 2022-11-02
[backend] SSH tasks wait in the queue if no Docker agent has SSH enabled


anthonygego:
**Describe the bug**
An SSH task can wait in the queue for an agent even if there is no SSH-enabled agent available.

**INGInious installation details**
- Version: v0.7+

**To Reproduce**
1. Deploy an INGInious instance without enabling SSH on Docker agents
1. Submit a task that requires SSH
1. The task wait undefinetely 

**Expected behavior**
The backend queue should return that no connected agent is able to perform the task and remove the job from the queue.

PR #940, opened 2023-04-20
[agent/docker] replace SSH opt-in by env types

anthonygego:
This PR addresses issue #887 in which SSH jobs can wait indefinitely in the queue if no agent has SSH capability enabled because jobs are added based on the environment availability and removed based on what agents are capable.

It replaces the `ssh_allowed` environment parameter opt-in by multiple environment types. This ensures the job will always be treated as it won't be sent to the queue if the environment type has not been announced. It also removes any  SSH awareness from the queue and keep that concept in the specific environment code.

Another solution would have been to make the client code aware of SSH and prevent the job from getting in the queue, but that would have been heavier in terms of lines of code and would have introduced agent specificities into the common code.

### DIFF
--- a/inginious/agent/__init__.py
+++ b/inginious/agent/__init__.py
@@ -48,7 +48,7 @@ class Agent(object, metaclass=ABCMeta):
     An INGInious agent, that grades specific kinds of jobs, and interacts with a Backend.
     """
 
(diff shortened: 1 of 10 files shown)


nrybowski:
LGTM

Drumor (approved):
LGTM

PR #941, opened 2023-04-26
[frontend/submission_manager] catch DocumentTooLarge exception

### LINKED ISSUE #697: Frontend fails to change submission status if container stdout > max DB document size
1) Frontend fails to change submission status if container stdout > max DB document size

    If a student prints an amount of data between the maximum size allowed for a MongoDB document, and the maximum size allowed by the agent, then it is never stopped until the frontend fails to update the document in database because stdout output is stored in the document. The submission then remains in ``waiting`` state, preventing the student to submit.

    There should be an additional check for that, probably directly in the agent as there should be no need for a >4MB stdout to be stored at the end, even for debugging.

2) Nothing prevents a too-long submission archive, but every component (Docker, backend, Mongo) can suffer from such a behaviour. Observed behaviour with GB produced archive was Docker socket timing out.

anthonygego:
Fixes #697.

This is a first mitigation and in all case useful, as the limit is set on the total command sent to the MongoDB server, which cannot be computed before.

Do not update user stats in case of error.

Runfile to test : 
      
      #!/bin/bash
      dd if=/dev/random of=test.out bs=4M count=100
      cat test.out | tr -dc '[:alpha:]' | feedback-msg -a
      feedback-result success

### DIFF
--- a/inginious/frontend/submission_manager.py
+++ b/inginious/frontend/submission_manager.py
@@ -76,16 +76,26 @@ class WebAppSubmissionManager:
         }
 
         # Save submission to database
-        submission = self._database.submissions.find_one_and_update(
-            {"_id": submission["_id"]},
-            {"$set": data, "$unset": unset_obj},
-            return_document=ReturnDocument.AFTER
-        )
+        try:
+            submission = self._database.submissions.find_one_and_update(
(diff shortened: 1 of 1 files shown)


Drumor (approved):
LGTM

PR #949, opened 2023-05-23
[base/kvm] Add KVM enabled base container

### LINKED ISSUE #939: Runtime environments allowing to run as root could provide more capabilities to the containers
Kata runtime could give more priviledge to the root user in the Docker container. Running IPMininet in a Docker container requires the ability of manipulating network namespaces.

When the runtime allows to run as root, pass --cap_add=ALL to the container. Do this here,
https://github.com/UCL-INGI/INGInious/blob/145fc847f751ecbf1c8b52858c3a2a3e47b52670/inginious/agent/docker_agent/_docker_interface.py#L137

**Describe alternatives you've considered**
Changing the use of Kata+Docker for virtme or smth else

nrybowski:
This PR adds a new base container able to launch a specific Linux kernel within a KVM.
New grading environments should inherit from this container as it is the case for the base container.
Task leveraging such grading environments must support the SSH feature of INGInious.

## TODO
- [x] Patch virtme to enable port forwarding for telnet
- [x] Add a 'KVM' label in the base container to dynamically enable KVM passthrough on the Agents
- [x] Add 'KVM' flag to Docker Agent
     - [x] Check if KVM is enabled on the Agent environment
- [x] Automate KVM launch in student container
- [x] Automate shell redirection to KVM in student container
- [x] Hardcode qemu-kvm src rpm url
- [ ] Add documentation

## TODO in further PRs
- Use serial port to spawn a shell (see virtme) in the VM rather than using telnet and port forwarding
- Log the commands typed in the telnet shell as here

https://github.com/UCL-INGI/INGInious/blob/a9962cad2a18caf87a73f10c627fc98f561f0e6c/base-containers/base/inginious_container_api/utils.py#L79

Such logs could be used in further grading scripts.


base-containers/kvm/Dockerfile
```diff
@@ -0,0 +1,83 @@
...
+	usbredir-devel \
+	yum-utils \
+	'pkgconfig(epoxy)' \
+	'pkgconfig(gbm)' \
+	'pkgconfig(libdrm)'
+
+# Apply patches and build QEMU
+WORKDIR /opt
+RUN 	yum-builddep -y qemu-kvm
+RUN 	yumdownloader --source qemu-kvm
```
anthonygego, line 32: As you are providing patches, it'll be more robust to hardcode the src rpm version/link. Otherwise at list the spec file will conflict once the maintainers push an new patch.

nrybowski: This fix was on my todo list for this PR :)

nrybowski:
This PR should solve https://github.com/UCL-INGI/INGInious/issues/939.

base-containers/kvm/Dockerfile
nrybowski: Fixed in https://github.com/UCL-INGI/INGInious/pull/949/commits/703ba30cf2c2d4d6d6b2dfdf9fb27e55c8045e28

nrybowski:
Some slowness is observed during the execution of a simple mininet command such as `mn --switch lxbr --topo tree,depth=2,fanout=8 --test pingall` at the startup of the fifth switch.
The VM has two cores and 256Mo of RAM but none of those ressources are used at their maximum.
Further investigation is required but do not prevent the integration of this feature which is anyway still experimental.

inginious/agent/docker_agent/_docker_interface.py
```diff
@@ -150,13 +165,14 @@ def create_container(self, image, network_grading, mem_limit, task_path, sockets
                 course_common_student_path: {'bind': '/course/common/student', 'mode': 'ro'}
             },
             runtime=runtime,
-            ulimits=[nofile_limit]
+            ulimits=[nofile_limit],
+            devices=['/dev/kvm'] if kvm and image_requires_kvm else []
```
anthonygego, line 169: I fear that unwanted behaviours can happen silently here. If the container image requires kvm and the agent does not support/enable it, no notification will be given. There seems to be no way for the user to know if the agent supports kvm, and, moreover, nothing prevents the job to be assigned to an inappropriate agent.

I'm however more likely to deploy an agent that do support kvm into a cluster of other agents that don't.

For me it would probably be more relevant to implement this as the ssh or nvidia filters : announce another agent type that will allow container images tagged with `org.inginious.kvm` to be listed as available environment, so that we can ensure appropriate routing and ensure the user selected the right environment.

Implementation of `nvidia` is here https://github.com/UCL-INGI/INGInious/commit/f842bee563f8eb7bca3957e4f5019c45dbd68e6d

However, I realize that implementing a kvm filter, as it may be compatible with other agent type, would require to duplicate all those one to have a kvm-enabled flavour. 

Maybe it would be useful to rethink the way this is done and add some kind of feature flags in addition to the agent type, so that we can handle `docker` or `nvidia` with `ssh` or `kvm` as supported features or not.

base-containers/kvm/Dockerfile
```diff
@@ -0,0 +1,102 @@
...
+RUN	wget "https://ftp.gnu.org/gnu/inetutils/inetutils-2.4.tar.xz" &&\
+	tar xf inetutils-2.4.tar.xz &&\
+	cd inetutils-2.4 &&\
+	./configure --disable-servers --disable-clients --enable-telnetd --enable-telnet &&\
+	make -j$(nproc)
+
+# =====================
+# KVM base container
+# =====================
+FROM 	${REGISTRY}/inginious/env-base:${VERSION}
```
anthonygego, line 69: Two `FROM` directives in the same Dockerfile ?

base-containers/base/bin/_run_student_intern
```diff
@@ -21,6 +21,7 @@ logger = setup_logger()
 # Check the runtimes
 runtime = sys.argv[1]
 parent_runtime = sys.argv[2]
+kvm_enabled = sys.argv[3] == "True"
```
anthonygego, line 24: I'd document the args at both sides, here and at the call side in `_docker_interface.py` (to avoid implementing a useless argumentparser that would be self documenting)

anthonygego (commented):
This is very a first pass. I'd like to see those issues addressed first because I'm wondering if there is a possibility to make this fully modular and if it would be relevant.

Indeed, the added lines seem very tied to the additional container architecture and make code harder to understand. I would know if it's possible to put these additional piece of codes in a kind of hook folder provided in the container image.

base-containers/kvm/Dockerfile
nrybowski: This is a multi-stage build (https://docs.docker.com/build/building/multi-stage/).
The first stage generates a patched QEMU RPM with P9FS enabled and the second stage is the actual container.

nrybowski: The first stage also builds the GNU implementation of telnet which is not packaged for RL8.

base-containers/base/bin/_run_student_intern
nrybowski: The arguments managment will be replaced by an argparser.

nrybowski:
[v1.32 of virtme](https://github.com/arighi/virtme-ng/releases/tag/v1.32) now supports SSH. We should bump the version and replace the telnet hack with this new feature.

anthonygego:
Can you rebase your branch against `main` (you'll probably have to cherry-pick) ? I think there should be no or very few conflicts to solve.

inginious/agent/docker_agent/_docker_interface.py
nrybowski: Three years later, the filtering system is implemented in #1154... We finally will be able to integrate KVM-based tasks in INGInious.

PR #959, opened 2023-06-22
Identify API request with HTTP Authorization Bearer

### LINKED ISSUE #185: Make webapp API work with API keys
Currently the webapp API requires users to login and client to keep the session cookie and send it back for each HTTP request. This login process will fail with some authentication methods such as the recently added SAML2, which require serveral web redirections...

A better way to handle this is to work with API keys and let users define the applications that are allowed to access their profile.

Another interesting feature would be to handle anonymous submissions (currently done via a plugin) quite the same way.

Drumor:
Resolves #185

### DIFF
--- a/inginious/frontend/pages/api/_api_page.py
+++ b/inginious/frontend/pages/api/_api_page.py
@@ -7,8 +7,7 @@
 
 import json
 import flask
-from flask import Response
-from flask_httpauth import HTTPTokenAuth
+from flask import Response,request
 
 import inginious.common.custom_yaml as yaml
 from inginious.frontend.pages.utils import INGIniousPage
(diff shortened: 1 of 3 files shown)


inginious/frontend/pages/api/_api_page.py
```diff
@@ -101,17 +101,20 @@ class APIAuthenticatedPage(APIPage):
...
         return APIPage._handle_api(self, (lambda *args, **kwargs: self._verify_authentication(handler, args, kwargs)), handler_args, handler_kwargs)
 
     def _verify_authentication(self, handler, args, kwargs):
-        """ Verify that the user is authenticated """
-        if not self.user_manager.session_logged_in():
+        """ Verify that request have a http authorization bearer and that token match with a user."""
+        if 'Authorization' in request.headers:
+            auth_header = request.headers.get('Authorization')
+            token = auth_header.replace('Bearer ', '')
+            if self.user_manager.get_userinfo_from_apikey(token) is None:
```
anthonygego, line 112: I'm wondering, at this stage of implementation, if it wouldn't be more interesting to use the Basic Authorization as it is done with WebDAV. Indeed, as the API key implementation is currently very straightforward, asking for the username would add a layer of complexity. Besides, this would be less confusing as the Bearer method is designed for tokens that are very more complex, encrypted, objects.

@nrybowski any thoughts on this ?

inginious/frontend/user_manager.py
```diff
@@ -402,6 +402,20 @@ def get_user_api_key(self, username, create=True):
             apikey = retval.get("apikey", None)
         return apikey
 
+    def get_userinfo_from_apikey(self,apikey):
+        """
+        Get the user info from a given API key.
+        :param: the api key
+        :return: the user based on the API key, None if the key match no one.
+        """
+        if apikey is None:
```
anthonygego, line 411: Empty string seems more probable than `None`, according to the calling code.

anthonygego:
I'll close this PR as it simply does not work :
- It does not ask for authorization (No `WWW-Authenticate` header with 401 response code is sent to the browser)
- It relies on the main session mechanism : therefore it does not work if the user is not logged in using the usual login prompt. This is not the expected behaviour when using API

Besides, the API session is therefore cookieless (ideally a real token in the future) and should not be the same as for the main platform. We need either to use the same (similar) cookieless mechanism as for LTI session or to rewrite the API so that it does not rely on the Flask session object, or, maybe simpler, look how to export the API as a blueprint with separated sessions (I think Flask support that kind of thing).

PR #960, opened 2023-06-23
Move Logging initialization before get_app call

### LINKED ISSUE #958: Can't see frontend logging entries in log file
**Describe the bug**
When webapp starts, I cannot see the frontend logging entries, such as "inginious.frontend - INFO - Starting a simple arch (backend, docker-agent and mcq-agent) locally", in the log file.

- Version: v0.8.7

Looking at the code, it seems that the cause of this issue is that in inginious-wepapp.py, the call to init_logging is made **after** the call to get_app (which relies on an initiated logger). I have tried moving the call to init_logging before the call to get_app, and it seems to have solved the issue.

Drumor:
fix #958

### DIFF
--- a/inginious-webapp
+++ b/inginious-webapp
@@ -58,11 +58,12 @@ if not configfile:
 
 # Load configuration and application (!!! For mod_wsgi, application identifier must be present)
 config = load_json_or_yaml(configfile)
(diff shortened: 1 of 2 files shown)


anthonygego:
`inginious-webdav`is also concerned. This also need to be backported to `inginious-0.8`.

PR #961, opened 2023-06-26
Automate backport workflow

Drumor:
Based on https://github.com/tibdex/backport

By labeling PRs that need to be backported, This should automate the process.
For our needs, we have to define a `backport inginious-0.X` label and set this label on PRs that needs to be backported.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,0 +1,26 @@
+name: Backport
+on:
+  pull_request_target:
+    types:
+      - closed
+      - labeled
+
+jobs:
+  backport:
+    name: Backport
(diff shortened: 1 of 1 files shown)


anthonygego:
I still think it's more convenient to do the backports manually for the moment, for several reasons : 
- Bad merges that need to be reverted happen
- Sometimes master and release branches are too divergent and backporting need manual conflict resolution.
- Fixes aren't always pushed via a PR: if the process is different things may be forgotten.
- We may forget to publish the associated minor release.
- Sometimes we realize a posteriori that some features from the master branch can be backported to the release branch. It is simpler to apply a selection of commits in the same order.

PR #964, opened 2023-08-09
[task_dispensers] Task settings migration

anthonygego:
This PR ends the long awaited process of task settings migration to task dispensers, first step to ease task sharing among courses.

As discussed with @nrybowski, the online but manual way was preferred to trigger the task file update. Therefore, a migration option is added in top of the task list page in case the current task dispenser detects legacy settings in the task files. When the upgrade is asked, the current task dispenser will import the task settings into its own data structure and then the fields imported will be removed from task file descriptors.

This is implemented by adding two function definitions to task dispensers : `has_legacy_tasks` and `import_legacy_tasks` as well as a `legacy_fields` attribute to workaround the fact task dispensers have no access to the descriptors to modify the files themselves. A default newly defined task dispenser will inherit these definitions but they will have no effect on the tasks. They are purposely not defined as abstract methods then.

PR #965, opened 2023-08-09
[task_dispensers] grouped actions on tasks

anthonygego:
This PR follows #964 and proposes a first implementation of grouped actions on tasks in TOC-based dispensers. This had become a must-have in order to separate course from task sets and move the course data in database only, making the task dispenser settings not modifiable from the files.

This implementation is HTML/JS based only, quite heavy and probably not the cleanest possible (it produces a lot of js code). However, I think it's worth to evaluate the user experience first before refactoring and making this definitive.

The implementation is based on the following requirements:
- Each task has multiple settings (quite a lot, actually)
- Each action will act on one setting only, otherwise this is not really useful, all selected tasks would have the same set of settings.
- The current setting values should be quickly available, as there is no need for grouped actions if we must click on each task to know what the setting is. 
- That is, it should be a way to perform all the actions without even opening the task specific modal.

and resolves this by:

1.  Adding a quick setting preview in the task list, using a filter (one setting at a time is shown)
1. Refreshing the quick preview each time a field is updated in the task edit modal. This is the dirtiest part of the code right now.
1. Allowing to select sections/tasks
1. Adding a grouped setting modal that show only the edit field selected in the filter. That is, if filtered on accessibility, only the accessibility can be changed in the modal.
1. Refreshing the tasks specific input fields each time the grouped action fields are changed, and trigger the change event of each of these fields, so that point 2 will also work when settings are edited through grouped actions.
1. Allowing to move and remove a bunch of tasks
1. Allowing to set the task settings for new tasks, because this is now independent.
1. Enabling, by default, a compact view of the task list, in order to apply the filters and save the changes more easily while navigating through a long list of tasks.

This whole UI, while still not being the optimal solution, is slowly becoming an implementation nightmare. It's probably the first signs that jQuery is not sufficient anymore as a Javascript frontend framework.

Another issue: this implementation is still not smartphone friendly. However, the main issue is the drag and drop that prevents appropriate usage of the rest of the UI. I think that either toggling on and off the drag and drop, or completely removing it, might be a first solution. But this would be for a future PR...


anthonygego:
Obsolete.

PR #975, opened 2023-10-12
[frontend/themes] new black theme + codemirror themes

Ananas120:
I hope this time the commits will be in the correct order !

I have replaced the `theme=` by `theme:` instead of using the "-" symbol as suggested, as some codemirror themes contain a "-" in theirname, which may raise issue when splitting


nrybowski:
@Ananas120 can we close https://github.com/UCL-INGI/INGInious/pull/931 since the current PR is a rebase of the previous one ?

anthonygego:
I'll close this PR to focus on #991.

PR #976, opened 2023-10-19
[Frontend] Adding line offset for code problems

AlexandreDoneux:
This PR adds the possibility through the task edit form to have a line offset for the code blocks in code problems. Adding a field to the form, checking it's value and applying the changes to the particular problem. An error will be displayed if the user inputs something other than a number.

### DIFF
--- a/inginious/common/tasks_problems.py
+++ b/inginious/common/tasks_problems.py
@@ -187,6 +187,7 @@ class CodeProblem(Problem):
                 offset = int(problem_content["offset"])
                 if offset < 1:
                     raise Exception("Line offset must be positive!")
+                problem_content["offset"] = offset
             except ValueError:
                 raise Exception("Line offset must be an integer!")
         return Problem.parse_problem(problem_content)
--- a/inginious/common/tasks_problems.py
+++ b/inginious/common/tasks_problems.py
@@ -182,7 +182,9 @@ class CodeProblem(Problem):
     @classmethod
     def parse_problem(self, problem_content):
         # Checking problem edit inputs
-        if len(problem_content["offset"]) != 0:
+        if len(problem_content["offset"]) == 0:
+            problem_content["offset"] = 1 
+        else:
             try:
                 offset = int(problem_content["offset"])
                 if offset < 1:
@@ -190,6 +192,7 @@ class CodeProblem(Problem):
                 problem_content["offset"] = offset
             except ValueError:
                 raise Exception("Line offset must be an integer!")
+
         return Problem.parse_problem(problem_content)
 
     @classmethod
--- a/inginious/common/tasks_problems.py
+++ b/inginious/common/tasks_problems.py
@@ -183,7 +183,7 @@ class CodeProblem(Problem):
     def parse_problem(self, problem_content):
         # Checking problem edit inputs
         if len(problem_content["offset"]) == 0:
-            problem_content["offset"] = 1 
+            del problem_content["offset"]
         else:
             try:
                 offset = int(problem_content["offset"])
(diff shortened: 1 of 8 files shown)


inginious/frontend/static/js/common.js
```diff
@@ -10,7 +10,7 @@ function init_common()
     colorizeStaticCode();
     $('.code-editor').each(function(index, elem)
     {
-        registerCodeEditor(elem, $(elem).attr('data-x-language'), $(elem).attr('data-x-lines'));
+        registerCodeEditor(elem, $(elem).attr('data-x-language'), $(elem).attr('data-x-lines'), $(elem).attr('data-x-first-line'));
```
anthonygego, line 13: Maybe you can try this instead of the `parseInt` + `isNaN` check below.

```suggestion
        registerCodeEditor(elem, $(elem).attr('data-x-language'), $(elem).attr('data-x-lines'), $(elem).attr('data-x-first-line') || 1);
```

anthonygego (changes requested):

inginious/common/tasks_problems.py
```diff
@@ -181,6 +181,14 @@ def input_is_consistent(self, task_input, default_allowed_extension, default_max
...
     def parse_problem(self, problem_content):
+        # Checking problem edit inputs
+        if len(problem_content["offset"]) != 0:
+            try:
+                offset = int(problem_content["offset"])
+                if offset < 1:
+                    raise Exception("Line offset must be positive!")
+            except ValueError:
+                raise Exception("Line offset must be an integer!")
         return Problem.parse_problem(problem_content)
```
anthonygego, line 192: Beware here, because you return the result of the parent class method only, so if you want to store `offset` as a integer in the task descriptor, this won't do the job.

inginious/frontend/pages/course_admin/task_edit.py
```diff
@@ -142,6 +142,8 @@ def POST_AUTH(self, courseid, taskid):  # pylint: disable=arguments-differ
 
             # Network grading
             data["network_grading"] = "network_grading" in data
+
```
anthonygego, line 145: This is a bit pedantic but to keep PRs the cleanest and easiest to review, please avoid to commit files with no useful changes.

I strongly recommend you to give a first review of your PR by yourself before submitting it for review. This is also useful to when debugging lines remain: there is no chance the PR will be merged in those cases so that automatically requires a second review.

inginious/common/tasks_problems.py
```diff
@@ -181,6 +181,15 @@ def input_is_consistent(self, task_input, default_allowed_extension, default_max
 
     @classmethod
     def parse_problem(self, problem_content):
+        # Checking problem edit inputs
+        if len(problem_content["offset"]) != 0:
```
anthonygego, line 185: Here you will skip the check if `problem_content["offset"]` is an empty string, that is, if the field is left empty or as is. Empty string will override the default `1` value when loaded  in the `CodeProblem` constructor.

This will lead to an empty string being passed through CodeMirror and a first line beginning at `0`... you should either save the default value, or remove the empty string from the settings dict in this function.

anthonygego:
Can you rebase this PR so that only your commits appear here ? Otherwise it's pretty hard to review.

AlexandreDoneux:
Rebase on `2e692aa7`.

inginious/common/tasks_problems.py
```diff
@@ -181,6 +181,18 @@ def input_is_consistent(self, task_input, default_allowed_extension, default_max
 
     @classmethod
     def parse_problem(self, problem_content):
+        # Checking problem edit inputs
+        if len(problem_content["offset"]) == 0:
+            problem_content["offset"] = 1
```
anthonygego, line 186:
```suggestion
            del problem_content["offset"]
```

This ease YAML file reading and saves some space.

anthonygego (approved):

Issue #358, opened 2018-09-30
Weak password hashing / salting


bartwojcik:
```
passwd_hash = hashlib.sha512(data["passwd"].encode("utf-8")).hexdigest()
```
How about using something like bcrypt, scrypt, Argon2 or PBKDF2 for user password hashing? Any of these should provide an orders of magnitude better security in case of database leak.

GuillaumeDerval:
> Maybe we can simply store the hash method in the user document (as this is not a secret) to automatically check if the password is OK with the appropriate hash function and update it at the same moment to the new one if it is not the current one used.

A simple method is to prepend the method to the hash. Something like "bcrypt-thehash" or "sha512-thehash". We don't even need to update the database: we can simply say that if it doesn't begin by "bcrypt-", it should be updated (not risk of collision as - is not an hexadecimal character).

 > We should also think of an automated procedure in case of a known leak. Users should at least be informed to reset their passwords. Ideally we can set the passwords as compromised and require the password reset before accepting the credentials.

This should be another issue, but I agree this is an issue we should tackle.

AlexandreDoneux:
We could use Argon2id as it is suggested by OWASP [here](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html). We could at first keep the old algorithm to ease the transition. Both algorithms could be available for authentication but users would be encouraged to change their password with the new algorithm. 

> A simple method is to prepend the method to the hash.

Argon2 hashes contain the different parameters used, including the variant. We can easily detect if a hash uses the new algorithm thanks to that. For example : `'$argon2id$v=19$m=65536,t=3,p=4$MIIRqgvgQbgj220jfp0MPA$YfwJSVjtjSU0zzV/P3S9nnQ/USre2wvJMjfCIjrTQbg'`

anthonygego:
> We can easily detect if a hash uses the new algorithm thanks to that.

We may have to change the algorithm again in the future. Prepending takes this case into account.

PR #978, opened 2023-10-24
[frontend] Adding the use of Argon2id for password hashing

AlexandreDoneux:
Related to issue #358 

This pull request contains changes regarding multiple elements : 
- Adding a new hashing function hash_password_argon2id and renaming the old one to hash_password_sha512.
- Supporting user authentication regardless of the algorithm hashing their password in the database. 
- Using the argon2 hash for new users (via self registering, administrator registering or during the installation process).
- Using the argon2 hash when a users resets it's password or changes it

Password hashes with argon2id are now stored in the database using a prepend "argon2id-" to allow future changes.

### DIFF
--- a/inginious/frontend/installer.py
+++ b/inginious/frontend/installer.py
@@ -566,7 +566,7 @@ class Installer:
         database.users.insert_one({"username": username,
                                    "realname": realname,
                                    "email": email,
-                                   "password": UserManager.hash_password_sha512(password),
+                                   "password": "argon2id-" + UserManager.hash_password_argon2id(password),
                                    "bindings": {},
                                    "language": "en"})
 
--- a/inginious/frontend/installer.py
+++ b/inginious/frontend/installer.py
@@ -566,7 +566,7 @@ class Installer:
         database.users.insert_one({"username": username,
                                    "realname": realname,
                                    "email": email,
-                                   "password": UserManager.hash_password(password),
+                                   "password": UserManager.hash_password_sha512(password),
                                    "bindings": {},
                                    "language": "en"})
 
--- a/inginious/frontend/installer.py
+++ b/inginious/frontend/installer.py
@@ -566,7 +566,7 @@ class Installer:
         database.users.insert_one({"username": username,
                                    "realname": realname,
                                    "email": email,
-                                   "password": "argon2id-" + UserManager.hash_password_argon2id(password),
+                                   "password": UserManager.hash_password(password),
                                    "bindings": {},
                                    "language": "en"})
 
--- a/inginious/frontend/pages/preferences/profile.py
+++ b/inginious/frontend/pages/preferences/profile.py
@@ -15,6 +15,7 @@ from argon2.exceptions import VerifyMismatchError
 from inginious.frontend.pages.utils import INGIniousAuthPage
 from inginious.frontend.user_manager import UserManager
 
+
 class ProfilePage(INGIniousAuthPage):
     """ Profile page for DB-authenticated users"""
 
@@ -55,23 +56,18 @@ class ProfilePage(INGIniousAuthPage):
             msg = _("Passwords don't match !")
             return result, msg, error
         elif self.app.allow_registration and len(data["passwd"]) >= 6:
-            ph = PasswordHasher()
-            match = {"username": self.user_manager.session_username()}
 
             if "password" in userdata:
-                user = self.database.users.find_one(match)
-                oldpasswd_hash_sha512 = UserManager.hash_password_sha512(data["oldpasswd"])
-                try:
-                    if oldpasswd_hash_sha512 == user["password"] or ph.verify(user["password"][9:], data["oldpasswd"]):
-                        passwd_hash = UserManager.hash_password(data["passwd"])
-
-                except VerifyMismatchError:
+                user = self.user_manager.auth_user(self.user_manager.session_username(), data["oldpasswd"], False)
+                if user is None:
                     error = True
                     msg = _("Incorrect old password.")
                     return result, msg, error
-
-            result = self.database.users.find_one_and_update(match,{"$set": {"password": passwd_hash}},
-                                                             return_document=ReturnDocument.AFTER)
+                else:
+                    passwd_hash = UserManager.hash_password(data["passwd"])
+                    result = self.database.users.find_one_and_update({"username": self.user_manager.session_username()},
+                                                                     {"$set": {"password": passwd_hash}},
+                                                                     return_document=ReturnDocument.AFTER)
 
         # Check if updating language
         if data["language"] != userdata["language"]:
--- a/inginious/frontend/pages/preferences/profile.py
+++ b/inginious/frontend/pages/preferences/profile.py
@@ -54,8 +54,8 @@ class ProfilePage(INGIniousAuthPage):
             msg = _("Passwords don't match !")
             return result, msg, error
         elif self.app.allow_registration and len(data["passwd"]) >= 6:
-            oldpasswd_hash = UserManager.hash_password(data["oldpasswd"])
-            passwd_hash = UserManager.hash_password(data["passwd"])
+            oldpasswd_hash = UserManager.hash_password_sha512(data["oldpasswd"])
+            passwd_hash = UserManager.hash_password_sha512(data["passwd"])
 
             match = {"username": self.user_manager.session_username()}
             if "password" in userdata:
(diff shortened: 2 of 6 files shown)


inginious/frontend/pages/preferences/profile.py
```diff
@@ -54,20 +55,23 @@ def save_profile(self, userdata, data):
...
-
+            ph = PasswordHasher()
             match = {"username": self.user_manager.session_username()}
+
             if "password" in userdata:
-                match["password"] = oldpasswd_hash
+                user = self.database.users.find_one(match)
+                oldpasswd_hash_sha512 = UserManager.hash_password_sha512(data["oldpasswd"])
+                try:
+                    if oldpasswd_hash_sha512 == user["password"] or ph.verify(user["password"][9:], data["oldpasswd"]):
```
anthonygego, line 65: Here you support retrocompatibility by just testing both algorithms. This may be a bit dangerous in case one incorrect password with the incorrect algorithm would match the stored hash (in this particular case that won't happen, but this remains something possible with same-length same-alphabet hashes)

I strongly recommend that we extract the appropriate method from what is prepended. In this case, you will just need to fetch the `methods[prefix]` hashing function and perform the check. If the verifying process is different for argon2id, then the dictionary can point to classes that implement both hashing and verifying methods.

inginious/frontend/user_manager.py
```diff
@@ -291,16 +293,22 @@ def auth_user(self, username, password):
...
         user = self._database.users.find_one(
-            {"username": username, "password": password_hash, "activate": {"$exists": False}})
+            {"username": username, "activate": {"$exists": False}})
+        ph = PasswordHasher()
+
+        try:
+            if user is None:
+                return None
+
+            elif self.hash_password_sha512(password) == user["password"] or ph.verify(user["password"][9:], password):
```
anthonygego, line 306: Same comment as above

anthonygego (changes requested):

inginious/frontend/installer.py
```diff
@@ -566,7 +566,7 @@ def configure_authentication(self, database):
         database.users.insert_one({"username": username,
                                    "realname": realname,
                                    "email": email,
-                                   "password": UserManager.hash_password(password),
+                                   "password": "argon2id-" + UserManager.hash_password_argon2id(password),
```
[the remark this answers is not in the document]
AlexandreDoneux: Created a new function to encapsulate the other password hashing functions. Used a unique variable in there as you suggested.

inginious/frontend/pages/preferences/profile.py
AlexandreDoneux: Modified UserManager.auth_user() to do so. 

Modified auth_user() method to allow it's use during password change, reducing duplicate code.

inginious/frontend/user_manager.py
```diff
@@ -286,21 +288,37 @@ def get_auth_methods(self):
...
+            {"username": username, "activate": {"$exists": False}})
+
+        def connect_user(username, user, do_connect):
+            return (user and self.connect_user(username, user["realname"], user["email"], user["language"],
+                                               user.get("tos_accepted", False)) if do_connect else user)
+
+        if user is None:
+            return None
+
+        if user["password"][:9] == "argon2id-":
```
anthonygego, line 309: You should extract the method so that this string should only be located in one dict and the code remains agnostic.

Adding a new hashing method should only consists in adding the method and appropriate `hash`and `verify` functions, and changing the current method.

inginious/frontend/user_manager.py
```diff
@@ -1034,9 +1052,31 @@ def generate_api_key(cls):
...
+    def hash_password_argon2id(cls, content):
+        """
+        :param content: a str input
+        :return a hash of str input
+        """
+        ph = PasswordHasher()
+        return "argon2id-" + ph.hash(content)
+
+    @classmethod
+    def hash_password(cls, content):
```
anthonygego, line 1072: A potential implementation could be `hash_password(cls, content)`  that indeed use the latest method, with an additional `verify_hash(cls, db_hash, password, method="sha512")` that takes the method

anthonygego (changes requested):
Could you please also rebase this PR so that it has no more conflicts ?

AlexandreDoneux:
Rebase on `2e692aa7`.

inginious/frontend/user_manager.py
```diff
@@ -286,21 +288,56 @@ def get_auth_methods(self):
...
+            {"username": username, "activate": {"$exists": False}})
+
+        def connect_user(username, user, do_connect):
+            return (user and self.connect_user(username, user["realname"], user["email"], user["language"],
+                                               user.get("tos_accepted", False)) if do_connect else user)
+
+        if user is None:
+            return None
+
+        method, db_hash = user["password"].split("-") if "-" in user["password"] else ("sha512", user["password"])
```
anthonygego, line 309:
```suggestion
        method, db_hash = user["password"].split("-", 1) if "-" in user["password"] else ("sha512", user["password"])
```

We should suppose that `-` may be part of the hash since we want to remain method-agnostic. No further check is needed as sha512 should not contain any `-`.

inginious/frontend/user_manager.py
```diff
@@ -286,21 +288,56 @@ def get_auth_methods(self):
...
+        :param do_connect: indicates if the user must be connected after authentification, True by default
+        :return: Returns a dict representing the user, or None if the authentication was not successful
         """
-        password_hash = self.hash_password(password)
-
         user = self._database.users.find_one(
-            {"username": username, "password": password_hash, "activate": {"$exists": False}})
+            {"username": username, "activate": {"$exists": False}})
+
+        def connect_user(username, user, do_connect):
```
anthonygego, line 302: Wrapping this is probably not needed :thinking:

inginious/frontend/user_manager.py
```diff
@@ -286,21 +288,56 @@ def get_auth_methods(self):
...
+            return (user and self.connect_user(username, user["realname"], user["email"], user["language"],
+                                               user.get("tos_accepted", False)) if do_connect else user)
+
+        if user is None:
+            return None
+
+        method, db_hash = user["password"].split("-") if "-" in user["password"] else ("sha512", user["password"])
+
+        if self.verify_hash(db_hash, password, method):
+            return connect_user(username, user, do_connect)
```
anthonygego, line 312:
```suggestion
            if do_connect:
                self.connect_user(username, user["realname"], user["email"], user["language"], user.get("tos_accepted", False))
            return user
            
```

inginious/frontend/pages/preferences/profile.py
```diff
@@ -54,20 +56,18 @@ def save_profile(self, userdata, data):
...
-                match["password"] = oldpasswd_hash
-
-            result = self.database.users.find_one_and_update(match,
-                                                             {"$set": {"password": passwd_hash}},
-                                                             return_document=ReturnDocument.AFTER)
-            if not result:
-                error = True
-                msg = _("Incorrect old password.")
-                return result, msg, error
+                user = self.user_manager.auth_user(self.user_manager.session_username(), data["oldpasswd"], False)
```
anthonygego, line 61: I'm a little bit confused with prior code already, but can you confirm that a user with no password (registered through social networks for instance) can actually set a password to use password authentication ?

AlexandreDoneux: They can now. I changed the structure for handling user password change in this PR. Tell me if your ok with it.

inginious/frontend/user_manager.py
```diff
@@ -1034,9 +1070,31 @@ def generate_api_key(cls):
...
         return hashlib.sha512(content.encode("utf-8")).hexdigest()
+
+    @classmethod
+    def hash_password_argon2id(cls, content):
+        """
+        :param content: a str input
+        :return a hash of str input
+        """
+        ph = PasswordHasher()
+        return "argon2id-" + ph.hash(content)
```
anthonygego, line 1087:
```suggestion
        return ph.hash(content)
```

inginious/frontend/user_manager.py
```diff
@@ -1034,9 +1070,31 @@ def generate_api_key(cls):
...
+        """
+        Encapsulates the other password hashing functions
+        :param content: a str input
+        :return a hash of str input
+        """
+
+        methods = {"argon2id": cls.hash_password_argon2id, "sha512": cls.hash_password_sha512}
+        latest_method = "argon2id"
+
+        return methods[latest_method](content)
```
anthonygego, line 1100:
```suggestion
        return latest_method + "-" + methods[latest_method](content)
```

anthonygego (changes requested):
LGTM. I'd recommend moving the `latest_method` prepending into the `hash_password` method as suggested, to keep the same behaviour for sha512 when saved, and to avoid repeating the identifier in the hashing method itself.

I let you test and confirm this is OK for you and then we'll be able to merge.

AlexandreDoneux:
> I'd recommend moving the `latest_method` prepending into the `hash_password` method as suggested

Done

PR #982, opened 2023-11-09
Grouped actions and tasksets

anthonygego:
This PR introduces grouped actions and tasksets under courses. Grouped actions concern the task dispenser admin view so that all the task parameters

- Grouped actions: 
  1. During edition, task settings are stored in a Javascript dictionary instead of printing as many modals as there are tasks in the HTML. Each time the modal is opened, a javascript function for each setting is called and fetches the taskid from a `data-` tag attribute and populates the modal fields based on the settings dictionary.
  2. As grouped actions would not have sense if modals have to be opened to see the current setting for each task, each setting can be shown below the task name, using a filter which is also used to know what grouped action to perform.
  3. A grouped action modal contains all the settings but only the one selected in the filter is shown.
  4. Whenever a field is modified, the settings dictionary is updated for the selected task(s).
  5. Moving and deleting tasks has also been added as grouped actions.

- Tasksets : as illustrated in the Figure below, courses have been divided in two separate objects where the remaining course is a task dispensing object located in database, while the tasksets actually manage the tasks themselves located in the filesystem.
Capture d’écran du 2023-11-09 16-08-26

    To ease the implementation, the course factory has been duplicated. A taskset factory keep the YAML file loading part and the new course factory is a mockup above the database to avoid modifying too much code at a time. 
    The latter also imports older courses by checking for a `course.yaml` file and if some entry already exists in database for the course id associated to the folder name.
    Several pages have been added to manage the taskset where the task dispenser is also used to provide course templating at the time of instantiation.
   The task dispenser is no more able to manage task files as courses only have read access to the tasksets.

**TODO:**

- [x]  Check for task categories in submission view
- [x] Fix task order imported from the older TOC dispenser structure `{"taskid2":1, "taskid1":0} -> ["taskid1", "taskid2"]`


nrybowski:
LGTM.

The [issue mentioned previously](https://github.com/UCL-INGI/INGInious/pull/964#issuecomment-1757570306) is resolved .

nrybowski (approved):

PR #985, opened 2023-11-23
[frontend] Changing accessibility and registration data structure

AlexandreDoneux:
This pull request has as goal the change of accessibility and registration data structure in in the code as well as for the storage in DB and yaml files. The objective is also to store the dates as ISO dates in the database and yaml files. This concerns course accessibility, course registration period and task accessibility.

These accessibilities where previously a single value (accessible/registration for courses and accessibility for tasks) and were structured as follows : 
* a true when always accessible
* a false when never accessible
* a concatenated string containing all the dates separated by a "/" 
   (ex: `"2023-12-06 09:00:00/2023-12-13 09:00:00/2023-12-13 09:00:00"`)

We will now only use the dates and use minimal and maximal dates to define the cases where the the courses and tasks are always or never accessible. Those minimal and maximal dates are also used in a custom accessibility when certain dates are not defined. You can for example only define a start date in the form and INGInious will then assume there is no end date to the task and it will always be accessible after the given start date.

TO DO : 
- [x] Change datastructure in AccessibleTime class and it's usage 
   - [X] adapt AccessibleTime 
   - [X] adapt usage basic INGInious app 
   - [x] adapt usage in contest plugin
   - [x] Change AccessibleTime default parameters and it's use 
- [X] Changing init structure 
   - [X] courses
   - [X] tasks 
- [x] Adapting forms to new structure
   - [X] Transforming datetime objects into strings 
   - [X] course settings
   - [X] task accessibility edit
   - [x] task accessibility grouped edit
- [X] Dates as ISODate() objects in MongoDB
   - [X] Storing (transform in datetime objects before storing in DB)
   - [X] Retrieving 
   - [X] Correctly handling min and max dates
- [X] Storing dates as timestamps objects in yaml files
   - [X] Storing 
   - [x] Retrieving
   - [x] Correctly handling min and max dates
- [x] Automatically change structure in DB and yaml files when switching version
   - [x] DB
      - ~~[ ] In already existing collections~~ => not needed
      - [x] During "courses" collection init in legacy courses import
      - [x] During legacy task import in task list edit
   - [x] YAML files
      - [x] During legacy task import in course template


AlexandreDoneux:
Rebase on cf8141c0ad1dace81ea936a07de80d4358e7314b.

AlexandreDoneux:
Rebase on  1095603.

AlexandreDoneux:
Rebase on 4de3a172.

inginious/frontend/task_dispensers/combinatory_test.py
```diff
@@ -49,10 +55,10 @@ def render_edit(self, template_helper, element, task_data, task_errors):
 
         taskset = element if isinstance(element, inginious.frontend.tasksets.Taskset) else None
         course = element if isinstance(element, inginious.frontend.courses.Course) else None
+        task_config = dict_data_datetimes_to_str(self._task_config)
```
anthonygego, line 58: I'm wondering if this could not rather be implemented as a Jinja filter rather a recursive function here.

inginious/frontend/task_dispensers/toc.py
```diff
@@ -115,9 +116,10 @@ def render_edit(self, template_helper, element, task_data, task_errors):
 
         taskset = element if isinstance(element, inginious.frontend.tasksets.Taskset) else None
         course = element if isinstance(element, inginious.frontend.courses.Course) else None
+        task_config = dict_data_datetimes_to_str(self._task_config)
```
anthonygego, line 119: Same comment as above.

inginious/frontend/task_dispensers/util.py
```diff
@@ -153,7 +155,7 @@ def get_value(cls, task_config):
 
 
 class SubmissionLimit(TaskConfigItem):
-    default = {"amount": -1, "period": -1}
+    default = dict({"amount": -1, "period": -1})
```
anthonygego, line 158: Is this necessary ? If so, why ?

inginious/frontend/task_dispensers/util.py
```diff
@@ -178,7 +180,7 @@ def get_value(cls, task_config):
 
 
 class Accessibility(TaskConfigItem):
-    default = False
+    default = dict({"start": datetime.min, "soft_end": datetime.max, "end": datetime.max})
```
anthonygego, line 183: Same, is `dict()` necessary here ?

inginious/frontend/course_factory.py
```diff
@@ -113,6 +114,12 @@ def _migrate_legacy_courses(self):
                     cleaned_taskset_descriptor["dispenser_data"] = taskset_descriptor.get("dispenser_data", {})
                 taskset_descriptor["tasksetid"] = courseid
                 taskset_descriptor["admins"] = taskset_descriptor.get("admins", []) + taskset_descriptor.get("tutors", [])
+                if "accessible" in taskset_descriptor:
+                    taskset_descriptor["accessible"] = change_access_structure(taskset_descriptor["accessible"])
+                if "registration" in taskset_descriptor:
+                    taskset_descriptor["registration"] = change_access_structure(taskset_descriptor["registration"])
+
+                taskset_descriptor = dict_data_str_to_datetimes(taskset_descriptor)
```
anthonygego, line 122: Same, if we only need `dict_data_str_to_datetimes` here, you can move it here to ease later deprecation.

inginious/frontend/templates/task_dispensers_admin/config_items/accessibility.html
```diff
@@ -83,21 +83,33 @@
...
             return;
         var accessibility = dispenser_config[taskid]["accessibility"];
         var value;
 
-        if (accessibility === true) {
-            value = 'true';
-        } else if (accessibility === false) {
-            value = 'false';
-        } else {
+        special_dates = ["1-01-01 00:00:00", "9999-12-31 23:59:59"]
```
anthonygego, line 92: We should print `datetime.min` or `datetime.max` (without the microseconds) instead, to ensure consistency.

inginious/frontend/templates/task_dispensers_admin/config_items/accessibility.html
```diff
@@ -114,62 +126,68 @@
             $("#task_" + taskid + " .accessibility").hide();
             if($(this).val() == "true") {
                 $("#task_" + taskid + " .accessibility-always").show();
-                dispenser_config[taskid]["accessibility"] = true;
+                dispenser_config[taskid]["accessibility"]["start"] = "1-01-01 00:00:00";
```
anthonygego, line 129: ...and replace all those hardcoded values by a clean reference.

inginious/common/custom_yaml.py
```diff
@@ -74,10 +76,18 @@ def _long_str_representer(dumper, data):
     def _default_representer(dumper, data):
         return _long_str_representer(dumper, str(data))
 
+    def _timestamp_representer(dumper, data):
+        formatted_date = data.strftime("%Y-%m-%dT%H:%M:%S.%fZ") if data != "" else None
+        # Ensure the year, month and day are always correctly formatted
+        formatted_date = formatted_date.replace(data.strftime("%Y"), "{:04d}".format(data.year), 1)
+        return dumper.represent_scalar('tag:yaml.org,2002:timestamp', formatted_date)
```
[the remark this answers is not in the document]
AlexandreDoneux: A None value for formatted_date generates an error on the web page : "NoneType isnot subscriptable".
Removing this check because _timestamp_representer() is only applied to datetime objects.

inginious/frontend/task_dispensers/util.py
AlexandreDoneux: Not necessary at all, removing it.

inginious/frontend/task_dispensers/combinatory_test.py
AlexandreDoneux: The task_config dict is transformed into a json on the frontend by the jinja filter "tojson". Datetime objects being not serializable we have to transform those datetimes into strings before this tojson transformation. However it doesn't seem there are appropriate jinja filters to do this. Not in a efficient way at least.

inginious/frontend/course_factory.py
AlexandreDoneux: It is used in two other places to transform the task accessibilities back into datetimes (during task accessibilities change in course and taskset template).

There maybe is a better place to place these functions nonetheless.

AlexandreDoneux:
Some code might need to be rewritten in settings.py : 

https://github.com/UCL-INGI/INGInious/blob/2306071d8b71701cfa2187a9afd03067558c60ec/inginious/frontend/pages/course_admin/settings.py#L24-L51

The data of the course is accessed through the course_factory : 
https://github.com/UCL-INGI/INGInious/blob/2306071d8b71701cfa2187a9afd03067558c60ec/inginious/frontend/pages/course_admin/settings.py#L32


However, the course is retrieved at the start of the method. We could rewrite a bit to use the course instead of ```self.course_factory.get_course_descriptor_content(courseid)```.

There is also the use of a protected member to use the min and max dates instead of `datetime.min` and `datetime.max.replace(microsecond=0)`. A method on course retrieving the content would allow us acces to those values.

inginious/frontend/course_factory.py
```diff
@@ -113,6 +114,12 @@ def _migrate_legacy_courses(self):
                     cleaned_taskset_descriptor["dispenser_data"] = taskset_descriptor.get("dispenser_data", {})
                 taskset_descriptor["tasksetid"] = courseid
                 taskset_descriptor["admins"] = taskset_descriptor.get("admins", []) + taskset_descriptor.get("tutors", [])
+                if "accessible" in taskset_descriptor:
+                    taskset_descriptor["accessible"] = change_access_structure(taskset_descriptor["accessible"])
```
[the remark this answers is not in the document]
AlexandreDoneux: It is used before the dispenser update during task settings changes in course and taskset template.

However, if we want to support legacy access structure ("start/soft_end/end") and change the structure only during the accessibility change we don't need `change_access_structure`.

AlexandreDoneux:
Keeping support for the legacy access format and changing the format on saving the settings of a course or task is not as easy as we thought for the moment.

As you suggested we can add support in the `AccessibleTime` __init__. For a course accessibility and registration period it only needs some tweaks. However task accessibility is more difficult as the accessibility of a task for the edit pages (taskset template and course tasks) is not passed through AccessibleTime.

For example in toc.py, task_config is passed to the jinja template and contains the data of the different tasks.
```py
    def render_edit(self, template_helper, element, task_data, task_errors):
        """ Returns the formatted task list edition form """
        config_fields = {
            "closed": SectionConfigItem(_("Closed by default"), "checkbox", False),
            "hidden_if_empty": SectionConfigItem(_("Hidden if empty"),"checkbox",False)
        }

        taskset = element if isinstance(element, inginious.frontend.tasksets.Taskset) else None
        course = element if isinstance(element, inginious.frontend.courses.Course) else None
        task_config = dict_data_datetimes_to_str(self._task_config)

        return template_helper.render("task_dispensers_admin/toc.html", element=element, course=course, taskset=taskset,
                                      dispenser_structure=self._toc, dispenser_config=task_config, tasks=task_data,
                                      task_errors=task_errors, config_fields=config_fields)
```
However, when tracing back from where the data comes from we can see it is retrieved from the database here, in course_factory.py, and is never processed by AccessibleTime : 

```py
    def get_course_descriptor_content(self, courseid):
        return self._database.courses.find_one({"_id": courseid})
```

In

inginious/frontend/course_factory.py
AlexandreDoneux: This occurrence has been removed. However, it is still used in the previously mentioned places.

AlexandreDoneux:
Rebase on b3a40952c36933b4741e61ddac83cfe9e2c1083a.

anthonygego:
Can you try to rebase against `main` ?

PR #987, opened 2023-12-06
Use the job wall clock timeout for time limit estimation

mpiraux:
By default the CPU time is used, which is not a real time metric. Using it to provide user feedback on job completion is not appropriate in multi-core environments.


nrybowski:
Not sure of the behavior of this metric in virtualized environments. It could be worth to perform some comparisons.

mpiraux:
image
image
Context for this PR

anthonygego:
There is no optimal solution to this issue right now. The wall time is better for multithreaded projects because it is always larger than the running time. However, for most monothreaded projects (and the most common exercise type), the value is far away from the reality.

I think you should keep track of this issue in #989 for later in order to provide a better time estimation.

Issue #893, opened 2022-11-25
Replace setup.py by pyproject.toml


nrybowski:
The `setup.py` installation method will be deprecated soon, see https://github.com/pypa/pip/issues/8559.

We should use the new `pyproject.toml` configuration format instead.

nrybowski:
Done in 991ee64387f0ea1c90dd3040e4f8a970a534f435.

PR #994, opened 2024-02-14
Pivot to pyproject.toml for setuptools packaging

AlexandreDoneux:
Using pyproject.toml is the new standardized way to declare packaging metadata.  As it is the new standard there will likely be less problems related to the packaging of INGInious.

_Rem_ : setup.py is still usable and the configuration in pyproject.toml overwrites that of setup.py. For the moment the scripts are left in setup.py because of the condition on INGINIOUS_COMPOSE environment variable that can not be done in toml files.

TO DO : 
- [x] Move  setup.py config to pyproject.toml
- [ ] Change test_suite depreciated parameter
- [ ] Fix inginious-synchronize script (`sh.CommandNotFound: cd`)


nrybowski:
See https://github.com/UCL-INGI/INGInious/issues/893

AlexandreDoneux:
A discussion is needed regarding the scripts in the INGInious package using docker compose. They are not needed and where ignored conditionnaly with a boolean in setup.py. It is however not possible with the pyproject.toml file.

Using two different pyproject.toml files does not seem to be achievable : 
1) The pip install command does not allow selection of a particular pyprojetc.toml.
2) Placing the pyproject.toml files in two sub-directories will not work as it seems pip copies only that directory in /tmp before the install. The source code is therefor not reachable for the install.

So either we accept the presence of unused scripts in the docker compose package (not that many) or we comment the pyproject.toml scripts before building it.

nrybowski:
This PR will be reconsidered after the definition of a new backend architecture and services separation.

PR #996, opened 2024-02-28
[frontend] Removing Walloon traduction

AlexandreDoneux:
This traduction is obviously not used and is not maintained


AlexandreDoneux:
Done with Weblate push #997

Issue #1007, opened 2024-08-20
Empty ID course


thremilien:
**Describe the bug**
You can create a course with an empty id resulting in multiple errors (especially in the url which is based on the course's id)

![image](https://github.com/user-attachments/assets/c356c04a-ddf8-4676-8737-10acc91c642d)
![image](https://github.com/user-attachments/assets/8aba1ae7-d9d0-4b19-8c69-3313ce2b42ed)

nrybowski:
@AlexandreDoneux Could you try to reproduce?

PR #1009, opened 2024-08-20
Fix empty course ID

thremilien:
Add the check for course ID
#1007


AlexandreDoneux:
This looks good to me. We indeed do that check when creating a taskset, tasks, etc.

nrybowski (approved):
LGTM

Issue #1001, opened 2024-04-09
[frontend/task_edit] Subproblems are not shown in specified order.


anthonygego:
**Describe the bug**

The problem data dictionary exported to the template here : 

https://github.com/UCL-INGI/INGInious/blob/6e14ae8edb65b0ce18059863655659ba9efea129/inginious/frontend/pages/course_admin/task_edit.py#L63

is a Python OrderedDict.

The Javascript JSON load method does not load the keys an an ordered dictionary anymore : 
https://github.com/UCL-INGI/INGInious/blob/6e14ae8edb65b0ce18059863655659ba9efea129/inginious/frontend/templates/course_admin/task_edit.html#L170

Therefore, the subproblems are shown in an unexpected order in the task edition interface.

**INGInious installation details**
- Version: v0.8+

**To Reproduce**
1. Create a task with problem ids "01", "02", ..."10".
2. The task editor may show "10" before all the others.

**Expected behavior**
Subproblems should be displayed in the order specified in the YAML file.

anthonygego:
A workaround is to export the dict as a list (using `OrdrerdDict.items()`) and adapt the `studio_load` method for such a data format.

thremilien:
I can't reproduce the bug, plus the project architecture seems to have moved (but the two highlighted lines haven't been modified), perhaps this has been corrected?

nrybowski:
@AlexandreDoneux Could you check whether this issue still relevant?

PR #1011, opened 2024-08-21
Fix order of subproblems

thremilien:
Change subproblem rendering from object to list to fix the order
#1001

### DIFF
--- a/inginious/frontend/static/js/studio.js
+++ b/inginious/frontend/static/js/studio.js
@@ -22,9 +22,8 @@ function refresh_codemirror()
  * Load the studio, creating blocks for existing subproblems
  */
 function studio_load(data)
-{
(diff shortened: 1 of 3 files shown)


inginious/frontend/static/js/studio.js
thremilien, line 1: Regenerate minified js

AlexandreDoneux:
The list of subproblems does not appear as it should be in the tab. I believe the current code gives a json to the `studio_load()` function. You are passing it an array containing another array representing each subproblem.

On another note I think we should also fix this issue in the task list in the taskset settings. It does not appear to be a problem with the task list in the course settings as well as the one displayed for students.

image

anthonygego:
This is probably not relevant anymore according to the comments and `main` branch. Feel free to re-open in the future.

Issue #998, opened 2024-03-14
[backend/submissions] Anonymous submissions download


ogoletti:
In the context of Computing Education Research or for statistics, we are sometimes interested in downloading data about submissions. In most cases, it is necessary to work with anonymized data.

It would be nice to have a button to export/download the submissions without user informations. 
Just replacing the user information with an "user"+id whould be enough imho. 
That is, it's sometimes important to track which submission has been made by a same user.

nrybowski:
There is already [a manual anonymization tool](https://github.com/UCL-INGI/INGInious/blob/master/utils/task_tester/inginious-submission-anonymizer) but it should be directly integrated with the web interface.

Issue #891, opened 2022-11-07
[frontend/submissions] Automatic join of users information in CSV exports


anthonygego:
**Is your feature request related to a problem? Please describe.**
Currently if you download the results of the submissions page query, you only get the username as user identification. A manual join has to be performed with the "Users" CSV to get more insight (realname, email, ...).

**Describe the solution you'd like**
It would be useful to perform the join automatically and serve a CSV file with the user info.

PR #1014, opened 2024-08-22
Better CSV Download

thremilien:
Add modal for csv download allowing to remove some fields from the csv. 
Perform a join with user table.
Add user id generation for anonymization

Correct some missplacement of data placement (when hovering button).

May need redesign.

image

Solved issues : 
#998 
#936 
#891


AlexandreDoneux:
There is no possibility to select which fields we want the data from. Also, not checking the the "submission fields" but checking the "user fields" leads to having only the data of the user for every submission. That is obviously unusable.

I suggest closing this so we can have a PR of our own and to avoid confusion.

Issue #926, opened 2023-03-24
[frontend/account creation] Unable to use special characters in username during account creation


MaximeJdDS:
**Is your feature request related to a problem? Please describe.**
When attempting to create a new account, users are unable to use special characters such as "é" in their username. The error message displayed only states "Invalid username format," which is not a appopriate error message.

**Describe the solution you'd like**
The best for the user is to be allowed to use such special characters but it can works too if a text tell us the rules before writing our username or during the writing.

**Additional context**
The issue occurs on all platforms (web, mobile, etc.).
The issue was tested using multiple browsers and devices.

thremilien:
The fact that we can't include special characters in username is deliberate. The mainn reason is to prevent character encoding issue when we write stuff in files. On the other hand this is a great idea to include text box with the rules, this will be added.

ogoletti:
While a tooltip with authorised characters is a great idea, choosing an alternative error message if possible is also more inclusive and respectful of the user. Telling someone a character from their id (potentially their name) can hurt.

thremilien:
> While a tooltip with authorised characters is a great idea, choosing an alternative error message if possible is also more inclusive and respectful of the user. Telling someone a character from their id (potentially their name) can hurt.

The restriction is only active for username. For real name all characters are available. I just made a PR, tell me if it's ok with you thoughts

ogoletti:
As suggested, I would instead provide a tooltip on the inputfield with the rules. 
BTW rules should be the proper ones... In your PR, you don't say that case is ignored. 
Also, instead of "Invalid username format" you could use "The username is limited to this range of characters: ..."

This answer elaborates a bit more on the topic: https://softwareengineering.stackexchange.com/a/330518

anthonygego:
Closing this, as mentioned in #1015

PR #1015, opened 2024-08-23
Edit message for user creation

thremilien:
Add rules for username in error message
#926


AlexandreDoneux:
Do we need a tooltip as suggested by @ogoletti or is the inscription above the form input enough ? 

If yes we can use a bootstrap tooltip and just add `title="{{ _('...') }}" data-toggle="tooltip" data-placement="right"` inside the input tag.

anthonygego:
*Alphanumeric* is enough explanation. If you don't know what it means, just open a dictionary, ask a research engine or even a chatbot. I'm closing this.

PR #1017, opened 2024-08-23
Issue with file extension in file subproblem

thremilien:
The code isn’t removing extra spaces when splitting the file extensions. So if you input something like ".pdf, .png", it ends up with " .png" (with a space) instead of just ".png". Because of this, files with the ".png" extension won’t be recognized properly.

### DIFF
--- a/inginious/common/tasks_problems.py
+++ b/inginious/common/tasks_problems.py
@@ -236,7 +236,7 @@ class FileProblem(Problem):
             if problem_content["allowed_exts"] == "":
                 del problem_content["allowed_exts"]
             else:
-                problem_content["allowed_exts"] = problem_content["allowed_exts"].split(',')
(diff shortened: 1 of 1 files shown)


inginious/common/tasks_problems.py
```diff
@@ -236,7 +236,7 @@ def parse_problem(self, problem_content):
             if problem_content["allowed_exts"] == "":
                 del problem_content["allowed_exts"]
             else:
-                problem_content["allowed_exts"] = problem_content["allowed_exts"].split(',')
+                problem_content["allowed_exts"] = [extension.strip() for extension in problem_content["allowed_exts"].split(',')]
```
AlexandreDoneux, line 239:
```suggestion
                problem_content["allowed_exts"] = problem_content["allowed_exts"].replace(" ", "").split(',')
```

AlexandreDoneux: Other than that, looks good to me.

Issue #1013, opened 2024-08-21
Conflict with werkzeug version on new installation


vvandenschrieck:
Hello, 

I'm currently deploying a new INGInious server, and run into what I think is a conflict with the werkzeug package version.
  
**INGInious installation details**
- Version: 0.8.7, installed via pip on a Debian 12 and running on Apache/WSGI  

**To Reproduce**
1.  Install INGInious with pip in a venv
2. Configure Apache WSGI following the INGInious doc
3. When displaying the main page, the INGInious app gives an error message.  In the logs, the following exception shows : 

```
Traceback (most recent call last):
  File "/inginious/lib/python3.11/site-packages/flask/app.py", line 904, in finalize_request
    response = self.process_response(response)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/inginious/lib/python3.11/site-packages/flask/app.py", line 1284, in process_response
    self.session_interface.save_session(self, ctx.session, response)
  File "/inginious/lib/python3.11/site-packages/inginious/frontend/flask/mongo_sessions.py", line 143, in save_session
    response.set_cookie(self.get_cookie_name(app), session_id,
  File "/inginious/lib/python3.11/site-packages/werkzeug/sansio/response.py", line 227, in set_cookie
    dump_cookie(
  File "/inginious/lib/python3.11/site-packages/werkzeug/http.py", line 1301, in dump_cookie
    if not _cookie_no_quote_re.fullmatch(value):
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: cannot use a string pattern on a bytes-like object
2024-08-21 20:38:30,215 - inginious.webapp.requests - INFO - 127.0.0.1:46372 - "- GET /" - 500 INTERNAL SERVER ERROR
```
The same exception occurs while running inginious directly with the inginious-webapp executable.  

**Temporary Fix**

I was able to solve this issue by downgrading Werkzeug to version 2.3.7 instead of 3.0.1.  Maybe this is a compatibility issue?

AlexandreDoneux:
Hello, thanks for your issue. This was a problem we were aware of and had fixed  (#972) . Unfortunately it was reverted by mistake.

anthonygego:
This should be fixed in v0.8.9 and `main`.

PR #1025, opened 2024-11-28
V0.8 dependecies fix

AlexandreDoneux:
This PR has as goal to set the package dependencies to exact versions as well as impose the 2.3.7 version of Werkzeug to avoid the reoccurring dependency problem related (see #1013).


AlexandreDoneux:
Also removing the `dataclasses` module and the limitation on `requests` because we are not supporting the project for python versions under 3.10.

PR #1026, opened 2024-12-04
fix typo regarding refactoring from course_factory to taskset_factory

AlexandreDoneux:
During a refactoring in order to get rid of the course_factory, some "course_factory" were replaced by "taskset_factory" and should not have been.

This PR simply restores the right variables.


anthonygego:
LGTM

PR #1027, opened 2024-12-08
Fix - removed indentation in tutorial

anaHue:
This is a PR to solve an issue related to the doc.

Looking at the tutorial of how to write an hello world exercise, it stated an example of the code to put in a template file to run the student code.

The problem is that the student code is inside the function and the funtion was called at the end of the function instead of being called after the function definition. This would result in not being able to run the student code but even by running the function it will be an infinite loop. 

To solve this, I had to remove one indentation level as Python works with identation at the right level to call a function. By testing, the code was indeed showing nothing before the change and displayed "hello world" after changes.

### DIFF
--- a/doc/teacher_doc/task_tuto.rst
+++ b/doc/teacher_doc/task_tuto.rst
@@ -72,10 +72,9 @@ in the webapp in the *Task files* tab of the *Edit task* page.
 #. Create a template file ``template.py``, where we will put the code of the student.
    ::
 
-       def func():
-           @    @question1@@
-
-           func()
+	def func():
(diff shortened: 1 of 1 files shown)


AlexandreDoneux:
You gave no explanation on what the problem was and how you fixed it. Please provide those in your PR description. 

Also provide proof there was a problem and that your changes are valid and do not cause more problems.

anaHue:
Indeed I forgot to give context as I thought it was clear of the changes I did. I added what you asked and hope it is greatly structured. Thanks for the feedback, I tend to forgot to bring context.

PR #1040, opened 2025-03-12
Use of the Beanie model for user handling 

AlexandreDoneux:
This PR draft is to display a possible way to use beanie as an ODM for DB interactions. Using an ODM would allow us the handle data more easily as objects and handle errors in a more secure way for the app. It would in a greater measure improve the code quality. Beanie being based on Pydantic also serves data validation purposes. 

This first PR focuses on the handling of users as it is one of the simplest structures we have so far. It includes for the moment : 
- An additional asyncio loop specifically for Motor calls (the python drive Beanie is based on).
- A User model and a couple of validators whose job it is to validate the data when creating a new document or fetching one from the database.
- Data validation and DB queries for the install script and user registration/password reset

As  the work was done exploring the capabilities of Beanie and Pydantic, multiple things still need to be implemented before considering changing this draft in a finished PR : 

- [ ] Implementing testing for the model 
- [ ] Using Beanie for Validation and queries at the other places in the app 
   - [ ] ...
- [ ] ...

The broader purpose for this User model is also to get rid of the UserManager singleton and move all their methods inside the use model if possible.


inginious/frontend/app.py
```diff
@@ -127,13 +130,28 @@ def get_app(config):
     :param config: the configuration dict
     :return: A new app
     """
-    # First, disable debug. It will be enabled in the configuration, later.
+
+    zmq_context, __ = start_zmq_asyncio_loop(config.get('debug_asyncio', False))
 
     config = _put_configuration_defaults(config)
     mongo_client = MongoClient(host=config.get('mongo_opt', {}).get('host', 'localhost'))
     database = mongo_client[config.get('mongo_opt', {}).get('database', 'INGInious')]
```
nrybowski, line 138: I guess that those line are still there only to ensure backward compatibility with the currently unchanged singletons?

nrybowski (commented):
After a very quick look, I think that the task folder should not be part of the PR :)

Also, to clarify the structure of the PR, I suggest (i) to squash the commits related to the "plumbing" of initializing Beanie and its related I/O loop in a single commit; (ii) add the User model and related tests in another commit and (iii) put the replacement of the current UserManager by the User model in a third commit.

inginious/frontend/app.py
AlexandreDoneux: That's right. We still need those until every call/request is replaced using Beanie.

inginious/frontend/pages/register.py
```diff
@@ -201,14 +205,19 @@ def POST(self):
         msg = ""
         error = False
         data = flask.request.form
+        loop = flask.current_app.motor_loop
         if "register" in data:
-            msg, error = self.register_user(data)
+            future = asyncio.run_coroutine_threadsafe(self.register_user(data), loop)
```
anthonygego, line 210: We should transform GET and POST methods into coroutine. Flask seems to embed some support for asyncio : https://flask.palletsprojects.com/en/stable/async-await/

We should probably just find out how to use this feature with the Flask `MethodView` class

PR #1041, opened 2025-03-25
[Feature] API tokens page

louis-ab:
This PR adds a page to the course administration menu than allows admins to create and manage API tokens so that we can later use them in an API.

The purpose of this future API will be to retrieve information and statistics on student submissions.

### DIFF
--- a/configuration.deploy.yaml
+++ b/configuration.deploy.yaml
@@ -12,5 +12,6 @@ session_parameters:
     timeout: 86400
 superadmins:
 - superadmin
+jwt_key: p3JcQcTwcnKxr39t7p6B7cbaE8CvuPPoOj3kE4HyoCrlKM10BLly72X3Q90uJ8kaDb
 tasks_directory: /inginious/tasks
 use_minified_js: true
--- a/configuration.deploy.yaml
+++ b/configuration.deploy.yaml
@@ -12,6 +12,7 @@ session_parameters:
     timeout: 86400
 superadmins:
 - superadmin
+# You can generate your jwt secret with "openssl rand -base64 64"
 jwt_key: p3JcQcTwcnKxr39t7p6B7cbaE8CvuPPoOj3kE4HyoCrlKM10BLly72X3Q90uJ8kaDb
 tasks_directory: /inginious/tasks
 use_minified_js: true
--- a/inginious/frontend/flask/mapping.py
+++ b/inginious/frontend/flask/mapping.py
@@ -47,6 +47,7 @@ from inginious.frontend.pages.course_admin.task_edit_file import CourseTaskFileU
 from inginious.frontend.pages.course_admin.danger_zone import CourseDangerZonePage
 from inginious.frontend.pages.course_admin.statistics import CourseStatisticsPage
 from inginious.frontend.pages.course_admin.search_user import CourseAdminSearchUserPage
+from inginious.frontend.pages.course_admin.api_tokens import CourseAPITokensPage
 
 
 def init_flask_maintenance_mapping(flask_app):
@@ -138,3 +139,5 @@ def init_flask_mapping(flask_app):
                            view_func=AdministrationUsersPage.as_view('administrationuserspage'))
     flask_app.add_url_rule('/administrator/user_action',
                            view_func=AdministrationUserActionPage.as_view('administrationuseractionpage'))
+    flask_app.add_url_rule('/<cookieless:sessionid>admin/<courseid>/api_tokens',
+                           view_func=CourseAPITokensPage.as_view('courseapipage'))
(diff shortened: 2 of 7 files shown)


configuration.deploy.yaml
```diff
@@ -12,5 +12,6 @@ session_parameters:
     timeout: 86400
 superadmins:
 - superadmin
+jwt_key: p3JcQcTwcnKxr39t7p6B7cbaE8CvuPPoOj3kE4HyoCrlKM10BLly72X3Q90uJ8kaDb
```
anthonygego, line 15: Can you add a comment on how to generate this string so that people could replace it easily on setup ?

The key should also be replicated with small explanation in `INGInious/docker-compose.yml`

As I guess this is a mandatory parameter, we should also update the installer script to generate it appropriately in `INGInious/inginious/frontend/installer.py`.

inginious/frontend/templates/course_admin/api_tokens.html
```diff
@@ -0,0 +1,94 @@
...
+                            <td class="text-secondary">{{ _("No description") }}</td>
+                        {% else %}
+                            <td>{{token["description"]}}</td>
+                        {% endif %}
+                        {% if str(token["expire"]) == "9999-12-31 00:00:00" %}
+                            <td class="text-secondary">{{ _("Doesn't expire") }}</td>
+                        {% elif token["expire"] >= now %}
+                            <td>Expire {{token["expire"].strftime("%Y %m %d %H:%M:%S")}}</td>
+                        {% else %}
+                            <td class="text-danger">Expired since {{token["expire"].strftime("%Y %m %d %H:%M:%S")}}</td>
```
anthonygego, line 85: We should support i18n here too.

inginious/frontend/flask/mapping.py
```diff
@@ -151,3 +152,5 @@ def init_flask_mapping(flask_app):
                            view_func=AdministrationUsersPage.as_view('administrationuserspage'))
     flask_app.add_url_rule('/<cookieless:sessionid>administrator/user_action',
                            view_func=AdministrationUserActionPage.as_view('administrationuseractionpage'))
+    flask_app.add_url_rule('/<cookieless:sessionid>admin/<courseid>/api_tokens',
```
anthonygego, line 155: This path needs to be updated.

inginious/frontend/pages/course_admin/api_tokens.py
```diff
@@ -0,0 +1,58 @@
...
+    def POST_AUTH(self, courseid):
+        """ POST request """
+        course, __ = self.get_course_and_check_rights(courseid)
+        user_input = flask.request.form
+
+        if "add_token" in user_input:
+            descr = user_input.get("descr", "")
+            expire = user_input.get("expiration", "")
+            if expire == "":
+                expire = datetime(9999, 12, 31)
```
anthonygego, line 32: Use https://docs.python.org/3/library/datetime.html#datetime.date.max

inginious/frontend/templates/course_admin/api_tokens.html
```diff
@@ -0,0 +1,94 @@
...
+                {% for token in tokens %}
+                    {% set tok_id = token["_id"] %}
+                    <tr id="{{loop.index}}">
+                        <td>ID : {{token["_id"]}}</td>
+                        {% if token["description"] == "" %}
+                            <td class="text-secondary">{{ _("No description") }}</td>
+                        {% else %}
+                            <td>{{token["description"]}}</td>
+                        {% endif %}
+                        {% if str(token["expire"]) == "9999-12-31 00:00:00" %}
```
anthonygego, line 80: It would be better not to hardcode the max date here, either these tests can be performed at the Python side, either date.max can be passed.

inginious/frontend/templates/course_admin/api_tokens.html
```diff
@@ -0,0 +1,94 @@
...
+                            <td class="text-secondary">{{ _("Doesn't expire") }}</td>
+                        {% elif token["expire"] >= now %}
+                            <td>Expire {{token["expire"].strftime("%Y %m %d %H:%M:%S")}}</td>
```
anthonygego, line 83: Why don't you use the same converter ? `str` or, as documented, `isoformat(' ')` ?

inginious/frontend/templates/course_admin/api_tokens.html
```diff
@@ -0,0 +1,94 @@
...
+<div class="card-body">
+    <form class="form-horizontal" id="remove_form" method="post">
+        <table id="table" class="table">
+            <tbody>
+                {% if msg == "removed" %}
+                    <p>{{ _("Token removed") }}</p>
+                {% endif %}
+                {% for token in tokens %}
+                    {% set tok_id = token["_id"] %}
+                    <tr id="{{loop.index}}">
```
anthonygego, line 73: Adding a prefix to the id is a good practice.

anthonygego (changes requested):
My comment from 2 weeks ago is still relevant.

anthonygego:
There are still conflicts. According to the diff, I suspect your local branch is not correctly rebased on 'main`. There should not be removals of `<cookieless>/` mappings in this PR.

PR #1042, opened 2025-03-25
[Feature] API endpoint for retrieving submissions data

louis-ab:
This PR adds an API that can be used to retrieve tasks and submissions data for a given course. The data is returned as JSON if the request has a valid token.

### DIFF
--- a/inginious/frontend/flask/mapping.py
+++ b/inginious/frontend/flask/mapping.py
@@ -47,6 +47,7 @@ from inginious.frontend.pages.course_admin.task_edit_file import CourseTaskFileU
 from inginious.frontend.pages.course_admin.danger_zone import CourseDangerZonePage
 from inginious.frontend.pages.course_admin.statistics import CourseStatisticsPage
 from inginious.frontend.pages.course_admin.search_user import CourseAdminSearchUserPage
+from inginious.frontend.pages.course_admin.api_tokens import CourseAPITokensPage
 
 
 def init_flask_maintenance_mapping(flask_app):
@@ -138,3 +139,5 @@ def init_flask_mapping(flask_app):
                            view_func=AdministrationUsersPage.as_view('administrationuserspage'))
     flask_app.add_url_rule('/administrator/user_action',
                            view_func=AdministrationUserActionPage.as_view('administrationuseractionpage'))
+    flask_app.add_url_rule('/<cookieless:sessionid>admin/<courseid>/api_tokens',
+                           view_func=CourseAPITokensPage.as_view('courseapipage'))
--- a/inginious/frontend/flask/mapping.py
+++ b/inginious/frontend/flask/mapping.py
@@ -19,7 +19,7 @@ from inginious.frontend.pages.preferences.profile import ProfilePage
 from inginious.frontend.pages.preferences.utils import PrefRedirectPage
 from inginious.frontend.pages.utils import SignInPage, LogOutPage
 from inginious.frontend.pages.register import RegistrationPage
-from inginious.frontend.pages.social import AuthenticationPage, CallbackPage, SharePage
+from inginious.frontend.pages.social import AuthenticationPage, CallbackPage
 from inginious.frontend.pages.course_register import CourseRegisterPage
 from inginious.frontend.pages.course import CoursePage
 from inginious.frontend.pages.tasks import TaskPage, TaskPageStaticDownload
@@ -40,7 +40,6 @@ from inginious.frontend.pages.course_admin.student_info import CourseStudentInfo
 from inginious.frontend.pages.course_admin.submission import SubmissionPage
 from inginious.frontend.pages.course_admin.submissions import CourseSubmissionsPage
 from inginious.frontend.pages.course_admin.task_list import CourseTaskListPage
-from inginious.frontend.pages.course_admin.tags import CourseTagsPage
 from inginious.frontend.pages.course_admin.audience_edit import CourseEditAudience
 from inginious.frontend.pages.course_admin.task_edit import CourseEditTask
 from inginious.frontend.pages.course_admin.task_edit_file import CourseTaskFiles
@@ -108,7 +107,11 @@ def init_flask_mapping(flask_app):
                            view_func=CourseSubmissionsPage.as_view('coursesubmissionspage'))
     flask_app.add_url_rule('/admin/<courseid>/tasks',
                            view_func=CourseTaskListPage.as_view('coursetasklistpage'))
+<<<<<<< HEAD
     flask_app.add_url_rule('/admin/<courseid>/edit/audience/<audienceid>',
+=======
+    flask_app.add_url_rule('/<cookieless:sessionid>admin/<courseid>/edit/audience/<audienceid>',
+>>>>>>> 310c79e0 (fix accidental erasure)
                            view_func=CourseEditAudience.as_view('courseditaudience'))
     flask_app.add_url_rule('/admin/<courseid>/edit/task/<taskid>',
                            view_func=CourseEditTask.as_view('coursedittask'))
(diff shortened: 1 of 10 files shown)


inginious/frontend/flask/mapping.py
```diff
@@ -154,3 +155,5 @@ def init_flask_mapping(flask_app):
                            view_func=AdministrationUserActionPage.as_view('administrationuseractionpage'))
     flask_app.add_url_rule('/<cookieless:sessionid>admin/<courseid>/api_tokens',
                            view_func=CourseAPITokensPage.as_view('courseapipage'))
+    flask_app.add_url_rule('/<cookieless:sessionid>admin/<courseid>/api_auth',
```
anthonygego, line 158: So in our case we are considering tokens generated from the user interface are already authorized tokens. They should be used to access any API page. You therefore have to implement a common authorization check on top of every API page POST/GET processing methods.

inginious/frontend/pages/course_admin/api_auth.py
```diff
@@ -0,0 +1,75 @@
...
+from flask import request, Response
+import jwt
+from datetime import datetime
+import json
+from bson.json_util import dumps
+
+from inginious.frontend.pages.course_admin.utils import INGIniousAdminPage
+
+
+class CourseAPIAuthPage(INGIniousAdminPage):
```
anthonygego, line 15: `INGIniousPage`  should be enough. The best option is to recycle `INGIniousAPI` page and harmonize with existing code.

inginious/frontend/pages/course_admin/api_auth.py
```diff
@@ -0,0 +1,75 @@
...
+                msg = dumps(self.database.submissions.find(params))
+            else:
+                msg = json.dumps({"message": "No data type specified"})
+        else:
+            # error message
+            msg = json.dumps({"message": msg})
+
+        return self.page(msg, code)
+
+    def decode_token(self, auth_header):
```
anthonygego, line 41: So this should be part of an abstract `Page` class, for instance.

inginious/frontend/pages/course_admin/api_auth.py
```diff
@@ -0,0 +1,75 @@
...
+        """ Returns decoded courseid or error, and http code """
+        key = str(self.app.jwt_key)
```
anthonygego, line 43: No need to parse into string each time. Just check this at app initialization.

inginious/frontend/pages/course_admin/api_auth.py
```diff
@@ -0,0 +1,75 @@
...
+
+        try:
+            if auth_header[:7] != "Bearer ":
```
anthonygego, line 46: Do not implement HTTP specs by yourself. Rely on external librairies that will handle this seamlessly while being updated (well, that remains a hope in practice) in the spec changes. Take a look at Flask-HTTPAuth for instance.

There may even be librairies that directly handle token authentication.

anthonygego (changes requested):
I strongly recommend you to look at the existing API implementation as it should be usable again in the future. But also to make a review of existing litterature. Look at tutorials, articles, existing third-party implementation, to make sure that:
- you understand correctly what is a REST API (the current implementation is not), what are the common authorization process (OIDC, OAuth,...) and how they work, token generation,...
- your implementation is quite state of the art

We decided to skip the authorization part here, where the user logs into the platform to generate a token with appropriate permission.

All these concepts should be part of your report as people will have to understand what and why you've done.

inginious/frontend/flask/mapping.py
```diff
@@ -138,3 +141,9 @@ def init_flask_mapping(flask_app):
                            view_func=AdministrationUsersPage.as_view('administrationuserspage'))
     flask_app.add_url_rule('/administrator/user_action',
                            view_func=AdministrationUserActionPage.as_view('administrationuseractionpage'))
+    flask_app.add_url_rule('/<cookieless:sessionid>admin/<courseid>/api_tokens',
+                           view_func=CourseAPITokensPage.as_view('courseapipage'))
+    flask_app.add_url_rule('/<cookieless:sessionid>admin/<courseid>/submissions_data',
```
anthonygego, line 146: The `cookieless` url map converter has been removed in the `main` branch. I doubt this would even work now.

I expect the routes to match/look like the existing API ones.

inginious/frontend/pages/course_admin/api_auth.py
```diff
@@ -0,0 +1,50 @@
...
+class DataAPIPage(INGIniousPage):
+    """ Token verification and handler for the data API """
+
+    def __init__(self):
+        super().__init__()
+        self.jwt_key = str(self.app.jwt_key)
+
+    def verify(self):
+        """ verify if a token is valid """
+        auth_header = request.headers.get('Authorization')
```
anthonygego, line 23: What happens if there is no such key in the headers dictionary ?

inginious/frontend/pages/course_admin/api_auth.py
```diff
@@ -0,0 +1,50 @@
...
+            expire = datetime.strptime(decoded["expire"], "%Y-%m-%d %H:%M:%S")
+            courseid = decoded["courseid"]
+        except:
+            abort(401, description="Invalid token.")
+
+        stored_token = self.database.tokens.find_one({"$expr": {"$eq": ["$_id", {"$toObjectId": tok_id}]}})
+        if stored_token is None:
+            abort(401, description="Authentification token not recognized.")
+
+        if expire < datetime.now():
```
anthonygego, line 40: It seems there exists facilities to check token data embedded in the `jwt.decode()`function : https://pyjwt.readthedocs.io/en/stable/api.html#jwt.decode

Can't we rely on this ?

inginious/frontend/pages/course_admin/api_auth.py
```diff
@@ -0,0 +1,50 @@
...
+        except:
+            abort(401, description="No authentification token found.")
+        try:
+            tok_id = decoded["_id"]
+            expire = datetime.strptime(decoded["expire"], "%Y-%m-%d %H:%M:%S")
+            courseid = decoded["courseid"]
+        except:
+            abort(401, description="Invalid token.")
+
+        stored_token = self.database.tokens.find_one({"$expr": {"$eq": ["$_id", {"$toObjectId": tok_id}]}})
```
anthonygego, line 36: Geee... such a complex request to fetch an element by its ObjectId ?? Import and use `bson.ObjectId` to perform a simple `_id` match request.

inginious/frontend/pages/course_admin/submissions_data.py
```diff
@@ -0,0 +1,28 @@
...
+from bson.json_util import dumps
+
+from inginious.frontend.pages.course_admin.api_auth import DataAPIPage
+
+class SubmissionsEndpoint(DataAPIPage):
+    """ Endpoint to retrieve submission data """
+
+    def GET(self, courseid):
+        """ GET request """
+        courseid = self.verify()
```
anthonygego, line 15: Embed this in your parent class.

inginious/frontend/pages/course_admin/submissions_data.py
```diff
@@ -0,0 +1,28 @@
...
+
+        # DB request parameters
+        params = {"courseid": courseid}
```
anthonygego, line 18: It seems you don't check the token course scope and give access to the whole database as long as the token is valid.

inginious/frontend/pages/course_admin/submissions_data.py
```diff
@@ -0,0 +1,28 @@
...
+        if "taskid" in request.args:
```
anthonygego, line 19: Some safety checks might be interesting, as well as, at least, the same level of request convenience as the existing GUI.

If a user only wants to fetch evaluation submissions, they should'nt filter the whole databases results in their own code.

anthonygego (changes requested):

inginious/frontend/pages/course_admin/api_auth.py
louis-ab: In this case request.headers.get('Authorization') will return an empty string

PR #1023, opened 2024-10-15
LTI 1.3

mpiraux:
This PR updates the LTI support of INGInious to LTI 1.3, along with a couple of internal changes.

The support of LTI 1.1 is dropped by these changes so all integrations must be updated when integrating this code into running instances.

I'll detail the contributions as they are integrated into the PR.

acf5b37a7acbeefd03759a7d27c7ef24cdfcbcd1 has been tested with a test instance of the syllabus and INGInious. I tried submitting code to a task and browsing submissions.

PR #1043, opened 2025-03-25
Add support for LTI 1.3

anthonygego:
This is a rebased and updated version of #1023 from @mpiraux, supporting both LTI1.1 and LTI1.3 versions :

> 2238b5e has been tested with a test instance of the syllabus and INGInious with the existing LTI code. I tried submitting code to a task and browsing submissions.

This is already merged. Below commits are yet to merge.

- e90377ab06790c1fc30942c4e5c9e9c5cb3c0a76 refactors current LTI implementation in order to add a new version more easily
- f738dd96669997f10b1904b8319a1752ee22a9d4 refactors the OutcomeManager (score publishers) in order to support  another LTI version
- 6a15dfe4848b9e2a19f7c8a55d7d888f4325a111 adds the config details to the course
- 0e01f8219e6931c4ccbee0fe341a1cf746c27f6d provides LTI1.3 account binding and task access
- bba31db648bf8891d7f681215e6fcb429e2b5e9c provides LTI1.3 score publishing
- bec26e1240beb2c417018bba8c48863c79c0d666 refactors score publishing using class inheritance.
- 8960522165c2d31046f88730ca097a2924e54c34 updates the documentation


inginious/frontend/flask/mongo_sessions.py
```diff
@@ -67,31 +66,26 @@ def _get_signer(self, app):
...
-            cookieless = True
-            sid = path_session
-        elif is_lti_launch:
-            cookieless = True
-            sid = None
-        else:
-            cookieless = False
-            sid = request.cookies.get(self.get_cookie_name(app))
+        if lti_session or is_lti_launch:
+            return None
```
anthonygego, line 81: My understanding here is that, before, we returned the result of `self.session_class(sid=None)` while here we return directly `None`. Is that equivalent ?

inginious/frontend/pages/utils.py
```diff
@@ -187,7 +188,7 @@ def GET(self, *args, **kwargs):
                     and not self.__class__.__name__ == "ProfilePage":
                 return redirect("/preferences/profile")
 
-            if not self.is_lti_page and self.user_manager.session_lti_info() is not None:  # lti session
+            if not self.is_lti_page and self.user_manager.session_lti_info() is not None:  # lti session, TODO(mp): Not sure whether it is still needed
```
anthonygego, line 191:
```suggestion
            if not self.is_lti_page and self.user_manager.session_lti_info() is not None: 
```
I think it is still. There were pages defined not to be accessible through an LTI session. For instance, groups features (even if they should not be by definition), admin pages,...

inginious/frontend/pages/utils.py
```diff
@@ -214,7 +215,7 @@ def POST(self, *args, **kwargs):
             if not self.user_manager.session_username() and not self.__class__.__name__ == "ProfilePage":
                 return redirect("/preferences/profile")
 
-            if not self.is_lti_page and self.user_manager.session_lti_info() is not None:  # lti session
+            if not self.is_lti_page and self.user_manager.session_lti_info() is not None:  # lti session, TODO(mp): Not sure whether it is still needed
```
anthonygego, line 218:
```suggestion
            if not self.is_lti_page and self.user_manager.session_lti_info() is not None:
```

inginious/frontend/pages/lti.py
```diff
@@ -52,78 +54,175 @@ class LTIBindPage(INGIniousAuthPage):
...
-        return session_id, cookieless_session["lti"]
-
-    def GET_AUTH(self):
-        input_data = flask.request.args
-        if "session_id" not in input_data:
-            return self.template_helper.render("lti_bind.html", success=False, session_id="",
-                                               data=None, error=_("Missing LTI session id"))
+    def _get_lti_session_data(self):
+        if not self.user_manager.session_is_lti():
+            return self.template_helper.render("lti_bind.html", success=False, data=None, error=_("Missing LTI session id"))
```
anthonygego, line 59: The method returns a tuple below, not here. Besides, I would avoid returning rendered templates directly and try to gather these calls to `render` in one unique method, even if this is not the case for most of the existing code.

inginious/frontend/lti_grade_manager.py
```diff
@@ -0,0 +1,153 @@
...
+    to submit grades later using the LTIGradeManager.
+    """
+    def __init__(self, database, courseid, taskid, *args, **kwargs) -> None:
+        self.database = database
+        self.query_context = (courseid, taskid)
+        self._session_cookie_name = ""  # Disables session scope mechanism in favor of query_context
+        super().__init__(*args, **kwargs)
+
+    def can_set_keys_expiration(self) -> bool:
+        return False  # TODO(mp): I think it's reasonable to clean LTI Launch messages further than a week away tho
```
anthonygego, line 33: Indeed.

inginious/frontend/pages/lti.py
```diff
@@ -52,78 +54,175 @@ class LTIBindPage(INGIniousAuthPage):
...
+        tool_desc = tool.get('description', 'N/A')
+        tool_url = tool.get('url', 'N/A')
+        context = launch_data['https://purl.imsglobal.org/spec/lti/claim/context']
+        context_title = context.get('context_title', 'N/A')
+        context_label = context.get('context_label', 'N/A')
+
+        can_report_grades = message_launch.has_ags() and tool_conf.get_iss_config(iss=message_launch.get_iss(),
+                                                                                  client_id=message_launch.get_client_id()).get('auth_token_url')
+
+        session_id = hashlib.sha256(launch_id.encode('utf-8')).digest().hex()  # TODO(mp): Make this more secure
```
anthonygego, line 187: This depends on what the launch id is based on. However, as it is probably shared externally, it would indeed be more secure to add some private hint in the session id generation to avoid session hijacking. Maybe based on the flask session secret parameter.

inginious/frontend/lti_grade_manager.py
```diff
@@ -0,0 +1,153 @@
...
+                    continue
+
+                try:
+                    message_launch = FlaskMessageLaunch.from_cache(message_launch_id, request=None, tool_config=course.lti_tool(), launch_data_storage=MongoLTILaunchDataStorage(self._app.database, courseid, taskid))
+                    launch_data = message_launch.get_launch_data()
+                    ags = message_launch.get_ags()
+
+                    if ags.can_put_grade():
+                        sc = Grade()
+                        # TODO(mp): Is there a better timestamp to set with the score? Submission time? Grading time?
```
anthonygego, line 88: Maybe it's better to always set the score as the latest for the platform (and therefore keep the `datetime.now()`). This way we ensure the score shown on INGInious is the one that is stored by the platform.

anthonygego (commented):
Globally this LGTM.

PR #1046, opened 2025-04-23
Api token page for non-admin users

louis-ab:
This PR adds a page to manage tokens for individual users in the regular course menu, like the one in the course admin menu.

The API endpoints and database fields are adapted with additional checks.


anthonygego:
I'm closing this as this set of PRs as they would need some rework and this feature is not a priority for now.

Issue #898, opened 2022-11-29
[agent_docker] Docker rlimits permission denied on macOS


anthonygego:
**Describe the bug**
A basic Docker install on macOS Ventura (M1) leads to the following error when running a submission : 

```
2022-11-29 14:19:10,432 - inginious.agent.docker - WARNING - Cannot start container! 500 Server Error for http+docker://localhost/v1.41/containers/c6925c36f1dd9a7b2c4f1ce091aef3507ac923154fdde9d84cdb4179a2205e63/start: Internal Server Error ("failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error setting rlimits for ready process: error setting rlimit type 7: operation not permitted: unknown")
Traceback (most recent call last):
  File "/Users/ogoletti/PycharmProjects/INGInious/venv/lib/python3.9/site-packages/docker/api/client.py", line 268, in _raise_for_status
    response.raise_for_status()
  File "/Users/ogoletti/PycharmProjects/INGInious/venv/lib/python3.9/site-packages/requests/models.py", line 1021, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 500 Server Error: Internal Server Error for url: http+docker://localhost/v1.41/containers/c6925c36f1dd9a7b2c4f1ce091aef3507ac923154fdde9d84cdb4179a2205e63/start
```

**INGInious installation details**
- Version: 0.8.2+
- Containers version (if applicable): same

**To Reproduce**
1. Install and run Docker
2. Install and run INGInious with default options
3. Run the tutorial first task
4. Feedback "cannot create container"

**Expected behavior**
Expected feedback

zvikfir:
I'm having the same issue. Is there a known fix to this?

nrybowski:
For reference, the recommended solution for this issue is cleanly close the Docker Agent and logging an error.
See https://github.com/INGInious/INGInious/pull/1051#issuecomment-2980438214 for more details.

PR #1051, opened 2025-06-13
Drop Use of Ulimits When Running on MacOS

zvikfir:
The rlimits syscall is not available when running on MacOS. Hence, I dropped the use of ulimits when running on MacOS and added a warning log.

Related to #898.

### DIFF
--- a/inginious/agent/docker_agent/_docker_interface.py
+++ b/inginious/agent/docker_agent/_docker_interface.py
@@ -12,6 +12,7 @@ from typing import List, Tuple, Dict
 
 import docker
 import logging
+import platform
 
 from docker.types import Ulimit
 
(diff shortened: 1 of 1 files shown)


nrybowski:
> I can't think of a reason someone would want to deploy INGInious on MacOS in production. However, developing on MacOS is quite common, and this small fix is the only one needed currently to support it out of the box.

On the one hand, ULimits are security features we leverage to ensure that student submissions won't harm the Agents. Dropping such limits in production deployment don't really make sense but _might_ be acceptable for development / testing purpose.
On the other hand, the issue here is not the patch size, but the fact that we are currently trying to reduce the project's code complexity and refactor some major parts of the project. We can not afford supporting multiple platforms at the moment without a clear requirement from the community.

As mentioned earlier, for testing / developing purpose, we recommend leveraging a Linux VM either to deploy the entire project or to deploy the Docker Agent only. In the latter case, developers should [adapt the `DOCKER_HOST` environment variable as documented](https://github.com/INGInious/INGInious/blob/main/doc/admin_doc/install_doc/installation.rst#with-local-backendagent-and-a-webserver).

That being said, we should adapt the documentation section to discuss the issue of Docker on MacOS.

I'm closing the PR for now but feel free to re-open it if at some point there is a need for production deployment of Docker Agents on MacOS.

PR #1066, opened 2025-07-25
[scripts] update docker repository to ghcr.io

### LINKED ISSUE #1062: [scripts/container-update] Replace hardcoded container URL to ghrc.io
The [container_update.py](https://github.com/INGInious/INGInious/blob/main/inginious/scripts/container_update.py) script used to update every INGInious environment installed along an INGInious instance use [here and there hard-coded references to the DockerHub with the outdated container nomenclature](https://github.com/INGInious/INGInious/blob/main/inginious/scripts/container_update.py#L54), e.g., `ingi/inginious-c-base`.

Since then, we moved to the Github registry and simplified the container nomencalture, i.e., `ghcr.io/inginious/env-base`. The [container_update.py](https://github.com/INGInious/INGInious/blob/main/inginious/scripts/container_update.py) script must be updated to reflect that change, as well as the related documentation.

See also #1056.

zyfy29:
fix #1062

replace the hard-coded dockerhub url with `ghcr.io/inginious/env-base` and `ghcr.io/inginious/env-default`

### DIFF
--- a/inginious/scripts/container_update.py
+++ b/inginious/scripts/container_update.py
@@ -51,7 +51,7 @@ def main():
         docker_connection = docker.from_env() 
         for image in docker_connection.images.list():
             for tag in image.attrs["RepoTags"]:
-                if re.match(r"^ingi/inginious-c-(base|default):v" + __version__, tag):
+                if re.match(r"^ghcr.io/inginious/env-(base|default):v" + __version__, tag):
                     stock_images.append(tag)
     except:
         print(FAIL + "Cannot connect to Docker!" + ENDC)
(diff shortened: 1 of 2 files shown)


nrybowski:
@zyfy29 Thanks for your contribution. I would prefer configuring the registry through an environment variable to make this script more flexible. [This the approach we use for the service images](https://github.com/INGInious/INGInious/blob/main/deploy/frontend.containerfile#L4).

zyfy29:
@nrybowski Thanks, I see it.
How do you think about adding a config item to indicate the source (dockerhub / ghcr) in `configuration.example.yaml` like what `database_update.py` does?
The `.env` file seems to be for docker compose so it might be ambiguous to read the `.env` directly from python code.

zyfy29:
How about the implementation now? I added the `python-dotenv` dependence to do so.

anthonygego:
LGTM. Thanks for the contribution.

PR #1067, opened 2025-08-01
Fix user creation using bind_user, and add option to force username to match binding name

haved:
The current implementation of `bind_user` seemingly mixes between `user` being a tuple, and being indexable as a dictionary containing `["username"]`, `["realname"]` etc.

The [frontend/plugins/auth/ldap_auth.py](https://github.com/INGInious/INGInious/blob/main/inginious/frontend/plugins/auth/ldap_auth.py) file calls `bind_user` with a simple tuple, which causes exceptions when creating new users.

This PR tries to fix this by expecting `user` to be a tuple, which works in my particular case, but I am unable to see if any code was relying on the old behavior.

### DIFF
--- a/inginious/frontend/user_manager.py
+++ b/inginious/frontend/user_manager.py
@@ -486,11 +486,13 @@ class UserManager:
         user = self._database.users.find_one_and_update({"activate": activate_hash}, {"$unset": {"activate": True}})
         return user is not None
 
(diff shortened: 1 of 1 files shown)


anthonygego:
Indeed, the issue was introduced by commit d958c29a158b9f11e904aedeb8654e4bb86a381f.
Thanks !

PR #1079, opened 2025-10-08
[cleanup 1] Template helper cleanup

anthonygego:
This PR cleans up the `TemplateHelper` implementation : 

- It replaces Python methods `user_selection_box`, `preferences_menu` and `course_admin_menu` by pure Jinja code. 
- Methods `is_lti()` and `tojson()` are also removed as there are redundant implementations available.
- Javascript and CSS hooks are simplified and called directly in the `layout.html` template as the plugin manager is available in the globals. `add_css` and `add_javascript` methods are not used anymore in the main codebase and are not exposed to plugins. They were removed too.
- It removes the `welcome_text` hook as it was only shown on LTI login page since the introduction of static pages and the `welcome_page` option.
- Base helpers are replaced by Jinja equivalent code
- `template_helper` hooks are called at the rendering time and not when `template_helper.call(...)` is called. 

The final goal is to completely remove the `TemplateHelper` and rely on core Flask/Jinja routines and the `PluginManager` for hooks.

This is the first of a series of cleaning patches.

### DIFF
--- a/inginious/frontend/template_helper.py
+++ b/inginious/frontend/template_helper.py
@@ -4,32 +4,19 @@
 # more information about the licensing of this file.
 
 """ TemplateManager """
-import os
-from functools import lru_cache
-
-from jinja2 import Environment, FileSystemLoader, select_autoescape
-import inginious
+from jinja2 import  FileSystemLoader
+from flask import current_app, render_template
+from inginious import get_root_path
 
 class TemplateHelper(object):
     """ Class accessible from templates that calls function defined in the Python part of the code. """
 
-    def __init__(self, plugin_manager, use_minified=True):
+    def __init__(self, plugin_manager):
         """
         Init the Template Helper
         :param plugin_manager: an instance of a PluginManager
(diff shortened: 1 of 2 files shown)


inginious/frontend/template_helper.py
```diff
@@ -64,77 +26,9 @@ def render(self, path, template_folder="", **tpl_kwargs):
...
+            current_app.jinja_loader = FileSystemLoader([get_root_path() + '/frontend/templates', template_folder])
 
-    def _json_safe_dump(self, data):
-        """ Make a json dump of `data`, that can be used directly in a `<script>` tag. Available as json() inside templates """
-        return json.dumps(data).replace(u'<', u'\\u003c') \
-            .replace(u'>', u'\\u003e') \
-            .replace(u'&', u'\\u0026') \
-            .replace(u"'", u'\\u0027')
\ No newline at end of file
+        current_app.jinja_env.globals.update(dict(self._plugin_manager.call_hook("template_helper")))
```
AlexandreDoneux: What does the "template_helper" represent and where is it defined ?

PR #1081, opened 2025-10-13
[cleanup 3] Refactor i18n

anthonygego:
This follows #1080 (and therefore depends on) in a series of cleaning patches.

It reimplements the `L10nManager` as a `i18n` module. Reference to `UserManager` is now fetched from the flask app globals if needed.

In addition to removing lines, this actually allows to progressively replace all implicit `_` calls to gettext by:

    from inginious.frontend.i18n import gettext as _

This does not change the course and task translation mechanism. However, it removes the passing of the `GNUTranslations` object from `ParsableText`. This was actually quite obscure as it theoretically allowed fetching translations for RST directives from the course and task translations. However, it is impossible to define an RST directive in the specific context of a course/task and was not planned for plugins. All directive translations should come from the main translations files.

### DIFF
--- a/inginious/frontend/task_problems.py
+++ b/inginious/frontend/task_problems.py
@@ -165,13 +165,28 @@ class DisplayableMultipleChoiceProblem(MultipleChoiceProblem, DisplayableProblem
     def get_type_name(cls, language):
         return _("multiple choice")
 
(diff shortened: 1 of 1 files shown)


AlexandreDoneux:
LGTM

PR #1083, opened 2025-10-16
[cleanup 5] Remove TaskFactory

anthonygego:
This follows the previous PR and : 

- Implements `Task.get()` and `Task.save()` and `Task.delete()` to mimick what would be done with a database object modeling
- Removes the `TaskFactory`
- Drop task cache invalidation when a `course.yaml` is reloaded. There actually is no reason (anymore?) to do that.
- Rewrites task caching mechanism. This now stores the file as key and associated object as value. The method to fetch from the cache takes the path and a function as argument that loads the object to cache if called. It could be easily reused for the course files in the near future.

### DIFF
--- a/inginious/frontend/pages/course_admin/task_edit.py
+++ b/inginious/frontend/pages/course_admin/task_edit.py
@@ -122,7 +122,7 @@ class CourseEditTask(INGIniousAdminPage):
             return error
 
         try:
-            t = Task(taskid, data, self.course_factory.get_course_fs(courseid))
+            t = Task(taskid, data, self.course_factory.get_course_fs(courseid).from_subfolder(taskid))
         except Exception as message:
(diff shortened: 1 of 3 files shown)


inginious/scripts/autotest.py
```diff
@@ -298,7 +299,7 @@ def main():
     fs_provider = LocalFSProvider(config["task_directory"])
 
     try:
-        course_factory, _ = create_factories(fs_provider, task_dispensers, problem_types)  # used for getting tasks
+        course_factory = CourseFactory(fs_provider, task_dispensers, PluginManager(), None)  # used for getting tasks
```
AlexandreDoneux: Same here, a PluginManager as argument.

AlexandreDoneux (commented):
LGTM

PR #1077, opened 2025-09-24
Replace zip archives by archive courses

AlexandreDoneux:
This PR aims to improve the way the course backups are stored in order to make them more accessible and avoid problems with course changes over time.

We currently backup courses by storing their database data in a compressed zip archive. This was implemented at the time due to limited storage space on the machine hosting our instance. an archive could be stored somewhere else. This however has multiple flaws : 
- Problems while compressing and decompressing the zip files during the backup and restore.
- Because we store only the the database data, we do not store the tasks. Some changes in the course or tasks could create incompatibilities when restoring backups.
- Consulting a backup tedious. You need to back up the course, restore an old backup and restore the most recent backup when done.

This PR replaces the zip archives by archive courses. We do this by duplicating a course we want to back up. That way all the course data are duplicated and changes on the original course will not impact the backup. All data related to the course stay in the database and simply are redirected to be linked to the course duplicate. A new MongoDB collection is used to link the courses to their archives to make them easier to find. The other way around, we can now find the original course from an archive course.

### Steps : 
- [x] Replace zip archive by course duplication
- [x] Removing use of the backup directory and backup restore
- [x] Handle Archive courses from lists
   - [x]  Hiding archives from course list
   - [x] Visual cues in the user's course list
- [x] <del>Linking course and archive courses </del>
   - [x] <del> From course : all the archives made from that course</del>
   - [x] <del> From an archive : link to the original course</del>

As discussed with @anthonygego, we will not implement the the links between courses and their archives, as well as storage of the archive info in the database. This is related to the decision to remove the singletons present in the project. 
These ideas will be reintroduced after a thorough cleaning of the codebase and the transfer of course related data to the database.

PR #1084, opened 2025-10-16
Course pinning, search and filter feature in user's course list

AlexandreDoneux:
This PR aims to improve the user experience regarding the user's course list. Currently displayed as a basic list, it can get difficult to find a specific course if you manage a lot of them. This PR brings multiple ways to improve this list : 
- the possibility to pin certain courses to find them in a designated list
- a basic search input to limit the courses displayed in the user's courses list
- filters to limit the displayed courses in that same list. For each filter (archived, LTI, hidden, pinned), you can choose whether to show courses that match or don’t match the filter.

This PR is based on the #1077 PR. Therefore it will stay as draft until those changes are accepted.

### DIFF
--- a/inginious/frontend/static/js/mycourses.js
+++ b/inginious/frontend/static/js/mycourses.js
@@ -37,7 +37,7 @@ function course_pin(event){
                 $(pinnedItem).find(".course_pin").attr("data-course-id", courseid);
 
                 $(pinnedItem).find(".pin_title").append(data["name"]);
-                $(pinnedItem).find(".pin_link").attr("href", data["path"]);
+                $(pinnedItem).find(".pin_title").attr("href", data["path"]);
                 $(pinnedItem).removeClass("d-none");
 
                 $(document).find("#pinned_list").append(pinnedItem);


inginious/frontend/templates/mycourses.html
```diff
@@ -49,70 +49,266 @@ <h3>{{ _("Last tried exercises") }}</h3>
...
+                    </div>
+                    <div class="form-group col-sm-4">
+                        <button type="submit" class="btn btn-block btn-primary">{{ _("Create new course") }}</button>
+                    </div>
+                </form>
+            {% endif %}
+
+
+<script>
+    $(document).on('click', '.course_pin', function(event) {
```
anthonygego: Put as maximum script as possible in the standalone JS files so that they can be minified. If data from the page has to be used in the code, use the `data-*`attribute of the element triggering the event.

inginious/frontend/user_manager.py
```diff
@@ -651,7 +651,33 @@ def user_saw_task(self, username, courseid, taskid):
             upsert=True
         )
 
-    def update_user_stats(self, username, course, task, submission, result_str, grade, state, newsub, task_dispenser):
```
anthonygego: This signature is changed but I don't see any change in its usage. Is this a fix ?

inginious/frontend/user_manager.py
```diff
@@ -651,7 +651,33 @@ def user_saw_task(self, username, courseid, taskid):
...
+        return False
+
+
+    def unpin_course(self, username, courseid):
+        data = User.objects(username=username).first()
+        if data:
+            pinned_courses = data.pinned_courses if data else []
+            if courseid in pinned_courses:
+                pinned_courses.remove(courseid)
+                User.objects(username=username).update(pinned_courses=pinned_courses)
```
anthonygego: I think you can use `data.save()` here.

inginious/frontend/templates/mycourses.html
```diff
@@ -51,70 +51,154 @@ <h3>{{ _("Last tried exercises") }}</h3>
...
+
+        var open_courses = {
+            {% for courseid, course in open_courses.items() %}
+            "{{ courseid }}": {
+                is_archive: {{ course.is_archive()|tojson }},
+                is_lti: {{ course.is_lti()|tojson }},
+                is_open_to_non_staff: {{ course.is_open_to_non_staff()|tojson }}
+            }{% if not loop.last %},{% endif %}
+            {% endfor %}
+        };
```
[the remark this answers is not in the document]
anthonygego: As long as we do not have a clear separation between models, controllers and views, this is not an issue.

inginious/frontend/static/js/mycourses.js
```diff
@@ -0,0 +1,150 @@
...
+        is_pinned = (temp_is_pinned === undefined) ? ( pinned_courses.includes(courseid) ? true : false ) : temp_is_pinned;
+
+        var course_values = Object({"is_archive": course.is_archive ? true : false,
+                                     "is_lti": course.is_lti ? true : false,
+                                     "is_hidden": course.is_open_to_non_staff ? true : false,
+                                     "is_pinned": is_pinned
+                                     });
+
+        for (var filter of Object.keys(course_values)) {
+            if (filter_values[filter] === null) {
```
anthonygego: Is the next test not sufficient ?

inginious/frontend/user_manager.py
anthonygego: OK but then you need to update its usage : https://github.com/AlexandreDoneux/INGInious/blob/UI_course_list/inginious/frontend/submission_manager.py#L79

inginious/frontend/user_manager.py
AlexandreDoneux: I used data.save() to get rid of the update query on the model. We could however simplify this even more with something like : 

```python
def unpin_course(self, username, courseid):
    modified = User.objects(username=username, pinned_courses=courseid).update(pull__pinned_courses=courseid)
    return bool(modified)
```

inginious/frontend/templates/mycourses.html
```diff
@@ -49,70 +49,279 @@ <h3>{{ _("Last tried exercises") }}</h3>
...
+                                        <h5 class="card-title">
+                                        <a href=`+data["path"]+`>`+data["name"]+`</a>
+                                        </h5>
+                                        <div class="card-text card-trunc">
+                                                `+data["description"]+`
+                                        </div>
+                                    </div>
+                                </div>
+                            </div>`);
+                    }
```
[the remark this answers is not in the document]
anthonygego: This is tagged as outdated so I guess not.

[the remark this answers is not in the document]
AlexandreDoneux: Thanks, I wanted to make sure the way I implemented this was okay to you.

inginious/frontend/templates/mycourses.html
```diff
@@ -51,70 +51,154 @@ <h3>{{ _("Last tried exercises") }}</h3>
...
+                                    {% if course.lti_url() %}
+                                    <a class="hidden_pin_link" target="_blank" href="{{ course.lti_url() }}">
+                                        {% endif %}
+                                        {{course.get_name(user_manager.session_language())}}
+                                        {% if course.lti_url() %}
+                                    </a>
+                                    {% endif %}
+                                {% else %}
+                                    <a href="{{ get_path('course', courseid) }}">{{course.get_name(user_manager.session_language())}}</a>
+                                {% endif %}
```
anthonygego: As said previously, I would use the same link used in the list below, which is the internal one.

inginious/frontend/static/js/mycourses.js
```diff
@@ -0,0 +1,120 @@
...
+            if (click_button.find("i").hasClass("fa-star-o")) { // pinning
+
+                click_button.find("i").removeClass("fa-star-o").addClass("fa-star text-warning");
+
+                var pinnedItem = $("#hidden_pin_card").clone()
+                $(pinnedItem).attr("id", `pinned-${courseid}`);
+                $(pinnedItem).find(".course_pin").attr("data-course-id", courseid);
+
+                $(pinnedItem).find(".pin_title").append(data["name"]);
+                $(pinnedItem).find(".pin_link").attr("href", data["path"]);
```
anthonygego:
```suggestion
                $(pinnedItem).find(".pin_title").attr("href", data["path"]);
```

PR #1088, opened 2025-10-31
[cleanup 9] Database models

anthonygego:
This PR follows the previous ones and:

- Introduces models for every database collection : Audience, CourseClass, Group, LTI data, Session, Submission, User, UserTask.
- Replaces all raw calls to `pymongo` by `mongoengine` calls and model manipulation.
- Removes the reference to the `pymongo.Database` passed throughout the app.

It adds further data validation and constitutes a first step into an appropriate separation of models, logic and views.

Note: this is breaking for all plugins

### DIFF
--- a/inginious/frontend/app.py
+++ b/inginious/frontend/app.py
@@ -4,21 +4,16 @@
 # more information about the licensing of this file.
 
 """ Starts the webapp """
-import builtins
 import os
 import sys
 import flask
 import jinja2
-import pymongo
 import oauthlib
 
 from binascii import hexlify
-from pymongo import MongoClient
 from werkzeug.exceptions import InternalServerError
-from bson.codec_options import CodecOptions
-from mongoengine import connect
+from mongoengine import connect, disconnect
 
-import inginious.frontend.pages.preferences.utils as preferences_utils
 from inginious.frontend.environment_types import register_base_env_types
 from inginious.frontend.arch_helper import create_arch, start_asyncio_and_zmq
 from inginious.frontend.plugins import plugin_manager
@@ -39,6 +34,7 @@ from inginious.frontend.task_dispensers.combinatory_test import CombinatoryTest
 from inginious.frontend.flask.mapping import init_flask_mapping, init_flask_maintenance_mapping
 from inginious.frontend.flask.mongo_sessions import MongoDBSessionInterface
 from inginious.frontend.flask.mail import mail
+from inginious.frontend.models import DBVersion
 
 def _put_configuration_defaults(config):
     """
@@ -115,10 +111,10 @@ def get_path(*path_parts):
     return "/".join(path_parts)
 
 
-def _close_app(mongo_client, client):
+def _close_app(client):
     """ Ensures that the app is properly closed """
     client.close()
-    mongo_client.close()
+    disconnect()
 
 
 def get_app(config):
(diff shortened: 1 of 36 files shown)


inginious/frontend/models/submission.py
```diff
@@ -0,0 +1,50 @@
...
+
+class Submission(Document):
+    courseid = StringField(required=True)
+    taskid = StringField(required=True)
+    username = ListField(StringField(), required=True)
+    input = FileField(required=True)
+    archive = FileField()
+    status = StringField(required=True)
+    submitted_on = DateTimeField(required=True)
+    response_type = StringField(default='rst') # Deprecated
```
AlexandreDoneux: What is `response_type` used for and where ?

inginious/frontend/models/session.py
```diff
@@ -0,0 +1,53 @@
...
+from mongoengine import Document, StringField, ListField, EmbeddedDocument, EmbeddedDocumentField, BooleanField
+from mongoengine import DateTimeField, DynamicField, MapField, ObjectIdField
+
+
+class LTIData(EmbeddedDocument):
+    version = StringField(required=True, default="")
+    email = StringField(required=True, default="")
+    username = StringField(required=True, default="")
+    realname = StringField(required=True, default="")
+    roles = ListField(StringField())
```
AlexandreDoneux: Can this field also benefit from the choices parameter ?

inginious/frontend/app.py
```diff
@@ -156,12 +156,9 @@ def get_app(config):
...
 
     flask_app.config.from_mapping(**config)
-    flask_app.session_interface = MongoDBSessionInterface(
-        mongo_client, config.get('mongo_opt', {}).get('database', 'INGInious'),
-        "sessions", config.get('SESSION_USE_SIGNER', False), True  # config.get('SESSION_PERMANENT', True)
-    )
 
-    flask.request_finished.connect(UserManager._lti_session_save, flask_app)
+    # config.get('SESSION_PERMANENT', True)
+    flask_app.session_interface = MongoDBSessionInterface(config.get('SESSION_USE_SIGNER', False), True)
```
AlexandreDoneux: Any reason why you are not using the SESSION_PERMANENT given in config ?

inginious/frontend/user_manager.py
```diff
@@ -255,35 +234,21 @@ def set_session_code_indentation(self, code_indentation):
...
-        self._session.clear()
-        flask.g.lti_session_id = session_id
-        flask.g.lti_session = {}
-
-        self._session["lti"] = session_dict
-
+        self._session.loggedin = False
+        for key, item in session_dict.items():
+            self._session.lti[key] = item
         return session_id
```
AlexandreDoneux: Question : flask.session points to a mongodb document and we don't need to use .save() to save it's changes in database because it is done in the MongoDBSession.save_session() method. Am I understanding that correctly ?

AlexandreDoneux (changes requested):
Here is a first review. I still need to the lti parts to understand correctly how LTI works. I have some general comments : 
- You sometimes put the model's query parameters in .objects() and sometimes .get(). I indicated it with a comment. Is there a reason for that ? 
- Using the `choices` parameter in field declarations in models allows us to define a limited set of values that field can take. It would be interesting to use that in some cases (some of which I indicated with comments)
- We could also use custom validation for the model's fields using the `validate` parameter. Althought that might best be kept for another pull request.

inginious/frontend/user_manager.py
anthonygego: Yes. The changes are saved at the end of the request processing and kept in memory all the while. All access to the session object during the request processing are therefore coherent.

inginious/frontend/models/session.py
```diff
@@ -0,0 +1,53 @@
...
+
+class Session(Document):
+    id = ObjectIdField(primary_key=True, default=lambda: ObjectId()) # id should be available at creation time
+    permanent = BooleanField(required=True)
+    is_lti = BooleanField(required=True, default=False)
+    loggedin = BooleanField(required=True, default=False)
+    auth_storage = MapField(DynamicField(), default={})
+    expiration = DateTimeField()
+    lti = EmbeddedDocumentField(LTIData, default=lambda: LTIData())
+    code_indentation = StringField(default="4")
```
AlexandreDoneux: choices again

inginious/frontend/models/submission.py
```diff
@@ -0,0 +1,59 @@
...
+    grade = FloatField(default=0.0) # TODO: use min_value and max_value and change container API
+    custom = MapField(StringField())
+    problems = MapField(StringField())
+    result = StringField(default="crash") # TODO: restrict possible values and change container API
+    stderr  = StringField()
+    stdout  = StringField()
+    tests = MapField(StringField())
+    text = StringField(default="")
+    user_ip = StringField()
+    state = StringField()
```
AlexandreDoneux: What does this value represent ? Could we use `choices` ?

AlexandreDoneux (changes requested):
As you mentionned, the plugins are broken. But we will fix that in another PR. 

LGTM other than these last comments.

PR #1095, opened 2026-01-12
Make app-local redirects respect the url root

thaugdahl:
If you're hosting INGInious with a base-URL different from `/`, *e.g.,* `/inginious`,  some redirects don't respect the base URL for redirects that are meant to redirect to another page in INGInious.

For instance, the page file found at `frontend/pages/index.py` has a redirect `redirect("/courselist")`. This doesn't respect the base path (`/inginious` in our case). This means that you'll land on `/courselist` instead of `/inginious/courselist`. Our proxy thus returns a `404`.

This PR uses `get_path` to generate a base-URL respecting redirect:
https://github.com/INGInious/INGInious/blob/7591cff06832d7d9fb7eba50d7d1d59501a42a04/inginious/frontend/app.py#L102

This reflects the way redirects are handled in, e.g. https://github.com/INGInious/INGInious/blob/7591cff06832d7d9fb7eba50d7d1d59501a42a04/inginious/frontend/pages/lti/v1_3/__init__.py#L120

Example diff:

```diff
-            return redirect("/courselist")
+            return redirect(self.app.get_path("courselist"))
```

**Testing:**
I have done limited testing. The index page now redirects you correctly to `/inginious/courselist`. Otherwise, the login flow works and I can navigate the interface without trouble.  Nonetheless, I haven't done robust testing.

### DIFF
--- a/inginious/frontend/pages/index.py
+++ b/inginious/frontend/pages/index.py
@@ -14,11 +14,11 @@ class IndexPage(INGIniousStaticPage):
     def GET(self):  # pylint: disable=arguments-differ
         """ Display main course list page """
         if not self.app.welcome_page:
-            return redirect("/courselist")
+            return redirect(self.app.get_path("courselist"))
         return self.show_page(self.app.welcome_page)
 
     def POST(self):  # pylint: disable=arguments-differ
         """ Display main course list page """
         if not self.app.welcome_page:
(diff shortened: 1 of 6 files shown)


anthonygego:
Well spotted ! LTGM, thanks !

PR #1106, opened 2026-04-08
Adding debug for containers in IDEs

AlexandreDoneux:
This PR aims to add the possibility of interactive debug inside the containers running jobs.

Steps : 
- [x] Pycharm debug for the INGInious script and the scripts inside inginious_container_api (using the `pydevd_pycharm` python package)
- [ ] ~~Broader usage for all IDEs using the Debug Adapter Protocol (DAP)~~ => for future PR

Do you think debug capabilities for the run files is an interesting addition ?

### DIFF
--- a/base-containers/base/bin/INGInious
+++ b/base-containers/base/bin/INGInious
@@ -20,6 +20,16 @@ import asyncio
 import struct
 import zmq
 import zmq.asyncio
+import pydevd_pycharm
+
+
+pydevd_pycharm.settrace(
+    'host.docker.internal',
+    port=5678,
+    stdout_to_server=True,
+    stderr_to_server=True,
+    suspend=False
+)
 
 
 ELF_MAGIC = b"\x7F\x45\x4C\x46"
--- a/base-containers/base/bin/INGInious
+++ b/base-containers/base/bin/INGInious
@@ -23,13 +23,15 @@ import zmq.asyncio
 import pydevd_pycharm
 
 
-pydevd_pycharm.settrace(
-    'host.docker.internal',
-    port=5678,
-    stdout_to_server=True,
-    stderr_to_server=True,
-    suspend=False
-)
+if os.environ["DEBUGGER"] == "True":
+    print("Starting in debug mode, waiting for pycharm to connect...")
+    pydevd_pycharm.settrace(
+        'host.docker.internal',
+        port=5678,
+        stdout_to_server=True,
+        stderr_to_server=True,
+        suspend=False
+    )
 
 
 ELF_MAGIC = b"\x7F\x45\x4C\x46"
--- a/base-containers/base/bin/INGInious
+++ b/base-containers/base/bin/INGInious
@@ -25,13 +25,16 @@ import pydevd_pycharm
 
 if os.environ["DEBUGGER"] == "True":
     print("Starting in debug mode, waiting for pycharm to connect...")
-    pydevd_pycharm.settrace(
-        'host.docker.internal',
-        port=5678,
-        stdout_to_server=True,
-        stderr_to_server=True,
-        suspend=False
-    )
+    try:
+        pydevd_pycharm.settrace(
+            'host.docker.internal',
+            port=5678,
+            stdout_to_server=True,
+            stderr_to_server=True,
+            suspend=False
+        )
+    except Exception as e:
+        print("Failed to connect to pycharm debugger: {}".format(e))
 
 
 ELF_MAGIC = b"\x7F\x45\x4C\x46"
--- a/base-containers/base/bin/INGInious
+++ b/base-containers/base/bin/INGInious
@@ -35,7 +35,6 @@ if os.environ["DEBUGGER"] == "True":
             suspend=False
         )
     except Exception as e:
-        print("Failed to connect to pycharm debugger: {}".format(e))
         debug_server_unavailable = True
 
 
--- a/base-containers/base/bin/INGInious
+++ b/base-containers/base/bin/INGInious
@@ -23,6 +23,7 @@ import zmq.asyncio
 import pydevd_pycharm
 
 
+debug_server_unavailable = False
 if os.environ["DEBUGGER"] == "True":
     print("Starting in debug mode, waiting for pycharm to connect...")
     try:
@@ -35,6 +36,7 @@ if os.environ["DEBUGGER"] == "True":
         )
     except Exception as e:
         print("Failed to connect to pycharm debugger: {}".format(e))
+        debug_server_unavailable = True
 
 
 ELF_MAGIC = b"\x7F\x45\x4C\x46"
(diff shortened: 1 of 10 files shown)


inginious/agent/docker_agent/_docker_interface.py
```diff
@@ -132,6 +132,9 @@ def create_container(self, image, network_grading, mem_limit, task_path, sockets
         if ports is None:
             ports = {}
 
+        # Let the container listen on port 5678 for remote debugging
+        ports["5678/tcp"] = None
+
```
anthonygego:
```suggestion
```

The debug server is launched on the IDE host. You don't need to expose an additional port.

base-containers/base/bin/INGInious
```diff
@@ -20,6 +20,16 @@ import asyncio
...
+import pydevd_pycharm
+
+
+pydevd_pycharm.settrace(
+    'host.docker.internal',
+    port=5678,
+    stdout_to_server=True,
+    stderr_to_server=True,
+    suspend=False
+)
```
anthonygego: What happens if the debug server is not listening ?

anthonygego (changes requested):
Maybe we should enable this only on demand/when needed. Either passing an environment variable from PyCharm and frowarding it from the agent to the container, or rely on the existing debug flag. 

If this fails silently if the server is not listening, this could be OK, but I guess in this case it would triy and time out. This is not a wanted behaviour as this would severly impact the execution time

A nice addition (not for this PR) would be to allow the user to specify the hostname and port of the debug server and keep those as default. If the docker stack is located on another host, the `host-gateway` IP won't match the debug server one. Besides, the hardcoded port may not be available on the developer host if it is already used by another service.

AlexandreDoneux:
The existing debug_mode is defined for each job separately and is given by the BackendNewJob message to the docker agent. So it is best to use a new environment variable to indicate if debug through the IDE is wanted. Any suggestions for a good variable name so it isn't confused for a parameter for the remote ssh debug ?


**For a future PR for remote debugging :** 
As suggested we might want to use debug_host and debug_ports to specify the remote agent used and allow debug on the containers on that agent. However it is indicated that those configuration parameters are only to be used is the backend is on local. Does that mean we currently can't do remote ssh debugging if the backend is on a remote machine ? 

```yaml
# Only for backend: local
local-config:
    # debug_host: localhost  # host that will be indicated to users when they want to use the task remote debug feature. Automatically guessed by
                             # default
    debug_ports: 64000-64100 #port range use for remote debug feature
```

anthonygego:
> Any suggestions for a good variable name so it isn't confused for a parameter for the remote ssh debug ?

`ide_debug` or simply `debugger`

Actually it might be more productive to directly add the options in the webapp config file and the docker agent CLI arguments.

> Does that mean we currently can't do remote ssh debugging if the backend is on a remote machine ?

If the docker agent is used on another host, the args are passed in the CLI. This `local-config` feature is used for small deployments and development.

In conclusion, `debugger_host` and `debugger_port` seems two good new options to add to `local-config` and docker agent CLI.

base-containers/base/bin/INGInious
AlexandreDoneux: Should I raise an error in the UI if the debug server is unavailable ?

anthonygego: It's probably more convenient to be warned if the server couldn't be joined. However, it may not be necessary to warn in the GUI (that is, trough the feedback, which may complicate implementation). Maybe we can just put this in the stdout/stderr that is forwarded in the submission dict through the *debug information*.

AlexandreDoneux:
A message is now added to stderr if the debug server is not available. It does not fail the job anymore. 

I have also fixed the time limit for a job to half an hour if the container debugging is enabled. Without it the job just stops after the set limit and debugging is not possible. Should I add that fixed limit as a config parameter ?

inginious/agent/docker_agent/__init__.py
anthonygego: Can't we put `or self._debugger` here to avoid duplicating these lines ?

doc/teacher_doc/testing.rst
```diff
@@ -38,4 +38,12 @@ Debugging bash scripts (run.sh)
...
 environment so there is no need to use a particular loader. ``source run.sh`` might be a better
-choice than running the script in a subshell to observe the state of variables after execution.
\ No newline at end of file
+choice than running the script in a subshell to observe the state of variables after execution.
+
+Container debugging through an IDE
+------------------------------
+INGInious also supports container debugging through an IDE. This is currently only supported for the INGInious
+script and other scripts/APIs in the container by using Pycharm's debugging. You need to create a python debug server
+in Pycharm using the port 5678.
```
anthonygego:
```suggestion
script and other scripts/APIs in the container by using Pycharm's debugging. You need to create a *Python Debug Server*
in Pycharm Run/Debug configurations, listening on port 5678, and add one or several *Path Mapping* between the local 
and remote files to enable checkpoints.
```

doc/admin_doc/install_doc/config_reference.rst
```diff
@@ -53,6 +56,9 @@ The different entries are :
     ``tmp_dir``
         A directory whose absolute path must be available by the docker daemon and INGInious at the same time. By default, it is ``./agent_tmp``.
 
+    ``tmp_dir``
```
anthonygego:
```suggestion
    ``debugger``
```

doc/admin_doc/install_doc/config_reference.rst
```diff
@@ -53,6 +56,9 @@ The different entries are :
...
+        Indicates if container debugging is enabled. It is available only for Pycharm's debugging with a python debug server using port 5678. It is disabled by default.
```
anthonygego:
```suggestion
        Enables container debugging if set to ``true``. It is available only for Pycharm's debugging with a python debug server using port 5678. It is disabled by default.
```

base-containers/base/bin/INGInious
```diff
@@ -20,6 +20,23 @@ import asyncio
...
+        pydevd_pycharm.settrace(
+            'host.docker.internal',
+            port=5678,
+            stdout_to_server=True,
+            stderr_to_server=True,
+            suspend=False
+        )
+    except Exception as e:
+        print("Failed to connect to pycharm debugger: {}".format(e))
+        debug_server_unavailable = True
```
anthonygego, line 38: It seems it already prints a stack trace so maybe there already is a try/except behind this one which becomes unecessary. But not sure...

AlexandreDoneux: There is indeed a stack trace : 

```
Could not connect to host.docker.internal: 5678
  File "/usr/local/lib/python3.8/site-packages/_pydevd_bundle/pydevd_comm.py", line 443, in start_client
    s.connect((host, port))
ConnectionRefusedError: [Errno 111] Connection refused
```

But I don't get wan't you want me to do. Should I remove the print ?

anthonygego: There is no need for duplicate messages. Either catch the first exception if possible, or simply remove the second if it this not stop execution.

doc/teacher_doc/testing.rst
```diff
@@ -1,4 +1,4 @@
-Debugging tasks
+-Debugging tasks
```
anthonygego: Is this change wanted ?

anthonygego:
LGTM.

PR #1107, opened 2026-04-09
[frontend/js] refactor submission reload

anthonygego:
This refactors the way old submissions are reloaded so that the operations are the same both at student side (task route) and admin side (submission route).

- The old URL argument `?load=<submissionid>` is removed.
- `loadOldSubmissionInput(id, with_feedback=true)` is replaced by `loadOldSubmissionInput(id)` followed by `waitForSubmission(id)`. This allows to throw duplicate code `loadOldFeedback`.
- `waitSubmission` is modified so that the first try happens immediately. This speeds up feedback reloading for one possible additional call per submission, which is neglictable.

This PR also empties the file content of file problems before returning to the user. It may slow down submission reloading drastically if uploaded files were big. Besides, they are not used when the submission is reloaded. A link is provided instead to allow the student to download them if needed. 

This last commit is not ideal because it is not agnostic of the problem structure. However, the code for downloading these files is not agnostic either. This will be refactored later in another PR.

### DIFF
--- a/inginious/frontend/pages/tasks.py
+++ b/inginious/frontend/pages/tasks.py
@@ -349,6 +349,10 @@ class BaseTaskPage(object):
             tojson["status"] = 'ok'
             # And also include input
             tojson["input"] = data.get_input()
+            # filter-out values for files to avoid useless traffic
+            for key, input_data in tojson["input"].items():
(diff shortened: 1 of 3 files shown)


AlexandreDoneux:
LGTM

PR #1110, opened 2026-04-24
Adding possibility for jobs not linked to a task or course

AlexandreDoneux:
This PR modifies code to be able to handle container jobs that are not linked to a course and task. Meaning you don't need to provide a course and task id. 

This is not yet a fully implemented feature and will be followed by other PRs that allow unique jobs to be used for a reworked jplag plugin.


inginious/agent/docker_agent/__init__.py
```diff
@@ -297,31 +298,40 @@ def __new_job_sync(self, message: BackendNewJob, future_results):
...
             mem_limit = int(limits.get("memory", 200))
-            run_cmd = message.environment_parameters.get("run_cmd", '')
+            #run_cmd = message.environment_parameters.get("run_cmd", '') # changed from empty string to None somewhere for course tasks
+            run_cmd = message.environment_parameters.get("run_cmd", None) # when set to None, the run_cmd does not exist in the data received by the container
         except:
             raise CannotCreateJobException('The agent is unable to parse the parameters')
 
-        course_fs = self._fs.from_subfolder(course_id)
-        task_fs = course_fs.from_subfolder(task_id)
+        if message.course_id and message.task_id:
```
anthonygego:
```suggestion
        if course_id and task_id:
```

inginious/agent/docker_agent/__init__.py
```diff
@@ -297,31 +298,40 @@ def __new_job_sync(self, message: BackendNewJob, future_results):
             time_limit = int(limits.get("time", 30))
             hard_time_limit = int(limits.get("hard_time", None) or time_limit * 3)
             mem_limit = int(limits.get("memory", 200))
-            run_cmd = message.environment_parameters.get("run_cmd", '')
+            #run_cmd = message.environment_parameters.get("run_cmd", '') # changed from empty string to None somewhere for course tasks
```
anthonygego:
```suggestion
```

inginious/agent/docker_agent/__init__.py
```diff
@@ -297,31 +298,40 @@ def __new_job_sync(self, message: BackendNewJob, future_results):
...
+                raise CannotCreateJobException(
+                    'Task unavailable on agent. Please retry later, the agents should synchronize soon. '
+                    'If the error persists, please contact your course administrator.')
 
         # Check for realistic memory limit value
         if mem_limit < 20:
             mem_limit = 20
         elif mem_limit > self._max_memory_per_slot:
-            self._logger.warning("Task %s/%s ask for too much memory (%dMB)! Available: %dMB", course_id, task_id,
+            if message.course_id and message.task_id:
```
anthonygego:
```suggestion
            if course_id and task_id:
```

inginious/agent/docker_agent/__init__.py
```diff
@@ -297,31 +298,40 @@ def __new_job_sync(self, message: BackendNewJob, future_results):
...
                                  mem_limit, self._max_memory_per_slot)
+            else:
+                self._logger.warning("A job asks for too much memory (%dMB)! Available: %dMB", mem_limit,
+                                     self._max_memory_per_slot)
             raise CannotCreateJobException(
                 'Not enough memory on agent (available: %dMB). Please contact your course administrator.' % self._max_memory_per_slot)
 
         if environment_type not in self._containers or environment_name not in self._containers[environment_type]:
-            self._logger.warning("Task %s/%s ask for an unknown environment %s/%s", course_id, task_id,
+            if message.course_id and message.task_id:
```
anthonygego:
```suggestion
            if course_id and task_id:
```

inginious/agent/docker_agent/__init__.py
```diff
@@ -368,24 +378,28 @@ def __new_job_sync(self, message: BackendNewJob, future_results):
         os.chmod(sockets_path, 0o777)
         os.mkdir(course_path)
 
-        # TODO: avoid copy
-        task_fs.copy_from(None, task_path)
-        os.chmod(task_path, 0o777)
+        if message.course_id and message.task_id:
```
anthonygego:
```suggestion
        if course_id and task_id:
```

inginious/agent/docker_agent/__init__.py
```diff
@@ -859,7 +873,7 @@ async def close_and_delete(student_container_id=student_container_id_loop):
                     except:
                         pass  # ignore
 
-                self._create_safe_task(close_and_delete(student_container_id_loop))
+                # self._create_safe_task(close_and_delete(student_container_id_loop))
```
anthonygego: why ?

inginious/agent/docker_agent/__init__.py
```diff
@@ -938,7 +954,8 @@ async def close_and_delete(student_container_id=student_container_id_loop):
 
             # Remove container
             try:
-                await self._docker.remove_container(container_id)
+                #await self._docker.remove_container(container_id)
+                pass
```
anthonygego: why ?

inginious/agent/__init__.py
```diff
@@ -166,34 +166,38 @@ async def __handle_new_job(self, message: BackendNewJob):
...
-            task_fs = self._fs.from_subfolder(message.course_id).from_subfolder(message.task_id)
-            if not task_fs.exists():
-                self._logger.warning("Task %s/%s unavailable on this agent", message.course_id, message.task_id)
-                raise CannotCreateJobException('Task unavailable on agent. Please retry later, the agents should synchronize soon. If the error '
-                                               'persists, please contact your course administrator.')
-
-            # Let the subclass run the job
+        if message.course_id is None and message.task_id is None:
+            # job not linked to a course/task
             await self.new_job(message)
```
anthonygego, line 172: You can probably return after the await here. This avoids a long indented else clause with no code after.

inginious/backend/backend.py
```diff
@@ -171,7 +171,7 @@ async def handle_client_get_queue(self, client_addr, _: ClientGetQueue):
         for job_id, content in self._job_running.items():
             agent_friendly_name = self._registered_agents[content.agent_addr].name
             jobs_running.append((content.msg.job_id, content.client_addr == client_addr, agent_friendly_name,
-                                 content.msg.course_id+"/"+content.msg.task_id,
+                                 content.msg.course_id+"/"+content.msg.task_id if content.msg.course_id and content.msg.task_id else "non task job",
```
anthonygego: In a first iteration, it would be more useful to have at least the environment name as hint.

inginious/client/client.py
```diff
@@ -269,7 +269,7 @@ def get_available_environments(self) -> Dict[str, List[str]]:
         """
         return self._available_environments
 
-    def new_job(self, priority, course, task, inputdata, callback, launcher_name="Unknown", debug=False, ssh_callback=None):
+    def new_job(self, priority, course, task, inputdata, callback, job_image=None, launcher_name="Unknown", debug=False, ssh_callback=None):
```
anthonygego: I don't see any case where we would want `job_image` to be `None`.

inginious/client/client.py
```diff
@@ -296,30 +298,50 @@ def new_job(self, priority, course, task, inputdata, callback, launcher_name="Un
...
+
+            ssh_callback = lambda _1, _2, _3, _4: None
+
+
+            msg = ClientNewJob(job_id, priority, courseid, taskid, problem_dict, inputdata,
+                               environment_type, environment, environment_parameters, debug, launcher_name)
+            self._loop.call_soon_threadsafe(asyncio.ensure_future,
+                                            self._create_transaction(msg, task=None , callback=safe_callback,
+                                                                     ssh_callback=ssh_callback))
+
```
anthonygego: Could be rewritten more nicely without repetition.

inginious/client/client.py
```diff
@@ -296,30 +298,50 @@ def new_job(self, priority, course, task, inputdata, callback, launcher_name="Un
...
+                               environment_type, environment, environment_parameters, debug, launcher_name)
+            self._loop.call_soon_threadsafe(asyncio.ensure_future,
+                                            self._create_transaction(msg, task=task, callback=safe_callback,
+                                                                     ssh_callback=ssh_callback))
+
+        else:
+            courseid, taskid = None, None
+            # testing values
+            environment_type = "docker"
+            environment = job_image if job_image else "inginious/default" # find another default, better way ?
```
anthonygego: There should be no default.

inginious/client/client.py
```diff
@@ -296,30 +298,50 @@ def new_job(self, priority, course, task, inputdata, callback, launcher_name="Un
...
+
+            environment_parameters = task.get_environment_parameters()
+
+            msg = ClientNewJob(job_id, priority, course.get_id(), task.get_id(), task.get_problems_dict(), inputdata,
+                               environment_type, environment, environment_parameters, debug, launcher_name)
+            self._loop.call_soon_threadsafe(asyncio.ensure_future,
+                                            self._create_transaction(msg, task=task, callback=safe_callback,
+                                                                     ssh_callback=ssh_callback))
+
+        else:
```
anthonygego, line 330: The whole `new_job` signature should be refactored so that it is agnostic of what we're sending as a job. Passing the environment either from the `task` argument or the `job_image` is not clean and confusing.

Maybe the submission manager should unpack the required task data first

anthonygego (changes requested):
This PR is not very clean. You should read your diff before submitting it.

Globally it seems good, but I'd rather rewrite the `client.new_job` method so that it becomes agnostic of whether we are launching a task or not. I'd modify this so that submission_manager` unpacks the task data and call this method with the appropriate arguments, and that a plugin or another component would call the this method using its own environment definition and inputdata.

This way the signature would be less confusing and the code would not need to handle different cases with default code.

inginious/agent/docker_agent/__init__.py
AlexandreDoneux: I forgot to uncomment some lines of code. The point was to be able to access the containers after their job was done.

inginious/client/client.py
AlexandreDoneux: How's this ? I am not entirely sure that is the way to go. Or at least it needs additional changes. 

The environment is now defined in the submission manager and the task, course, environment type and environment are passed to `new_job()` through à `job_info` dict. I have added a `add_unique_job()` function.  Do you think it is a good idea or should the submission manager only handle asynchronous student submission ? Can we imagine a similar method placed somewhere else or should we call all unique jobs through `Client.new_job()` and `ClientSync.new_job()` ?

AlexandreDoneux: Now explaining it I thing the second proposition is better but I could use your opinion.

AlexandreDoneux: Discussed in person

inginious/frontend/submission_manager.py
```diff
@@ -185,8 +185,9 @@ def replay_job(self, course, task, submission, task_dispenser, copy=False, debug
 
         # Don't enable ssh debug
         ssh_callback = lambda host, port, user, password: self._handle_ssh_callback(submissionid, host, port, user, password)
+        job_info = {"course": course, "task": task, "environment_type": task.get_environment_type(), "environment": task.get_environment()}
```
anthonygego:
```suggestion
        job_info = {"course": course, "task": task, "environment_type": task.get_environment_type(), "environment": task.get_environment_id()}
```
I suppose... This fails directly and there is no scenario in which you can submit without facing the issue. Please perform a minimal functionality check before submitting for review.

AlexandreDoneux:
Sorry for the errors that slipped in the code. I worked on another branch on multiple features at once before importing the code I thought was relevant. It is clearly not the most efficient/foolproof way of working. 

I fixed those and ran a courseless/taskless job by calling a synchronous job. However this is not yet possible in the current state of INGInious. Do you have in mind a way for you or other reviewers to check the feature is working as expected.

I also had another question : 
- While creating a `ClientNewJob` message in client.py, should the task problems dictionary passed be a `None` or an empty dict ? Both work opposed to the environment parameters that have to be a dict.

anthonygego:
> * While creating a `ClientNewJob` message in client.py, should the task problems dictionary passed be a `None` or an empty dict ? Both work opposed to the environment parameters that have to be a dict.

It's probably safer to default to an empty dictionary and maybe enforce it. The `getinput` API at the end is not meant to handle a `None` object.

inginious/client/client.py
```diff
@@ -296,31 +297,38 @@ def new_job(self, priority, course, task, inputdata, callback, launcher_name="Un
         job_id = str(uuid.uuid4())
         safe_callback = _callable_once(callback)
 
+        course = job_info["course"]
+        task = job_info["task"]
```
anthonygego: This forces `course` and `task` keys to be present in the dict. I'd rather default to `None` and/or explicitly document that these keys are expected (and besides document the other needed keys `environment` and `environment_type`) instead of simply:
>a dict containing all the needed information about the job 

Otherwise lgtm.

anthonygego:
Well well well, I realize by testing the code that empty `/task` and `/course/common` folders are still mounted. This may become confuse as if you put your `run` file in `/task/` or `/course/common/` it won't be accessible.

There is a choice to be made here:
- Either not mounting those useless folders but in this case it's worth making sure they exists in the base container to avoid breaking the other scripts. I'm not sure how much it impacts existing code.
- Either documenting it. I think it would be OK at this stage.

In both cases it's worth adding docs under a *Developers documentation/Extensions/Taskless jobs* section. This should specify the valid paths for the entrypoints and how to use the client.

doc/dev_doc/extensions_doc/taskless_jobs.rst
```diff
@@ -0,0 +1,12 @@
+Taskless jobs
+=======
+
+INGInious can handle taskless jobs. That means jobs that are not linked to a task or course.
+They can be used to run background jobs without needing to create a course.
+You can run a taskless job by using the `new_job()` methode from the `Client` or `ClientSync` class, depending on if you
```
anthonygego:
```suggestion
You can run a taskless job by using the `new_job()` method from the `Client` or `ClientSync` class, depending on if you
```

anthonygego: Can you provide a short demo code ? This can be as a demo plugin that simply run an Hello World task.

anthonygego (changes requested):
- The code fails in case the jobs must be put in waiting queue:

      2026-06-17 13:51:49,965 - inginious.backend - ERROR - An exception occurred while running a Task
        File "/xxxxxx/INGInious/inginious/backend/backend.py", line 179, in handle_client_get_queue
          jobs_waiting = [(job.job_id, job.client_addr == client_addr, job.msg.course_id+"/"+job.msg.task_id, job.msg.launcher,
                                                                       ~~~~~~~~~~~~~~~~~^~~~
      TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'

  How to reproduce: just remove the environment from your container images.
- There is still no mechanism that prevents taskless environments to be listed as grading environments.
  Either :
    - Use `org.inginious.taskless.name` label and do not advertise these tags up to the webapp
    - Use a boolean `org.inginious.grading.advertise` label to explicitly advertise an environment. This may allow using an environment as both grading and taskless environment.

anthonygego (changes requested):
The key `advertised` does not exist for the MCQ agent environment. A default value should be set.

      File "/xxx/INGInious/inginious/backend/backend.py", line 282, in handle_agent_hello
        env_dict[name] = EnvironmentInfo(environment_info["id"], environment_info["created"], [agent_addr], environment_type, environment_info["advertised"])
                                                                                                                              ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
    KeyError: 'advertised'

anthonygego:
LGTM

PR #1111, opened 2026-04-30
Add JPLAG docker environment

AlexandreDoneux:
This PR adds a JPLAG Docker environment for later implementation of JPLAG analyses as jobs. This environment takes a zip files containing directories containing files that will be compared by JPLAG. It decompresses it and runs the JPLAG jar.  

JPLAG by default opens a GUI when run to display the results. Here, however, JPLAG only generates the compressed file (.jplag) that can be read afterwards with the same JPLAG jar. [See JPLAG documentation](https://github.com/jplag/JPlag/wiki/7.-Report-Viewer)

To use it outside the INGInious architecture you override the the INGInious script started by the container and replace it with a simple execution of the run script. The result file is saved on the container at `/jplag/result/results.jplag`. Run this docker command to run it. Here we have defined python3 as the language used for the plagiarism detection. A lot of languages are accepted ([see documentation](https://github.com/jplag/JPlag)). 

```sh
docker run  -v "$(pwd)/test_submissions.zip":/jplag/input/submissions.zip   
-e JPLAG_LANGUAGE=python3     
--entrypoint /bin/bash   
jplag:latest   
-c "python3 /course/run.py"
```

### DIFF
--- a/inginious/frontend/plugins/jplag/container/Dockerfile
+++ b/inginious/frontend/plugins/jplag/container/Dockerfile
@@ -4,14 +4,8 @@
 # Create a .zip file containing one directory per student, each containing the student's submission. The .zip file should be named "submissions.zip".
 # Then, run the container with the following command :
 #
-#docker run -v "$(pwd)/test_submissions.zip":/jplag/input/submissions.zip -e JPLAG_LANGUAGE=python3 --entrypoint python3 jplag:latest
-# -c "
-# import sys
-# from unittest.mock import MagicMock
-# sys.modules['inginious_container_api'] = MagicMock()
-# sys.modules['inginious_container_api.feedback'] = MagicMock()
-# exec(open('/course/run.py').read())
-# "
+# docker run  -v "$(pwd)/test_submissions.zip":/jplag/input/submissions.zip -e JPLAG_LANGUAGE=python3
+# --entrypoint /bin/bash  jplag:latest  -c "python3 /course/run.py"
 #
 # The mock is necessary to avoid import errors, as the container expects to be run in an INGInious environment.
 
--- a/inginious/frontend/plugins/jplag/container/Dockerfile
+++ b/inginious/frontend/plugins/jplag/container/Dockerfile
@@ -1,0 +1,37 @@
+# DOCKER-VERSION 1.1.0
+
+# Environment variables for the JPLAG environment
+# JPLAG_LANGUAGE (required) : The programming language to be used for the plagiarism detection. Supported values are available at https://github.com/jplag/JPlag. Be aware of the version used.
+# SIM_THRESHOLD : The similarity threshold for the plagiarism detection. Default is 0 (0% similarity).
+
+
+ARG   VERSION=latest
+ARG   REGISTRY=ghcr.io
+FROM  ${REGISTRY}/inginious/env-base:${VERSION}
+
+LABEL org.opencontainers.image.source=https://github.com/INGInious/containers
+LABEL org.opencontainers.image.description="Single test JPLAG job environment"
+LABEL org.inginious.grading.name="jplag"
+
+
+# Install Java SE 21
+RUN dnf clean metadata && \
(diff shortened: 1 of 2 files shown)


inginious/frontend/plugins/jplag/container/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+# The mock is necessary to avoid import errors, as the container expects to be run in an INGInious environment.
+
+
+ARG   VERSION=latest
+ARG   REGISTRY=ghcr.io
+FROM  ${REGISTRY}/inginious/env-base:${VERSION}
+
+LABEL org.opencontainers.image.source=https://github.com/INGInious/containers
+LABEL org.opencontainers.image.description="Single test JPLAG job environment"
+LABEL org.inginious.grading.name="jplag"
```
anthonygego, line 19: Put in comment that we should replace/replace this in another PR because this kind of environment is not meant to be advertised as a grading environment. Maybe a opt-out "advertise" label can be a good implementation for this.

inginious/frontend/plugins/jplag/container/run.py
```diff
@@ -0,0 +1,97 @@
...
+#   │   ├── Main.java
+#   │   └── util/Helper.java
+#   └── charlie/
+#       └── answer.py
+#
+
+if not os.path.isfile(ZIP_PATH):
+    print(f"ERROR: submissions zip not found at '{ZIP_PATH}'.")
+    feedback.set_global_result("failed")
+    raise SystemExit(1)
```
anthonygego, line 50: I don't know what kind of LLM model you are using but it should be worth tweaking it or replacing it. 

This is not idiomatic Python nor correct excception-driven development as, in the latter case, exceptions are made to be caught/handled. 

This definition is used internally to kill the Python interpeter. You should use `exit` to do things properly.

inginious/frontend/plugins/jplag/container/Dockerfile
```diff
@@ -0,0 +1,41 @@
+# DOCKER-VERSION 1.1.0
+
+# usage outside of INGInious :
+# Create a .zip file containing one directory per student, each containing the student's submission. The .zip file should be named "submissions.zip".
+# Then, run the container with the following command :
+#
+# docker run  -v "$(pwd)/test_submissions.zip":/jplag/input/submissions.zip -e JPLAG_LANGUAGE=python3
+# --entrypoint /bin/bash  jplag:latest  -c "python3 /course/run.py"
```
anthonygego, line 8: I expected the code to run as a typical submission. In the case of a typical submission no volume is mounted, input is transferred via the agent internal messages. Mounting volumes on remote agents in not a bad idea on paper, but would require substantial changes in the agent code a it requires sharing the docker stack on the network.

If you want to support both, please add a decent documentation on how end users should use this, and/or provide a wrapper such as the ones in `utils/scripts`.

anthonygego (changes requested):
I'm not sure if this has to be considered finished or not. It is not clear whether you want to handle the container stack directly or pass through the agent code. In the latter case, I'd expect some input fetching code. The presence of a call to the feedback API in the snippets prvided is very confusing about how it is supposed to be used now and/or how it's going to evolve.

In all cases, I would keep this an external feature. Please migrate the internal JPlag plugin to the INGInious Github organization first on work on this repo.

The main repository should be later changed to allow these kinds of environment not to be advertised as grading environments.

AlexandreDoneux:
As discussed, this PR was a way to show a first draft for the JPLAG environment. It was implemented in a way to test outside the INGInious architecture as it depends on this [PR](https://github.com/INGInious/INGInious/pull/1110). 

A new [repository](https://github.com/INGInious/plugin-jplag) has been created to store the JLAG plugin outside the main architecture. I have moved the existing plugin and webserver from gitlb and moved all already existing code from this PR there.

anthonygego:
I'm closing this for now.

PR #1114, opened 2026-05-07
refactor: normalize code style and formatting across repo

remivanboxem:
- Added `Ruff` dependency and rules inside `pyproject.toml`
- Format the entire project using `ruff --config pyproject.toml format inginious/` command.

This change allows having a clearer codebase and should not add regression. To enforce this stylistic change, maintainers can either:
- refuse any PRs or commits that doesn't respect this style
- add a GitHub action workflow that will automatically reformat any incoming PR.


anthonygego:
The changes proposed sometimes seem way less readable than the previous lines due to an abusive use of new lines. Is this really the recommended standard ?

In any case I'm not sure we would need to update the entire codebase once because it would break any work made above: our work, but also community work that we want to respect even if those changes don't come back to upstream. 

I think this is something we could simply set for future PRs. Time would do the necessary and replace the old code. However I'd like another thought on this (@nrybowski) because I dont really facny having a lot of new lines everywhere.

nrybowski:
Agree with @anthonygego, such cleanup will appear naturally while we refactor the code. The issue is more deep than just cosmetic adjustments, e.g., some function may need refactor to accept dataclasses as parameters rather than very long lists of arguments. There is also a lot of duplicated code that needs to be merged in functions, and so on.
Just re-formatting the code will not improve the codebase quality.

PR #1115, opened 2026-06-02
[agent_docker] appropriate selinux labelling

anthonygego:
It seems our README has mentioned the code works with SELinux for 11 years without being the case. 

This patch enables SELinux labelling on the mounted volumes. Since a temporary folder is created for each submission, I decided to apply a strict labelling (`Z`) and associates two random MCS categories (as the container runtime does by default) to all containers with the same parent (of the same submission). This allows to share the Unix socket created in the parent container with the student containers.

This last operation is done by using the parent sockets path as the seed for the random function to avoid keeping more state in memory and a more important refactor.

This is probably not perfect, but still better than deactivating SELinux to allow the system to work.

This patch was tested on a system with SELinux disabled at boot time and the options seem to be ignored, so it should be the case on distros that are not shipped with SELinux without further checks.

### DIFF
--- a/inginious/agent/docker_agent/_docker_interface.py
+++ b/inginious/agent/docker_agent/_docker_interface.py
@@ -7,7 +7,7 @@
     (not asyncio) Interface to Docker
 """
 import os
-from datetime import datetime
+import random
 from typing import List, Tuple, Dict
 
 import docker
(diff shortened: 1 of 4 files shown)


inginious/agent/docker_agent/_docker_interface.py
```diff
@@ -274,3 +277,10 @@ def list_runtimes(self) -> Dict[str, str]:
         :return: dict of runtime: path_to_runtime
         """
         return {name: x["path"] for name, x in self._docker.info()["Runtimes"].items()}
+
+    def _get_security_opts(self, seed: str) -> str:
+        """
+        :return: SELinux MCS label based on the given seed
+        """
+        c1, c2 = random.Random(seed).sample(range(1024), 2)
```
nrybowski, line 286: Will change this in favor of the `secrets` API to make codacy happy.

nrybowski (commented):
> This allows to share the Unix socket created in the parent container with the student containers.

In the future, we will leverage the `IPC` capability of OCI containers to share common memory between containers. This will avoid to explicitly mount the Unix socket between the containers through the host filesystem.

Also, we will only expose single Unix socket per grading container rather than one per run_student process.

Otherwise, LGTM for now. I still will rebase on main and give it a test run.

inginious/agent/docker_agent/_docker_interface.py
nrybowski: The seed is required to make sure that the grading container and student containers derive the same labels. The `secrets` API does not allow seeding. These random values are not cryptographically relevant, let's keep that for now.

nrybowski:
For now, I have to disable SELinux labeling for the `agent-docker` so that it can access the mounted docker socket:

```console
$ git diff docker-compose.yml
diff --git a/docker-compose.yml b/docker-compose.yml
index f413883b..bba12c18 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -73,6 +73,8 @@ services:
         - REGISTRY=${REGISTRY}
     environment:
       BACKEND: "tcp://backend:2001"
+    security_opt:
+      - "label=disable"
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock:z
       # See https://github.com/INGInious/INGInious/issues/352
```
Otherwise SELinux bans the access to the socket: `SELinux is preventing /usr/bin/python3.12 from connectto access on the unix_stream_socket /run/docker.sock.`

nrybowski:
Rebase on 2585840e54b7fedf71916beda7a7929472ca4b78.

nrybowski (approved):
Tested on fresh RL10 install, a run_student container successfully runs. Merging.

PR #1116, opened 2026-06-04
[agent_docker] add support for cgroups v2

anthonygego:
This PR follows #1115 and adds full support for cgroups v2. It preserves the cgroups v1 support as this is still the default in some RHEL versions.

- The Docker interface file is modified to detect the cgroups version return by the Docker API
- With cgroups v1, the OOM Killer is disabled. OOM is caught in Docker events, and the container is then killed by INGInious.
- With cgroups v2, the container checks if `memory.oom.group` set to 1 to ensure all the container tasks are killed by the OOM killer once one it runs out of memory, since the OOM killer cannot be disabled anymore. This results in the container being killed by the system and INGInious checking if it died from OOM.
- The documentation is updated and now explains how to ensure systemd set the  `memory.oom.group` flag to 1 for containers spawned by the Docker daemon.

This full support for cgroups v2 allows to runs the code with rootless podman socket set as `DOCKER_HOST`. Since podman does not support enumerating container runtimes from the API and that podman-py does not implement container attach yet, I suggest to keep Docker as the recommended runtime for now, yet I mention it is compatible in the doc.

### DIFF
--- a/base-containers/base/Dockerfile
+++ b/base-containers/base/Dockerfile
@@ -7,7 +7,7 @@ ENV LC_ALL en_US.UTF-8
 
 LABEL org.opencontainers.image.source=https://github.com/INGInious/containers
 LABEL org.opencontainers.image.description="Base INGInious grading environment."
-LABEL org.inginious.grading.agent_version=3
+LABEL org.inginious.grading.agent_version=4
 
 ARG PYTHON_VERSION=3.14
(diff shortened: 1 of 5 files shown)


nrybowski:
Rebase on 564e11067f13bc092f7e157ec45db2cd9c094c88.

nrybowski (approved):
Tested on RL10, works as expected. Merging.

PR #1118, opened 2026-06-29
deploy: Make the compose deployment more configurable and production ready

### LINKED ISSUE #317: FastCGI detection breaks Gunicorn and other WSGI servers
These lines in inginious-webapp do not work properly:
```python
if 'SERVER_SOFTWARE' in os.environ:  # cgi
    os.environ['FCGI_FORCE_CGI'] = 'Y'

if 'PHP_FCGI_CHILDREN' in os.environ or 'SERVER_SOFTWARE' in os.environ:  # lighttpd fastcgi
    print(os.environ)
    import flup.server.fcgi as flups
    flups.WSGIServer(application, multiplexed=True, bindAddress=None, debug=False).run()
```

Gunicorn is a WSGI server but defines `SERVER_SOFTWARE`, and a such we start the flup FCGI server.

nrybowski:
This PR attempts to make the docker-compose deployment more user-friendly by adding configuration knobs as environment variables.

#### TODO
- [x] Add Caddy reverse proxy
- [x] Add documentation
- [x] Expose webdav


base-containers/base/Dockerfile
nrybowski: Should fix python version.

deploy/init-mongo.js
```diff
@@ -0,0 +1,30 @@
+let password =  process.env.INGINIOUS_SUPERADMIN_PASSWORD;
```
nrybowski, line 1: Should enforce argon2id encoding.

base-containers/base/Dockerfile
nrybowski: Fixed in https://github.com/INGInious/INGInious/pull/1118/commits/21755553172ab74ecb1c707bc5be4b89403781c2

nrybowski:
Depends on #1135 .

docker-compose.yml
```diff
@@ -28,7 +34,32 @@ services:
...
+    volumes:
+      - ./deploy/init-mongo.js:/docker-entrypoint-initdb.d/init-mongo.js
+      - ./mongo:/data
+    environment:
+      MONGO_INITDB_DATABASE: "INGInious"
+
+      # MANDATORY ARGUMENT, INITIALISATION WILL FAIL IF NOT PROVIDED.
+      # Output of 'echo -n "superadmin" | argon2  ASUPERSECURESALT -id -t 3 -m 16 -l 64 -e'.
+      # $ are doubled because of variable expansion in compose env var.
+      INGINIOUS_SUPERADMIN_PASSWORD: '$$argon2id$$v=19$$m=65536,t=3,p=1$$QVNVUEVSU0VDVVJFU0FMVA$$NpeSDEsx+FwrIS277T9o/Q4z4Ypicibs3LiLZrftXsBUmBxFnK8FKa45w5q85ewM+dOlzZPMFdAmxpmDVAmLlQ'
```
anthonygego, line 47: Can't we think of generating the password and priting it in the logs at creation time in case it is not specified ?

I'm also wondering if that wouldn't be more convenient to put the environment variables in a separate file.

nrybowski: > Can't we think of generating the password and priting it in the logs at creation time in case it is not specified ?

We could if required, setting the default password to `superadmin` is the historical method for the example deployment of INGInious.

> I'm also wondering if that wouldn't be more convenient to put the environment variables in a separate file.

Setting a .env file is another classical way to define environment variables for docker compose but since we do not expose many variables, I think that keeping everything in a single file is more self-explanatory. Instance administrators are free to customize their compose file at their will.

nrybowski:
> Depends on #1135 .

Now that #1135 has been merged, we should rebase and retest the webdav deployment.

nrybowski:
Rebase on 48f2b05a08741d37be4041e86f29ab7847a7bcd3.

nrybowski:
Removed c68b7ae5f38f329461d8644a7ff462d5df7954f4 as it was already applied on main.

docker-compose.yml
```diff
@@ -142,5 +165,36 @@ services:
     networks:
       - inginious
 
+  webdav:
```
anthonygego, line 156: this, as the webapp, depends on the database container too, to fetch the user webdav password and check for course access.

deploy/Caddyfile
```diff
@@ -0,0 +1,8 @@
+# Replace me with dedicated FQDN.
+inginious.local localhost {
```
anthonygego, line 2: I'm ok starting with this in this PR. But we should move to WSGI to qualify this as production ready and not rely on the werkzeug server.

deploy/init-mongo.js
```diff
@@ -0,0 +1,30 @@
...
+let email = process.env.INGINIOUS_SUPERADMIN_EMAIL;
+email = email != undefined ? email : "superadmin@inginious.org";
+
+let lang = process.env.INGINIOUS_SUPERADMIN_LANG;
+lang = lang != undefined ? lang : "en";
+
+let indent = process.env.INGINIOUS_SUPERADMIN_INDENT;
+indent = indent != undefined ? indent : "4";
+
+db.users.insertOne({
```
anthonygego, line 22: This script is meant to replace the `db_setup.py` file if I understand well. It should also check if there is no duplicate then.

inginious/frontend/webdav.py
```diff
@@ -199,19 +202,29 @@ def get_app(config):
     """ Init the webdav app """
     connect(config.get('database', 'INGInious'), host=config.get('mongo_opt', {}).get('host', 'localhost'), tz_aware=True)
 
+    register_base_env_types()
+    register_problem_types(get_default_displayable_problem_types())
```
anthonygego, line 208: This is needed and is enough in the context of this PR. 

However, it does not handle plugins. It makes me feel a bit unconfortable because I was pretty sure having changed the code not to load tasks descriptors (because this is actually not needed) and bypass this kind of issues.

configuration.deploy.yaml
```diff
@@ -13,3 +13,4 @@ superadmins:
 - superadmin
 tasks_directory: /inginious/tasks
 use_minified_js: true
+webdav_host: https://localhost/dav
```
anthonygego, line 16: This is the value that will be shown in the UI. If this has to be set manually, I'd put a line in the docs to put attention on this.

docker-compose.yml
anthonygego: OK let's keep this as is for now.

anthonygego (changes requested):

AlexandreDoneux:
Regarding your docker compose. The backup directory is depreciated and can be removed from the volumes. Also, the task directory path is hardcoded and should use the value from the configuration.

I was also thinking about an easy way to make the instance only accessible on localhost.

AlexandreDoneux:
Due to the mongodb version upgrade in the docker compose, pre-existing persisting data in the db directory raises an error and stops the db container shortly after starting. Maybe add something to the `init-mongo.js` to make the data compatible for new version.

Here is the related error in the container's logs : 

> db-1         | {"t":{"$date":"2026-08-12T09:55:58.581+00:00"},"s":"F",  "c":"CONTROL",  "id":20573,   "ctx":"initandlisten","msg":"Wrong mongod version","attr":{"error":"UPGRADE PROBLEM: Found an invalid featureCompatibilityVersion document (ERROR: Location4926900: Invalid featureCompatibilityVersion document in admin.system.version: { _id: \"featureCompatibilityVersion\", version: \"6.0\" }. See https://docs.mongodb.com/master/release-notes/8.0-compatibility/#feature-compatibility. :: caused by :: Invalid feature compatibility version value '6.0'. Expected one of the following versions: ''8.3', '8.2', '8.0'. See https://docs.mongodb.com/master/release-notes/8.0-compatibility/#feature-compatibility.). If the current featureCompatibilityVersion is below 8.0, see the documentation on upgrading at https://docs.mongodb.com/master/release-notes/8.0/#upgrade-procedures."}}

nrybowski:
> Regarding your docker compose. The backup directory is depreciated and can be removed from the volumes. Also, the task directory path is hardcoded and should use the value from the configuration.

~The deprecation of the backup directory has been lost in the previous PRs on the compose file. Will reapply.~

Fixed in e41f3ba26ddeff3e9571a30e8b9ac4a9db5bdbbf.

deploy/init-mongo.js
nrybowski: The advantage of using the built-in initialisation method of mongo is that the engine ensures by itself that this script is ran only once, so we do not have to check ourselves.
That being said, we should suppress the `db_setup.py` script in this PR.

configuration.deploy.yaml
nrybowski: It must reflect the configuration of the reverse proxy. I will document that.

docker-compose.yml
nrybowski: Fixed in 1e6485f7b5a0eb79a4341bbb8172464198138978

deploy/Caddyfile
nrybowski: Webdav is served by Gunicorn in 1e6485f7b5a0eb79a4341bbb8172464198138978.
Still have to also serve the frontend with it.

inginious/frontend/webdav.py
nrybowski: Additional comments on this one?

deploy/init-mongo.js
nrybowski: Fixed in e41f3ba26ddeff3e9571a30e8b9ac4a9db5bdbbf

deploy/Caddyfile
nrybowski: Frontend is served with Gunicorn in 5ea6bbdbd61fadb0e3865d386fa1d49d75a2c616.

nrybowski:
@AlexandreDoneux I suggest bumping the mongod version in another PR.

inginious/scripts/webapp.py
nrybowski, line 83: @anthonygego I'm not familiar with that fix, is it still used or can we deprecate it since it is not required to run behind Caddy?

nrybowski: After discussion, we will keep that for now as it does not impact current compose-based deployment.

configuration.deploy.yaml
nrybowski: Done in d95cdf03301c8d3f749479881ae2083457cc5519.

deploy/init-mongo.js
anthonygego, line 22: I'll provide a more elaborate review later but while I'm thinking of it: this script should probably also create the index in the `users` collection. 

I don't think mongoengine will ensure it if the collection already exist. Maybe this can be checked without investing more effort.

BTW I realize the mongoengine model does not specify an index on the `email` key yet.

inginious/scripts/webapp.py
```diff
@@ -50,16 +38,35 @@ def main():
 
     # Init logging
     init_logging(config.get('log_level', 'INFO'))
-    logging.getLogger("inginious.webapp").info("http://%s:%d/" % (host, int(port)))
 
     application, close_app_func = inginious.frontend.app.get_app(config)
 
     if 'SERVER_SOFTWARE' in os.environ:  # cgi
         os.environ['FCGI_FORCE_CGI'] = 'Y'
```
anthonygego: This can be deleted too if we throw away fastcgi support

deploy/Containerfile
```diff
@@ -87,3 +86,20 @@ RUN --mount=source=.git,target=.git,type=bind \
...
+COPY inginious/frontend inginious/frontend/
+COPY inginious/client inginious/client/
+COPY inginious/scripts inginious/scripts
+
+# See https://github.com/pypa/setuptools_scm/#usage-from-docker
+RUN --mount=source=.git,target=.git,type=bind \
+    pip"${PYTHON_VERSION}" install --no-cache-dir -e . &&\
+    pip"${PYTHON_VERSION}" install 'gunicorn==26.0.0'
+
+CMD ["sh", "-c", "python${PYTHON_VERSION} -m gunicorn.app.wsgiapp 'inginious.scripts.webdav:get_app()' --bind 0.0.0.0 --workers 1"]
```
anthonygego: I don't know what Github did with my previous comment. So there are scripts in `frontend/wsgi` that were useful for that. Historically, both have been moved to the `inginious` package due to constraint with pyporject.toml.

I don't remeber exactly why they were put apart. If that works, they can be merged again. However, mod_wsgi requires `application` to be available in the module.

This can also be done in a separate PR and rely on the `frontend/wsgi`  in this one.

nrybowski: The modifications in `frontend/scripts/web{app,dav}.py` are basically the same as the content in `frontend/wsgi`. We could maybe remove `frontend/wsgi` to avoid duplicating code?

anthonygego: Yes, just make sure `mod_wsgi` can do `from inginious.scripts.web{app,dav} import application`.

inginious/frontend/webdav.py
anthonygego: Two possibles solutions:
- Keep this code and perform problem discovery from plugins (that is : https://github.com/INGInious/INGInious/blob/efb62f88783d32d29b739a7e4db42239d8331a8f/inginious/frontend/plugins/plugin_manager.py#L70-L76 )
- Adapt `inginious.frontend.task_dispensers.toc.TableOfContents` code not to load the tasks when it is initialized. This would be the preferred (and most efficient) way. But this doesn't seem trivial right now. We can go on with the solution above and change this later.

anthonygego (changes requested):

nrybowski:
This PR will fix #317 by deprecating the faulty lines.

inginious/frontend/webdav.py
nrybowski: We will keep the current solution for now and document the fact that webdav requires interpreting the tasks, which is maybe not mandatory. Modifying that behavior can be implemented in subsequent PRs.

deploy/init-mongo.js
nrybowski: Fixed with 9fa375cf05737aee08340b53879797861b3597c0.

deploy/Containerfile
nrybowski: Fixed in 9ac1772670c83ab8771bcb970252299dfd081061

inginious/scripts/webapp.py
nrybowski: Done with 9ac1772670c83ab8771bcb970252299dfd081061

docker-compose.yml
```diff
@@ -129,18 +150,50 @@ services:
     volumes:
       - ./configuration.deploy.yaml:/inginious/configuration.yaml:z
       - ./tasks:/inginious/tasks:z
-      - ./backups:/inginious/backups:z
     ports:
       - 9000:8080
```
anthonygego: I guess this port publishing is not useful anymore

inginious/frontend/webdav.py
anthonygego: Still to do.

anthonygego (changes requested):

docker-compose.yml
nrybowski: I propose to keep it for debugging purpose if one wants to bypass the caddy proxy. We could document that.

anthonygego: But is there something listening ?

nrybowski: The port has changed since we serve with Gunicorn now https://github.com/nrybowski/INGInious/blob/b93a2fbcbcfef1ee129e7dc0cb17e836dbe405aa/deploy/Caddyfile#L4. I will just remove these lines since the compose method aims for production ready deployments.

nrybowski: Fixed in d059df56270ebe5b342b28afb9ecddff9dda57b5

inginious/frontend/webdav.py
nrybowski: Fixed in df4404e5b4337827817e25d30224aee400d8234f

nrybowski:
Rebased on 707f4297c9433d80dd2f214cde74eded8b5586f3

deploy/init-mongo.js
nrybowski: Left for later PR.

inginious/scripts/webdav.py
```diff
@@ -52,24 +41,36 @@ def main():
...
+    parser.add_argument("--port", help="Port to listen to. Default is 8080.", type=int, default=os.environ.get("INGINIOUS_WEBDAV_PORT", "8080"))
+    args = parser.parse_args()
 
-    # Ensure WsgiDAV receive limited streams for PUT requests
-    application = limited_input_middleware(application)
+    host = args.host
+    port = args.port
+    configfile = args.config
+
+    application = get_app(configfile)
```
anthonygego: We loose the parsed `config` dict reference here.  Same for `webapp.py` file.

Probably the most simple thing to do is to make `get_app` return a tuple `config, application`, retrieve it here and below at the end of file, and change the gunicorn line by `inginious.scripts.webdav:application`.

inginious/scripts/webapp.py
anthonygego: We also loose this reference here.

anthonygego (changes requested):

anthonygego:
For the future it is probably interesting to deprecate `inginious-webapp` and `inginious-webdav` as standalone script and rely entirely on gunicorn, since it is quite a convenient command.

nrybowski:
> For the future it is probably interesting to deprecate `inginious-webapp` and `inginious-webdav` as standalone script and rely entirely on gunicorn, since it is quite a convenient command.

Documented in #1158.

inginious/scripts/webapp.py
nrybowski: Fixed in d2ddb91a75c8fcd21a8457f73a109132ec159537.

inginious/scripts/webdav.py
nrybowski: Fixed in afb34b3c00defd7ccaeeddd961123f85429e4e56.

anthonygego (approved):

PR #1121, opened 2026-07-09
Token-authenticated API

AlexandreDoneux:
This PR adds two endpoints to retrieve submissions using a token and a way for a user to generate it.

The token can be generated from the user preferences, in the page "API token". It is implemented using a JWT, allowing an automatic expiration.

The endpoints : 
- `/api/v0/token/courses/<courseid>/submissions`
   - Returns the submissions of a course, and the data related to them.
- `/api/v0/token/courses/<courseid>/<taskid>/submissions`
   - Returns the submissions of a particular task

Both are POST endpoint due to the use of parameters passed through a JSON body. Passing them as arguments or parameters through the url could become cumbersome.
That body can contain : 
- select: "all" (default), "best", "last" : select all submissions, the best submission per student, or the last submission per student
- username: a list of usernames to filter the submissions. If none is provided (or it is empty), the submissions for all users are returned
- format: "json" (default), "csv" : format of the response.

Yet to bee implemented : 
- [x] Removing test endpoint
- [x] Add csv formatting for the endpoint's response
- [x] Add the JWT lifetime and secret in the webapp configuration
- [x] Work on api token page in the UI, multiple tokens with invalidation, token only visible once at generation
- [x] Store hash of token 
- [x] store token secret and lifetime in config
- [x] Change endpoint names
- [x] Complete doc


inginious/frontend/pages/api/_api_page.py
```diff
@@ -112,6 +114,38 @@ def _verify_authentication(self, handler, args, kwargs):
...
+        """ Verify that the given token is valid """
+
+
+        auth_header = flask.request.headers.get("Authorization", "")
+        if not auth_header.startswith("Bearer "):
+            raise APIForbidden("Missing or malformed Authorization header")
+        token = auth_header.removeprefix("Bearer ").strip()
+
+        JWT_SECRET = "your-secret-key"
+        JWT_ALGORITHM = "HS256"
```
anthonygego: This should be put in the webapp config

inginious/frontend/pages/preferences/apitoken.py
```diff
@@ -0,0 +1,65 @@
...
+
+from inginious.frontend.pages.utils import INGIniousAuthPage
+from inginious.frontend.models import User
+
+
+# test values to move to config file or configure in api token page (lifetime)
+# one secret key per user ? If the secret key is found out, every user is at risk.
+JWT_SECRET = "your-secret-key"
+JWT_ALGORITHM = "HS256"
+TOKEN_LIFETIME = datetime.timedelta(days=365)  # Token lifetime of 1 year
```
anthonygego: This should be put in the webapp config, and absolutely not hardcoded and repeated in files

inginious/frontend/pages/api/_api_page.py
```diff
@@ -112,6 +114,38 @@ def _verify_authentication(self, handler, args, kwargs):
         return handler(*args, **kwargs)
 
 
+class APITokenAuthPage(APIAuthenticatedPage):
```
anthonygego: Why do you inherit from ``APIAuthenticatedPage`` and override the ``_verify_authentication`` method ? The goal is to replace the session-based authentication that is not relevant anymore.

inginious/frontend/pages/api/_api_page.py
```diff
@@ -112,6 +114,38 @@ def _verify_authentication(self, handler, args, kwargs):
...
+        JWT_ALGORITHM = "HS256"
+
+        try:
+            payload = jwt.decode(token, JWT_SECRET, algorithms=[JWT_ALGORITHM])
+        except jwt.ExpiredSignatureError:
+            raise APIForbidden("Your token has expired, please generate a new one.")
+        except jwt.InvalidTokenError:
+            raise APIForbidden("Invalid token, please generate a new one.")
+
+        self.user = User.objects(username=payload["username"]).first()
```
anthonygego: It's probably a good practice to check against the database too. 

This is a good security practice. Imagine the secret key is lost (or discovered hardcoded in the public codebase...). Hackers can forge new tokens themselves, making the leak even more dangerous as everyone will be concerned.

By the way, you need to allow invalidating tokens before they expire.

inginious/frontend/templates/apitoken.html
```diff
@@ -0,0 +1,61 @@
...
+    </div>
+{% endfor %}
+
+<div class="col-sm-12">
+    <p>
+        {{ _("Your API token allows external applications and scripts to authenticate as you and interact with "
+             "INGInious on your behalf, without needing your password. Keep it secret: anyone who has it can "
+             "access the API using your account permissions. If you believe your token has been compromised, "
+             "generate a new one below to invalidate the old one immediately. See the INGInious documentation "
+             "for usage.") }}
```
anthonygego, line 37: - The token is more or less a clear password with an expiration date that you store somewhere in automated scripts.
- This means you ideally need to provide a token per external source of authentication and not an only one.

This constraint can be imposed in a first implementation as this is not so critical but rather a convenience issue. However, there should be a way for a user to simply invalidate its curent token without generating a new one.

inginious/frontend/pages/preferences/apitoken.py
```diff
@@ -0,0 +1,65 @@
...
+class APITokenPage(INGIniousAuthPage):
+    """ Page to view or generate an API token for the user """
+
+    def GET_AUTH(self):
+        """ GET request """
+
+        user = User.objects(username=session["username"]).first()
+
+        try:
+            payload = jwt.decode(user.apitoken, JWT_SECRET, algorithms=[JWT_ALGORITHM])
```
anthonygego: This, here, is a critical security issue and **should** be addressed. 

I'm not sure people store tokens in their clear form in database. They rather show the clear token once it is generated and then keep a hash and expiration date in database, leaving the users on their own if they lose it.

As said before tokens may be seen as clear passwords. If there is a database leak or unauthorized access for a reason or another, the tokens can be used freely to act as another user. 

Of course you can invalidate tokens once you know they have been leaked. But you need to know that. Leaks can be orchestrated by "legitimate" people having access to database backups for instance.

inginious/frontend/pages/api/submissions.py
```diff
@@ -214,3 +219,215 @@ def API_POST(self, courseid, taskid):  # pylint: disable=arguments-differ
...
+            Returns 403 Forbidden if the user does not have access to the course/task.
+            Returns 404 Not Found if the course/task does not exist.
+        """
+
+
+        # Does APITokenAuthPage automatically checks for APIForbidden? -> normally yes,
+
+        # TODO : get user from token, for now we use a hardcoded username
+
+        return 200, {"message": "Submissions endpoint is reachable and user has access."}
```
anthonygego: What is the goal of this ?

anthonygego (changes requested):
I did a first pass of things that shoud be addressed before going further.

inginious/frontend/pages/api/_api_page.py
AlexandreDoneux: I check against the database right under this bit of code.  An `InvalidTokenError` is raised when the token couldn't be decoded. Meaning I'm checking if the jwt has the right format. 
The tokens can be invalidated by the user by generating a new token that will replace the current one.

anthonygego: > I check against the database right under this bit of code. 

OK : 

https://github.com/AlexandreDoneux/INGInious/blob/bb84a142f72a554f6ad3583a2cc6d81c704aa868/inginious/frontend/pages/api/_api_page.py#L144

This should be against a hash though.

inginious/frontend/pages/preferences/apitoken.py
AlexandreDoneux: No worries, this was not going to stay there. Should we also allow the choice of the algorithm ?

inginious/frontend/pages/api/_api_page.py
AlexandreDoneux: Not sure I understood that remark. I overrode the `_verify_authentication` method to replace the session-based authentication.

anthonygego: Yes so the old method should not exist anymore.

inginious/frontend/pages/preferences/apitoken.py
anthonygego: Good remark. I don't think this is a strict requirement. However, we may need to change it in the future and therefore it may be useful to support an old algorithm for not expired tokens, or simply to indicate they are not valid anymore. This means in both cases the algorithm should be stored with the hash, as this is the case for passwords.

AlexandreDoneux: If the algorithm used changes, an error will be sent back during an API call due to `jwt.decode()` raising an `InvalidTokenError`. It does not display which token exactly needs to be regenerated but I believe it should be the user's responsability to correctly use the token descriptions to know which one they used.

anthonygego: Then they should at least be able to know why it has become invalid if the user interface still show them as not expired, otherwise this becomes confusing whether they did something bad or it was a bug or even implementation choice.

Probably not necessary to plan a migration mechanism now. But in this case the algorithm should not be parameterizable to avoid any issue. Just choose something that is state of the art for now. The problem is similiar to the password one. It can be solved later using the same techniques.

AlexandreDoneux:
I have addressed your current remarks. I have on my end other questions I would like to submit to you : 
- endpoint names : I have for the moment used `/api/v0/token` in the endpoints I have added. However since we don't use the flask session anymore all endpoints use th token authentication and this distinction is not needed anymore. I will change that. But do you want to change the version of the api ?
- POST input : I used the JSON input format for the endpoints I added. Should we stick to the form format ? It is necessary for the endpoint to add a submission for a file problem as JSON input does not support files. I therefore added both formats for that endpoint. But it might be better to stick to one format.
- I kept a HS256 symmetric algorithm for JWT as it is simple and the INGInious "backend" is the only one encoding and decoding the JWTs. Is i enough or should we use an asymmetric algorithm ?

anthonygego:
> * endpoint names : I have for the moment used `/api/v0/token` in the endpoints I have added. However since we don't use the flask session anymore all endpoints use th token authentication and this distinction is not needed anymore. I will change that. But do you want to change the version of the api ?

Yes, I think we can, even if the interface doesn't change. That will invalide any old usage of the API. No need to preserve any retrocompatibility, the thing was broken for years.

> POST input : I used the JSON input format for the endpoints I added. Should we stick to the form format ? 

No JSON's probably good.

> It is necessary for the endpoint to add a submission for a file problem as JSON input does not support files. I therefore added both formats for that endpoint. But it might be better to stick to one format.

This is not something bad I guess. This practice can be found in other software. For instance, in Grist, the endpoint for creating a new document supports the two body schema, json and multipart : https://support.getgrist.com/api/#tag/docs/operation/importDoc

The main requirement remains that the API is usable through standard librairies and methods.

> I kept a HS256 symmetric algorithm for JWT as it is simple and the INGInious "backend" is the only one encoding and decoding the JWTs. Is i enough or should we use an asymmetric algorithm ?

I guess that's enough, indeed, as long as the symmetric algorithm is robust enough for a few years.

inginious/frontend/pages/api/submissions.py
```diff
@@ -46,7 +51,7 @@ def _get_submissions(submission_manager, user_manager, courseid, taskid, with_in
     for submission in submissions:
         submission = submission_manager.get_feedback_from_submission(
             submission,
-            show_everything=user_manager.has_staff_rights_on_course(course, session.username)
+            show_everything=user_manager.has_staff_rights_on_course(course, username) #TODO : use username from token.
```
anthonygego: I guess that this TODO is not relevant anymore ?

inginious/frontend/pages/api/submissions.py
```diff
@@ -188,19 +196,23 @@ def API_POST(self, courseid, taskid):  # pylint: disable=arguments-differ
...
+                    user_input[pid] = flask.request.form.getlist(pid)
+                elif problem.input_type() == dict:
+                    user_input[pid] = flask.request.files.get(pid)
+                else:
+                    user_input[pid] = flask.request.form.get(pid)
 
         user_input = task.adapt_input_for_backend(user_input)
 
-        if not task.input_is_consistent(user_input, current_app.config('ALLOWED_FILE_EXTENSIONS'),
+        if not task.input_is_consistent(user_input, current_app.config.get('ALLOWED_FILE_EXTENSIONS'),
```
anthonygego:
```suggestion
        if not task.input_is_consistent(user_input, current_app.config['ALLOWED_FILE_EXTENSIONS'],
```

All the fields should be present in the config, so let the app crash if this is not the case rather than facing a possible `None` value.

inginious/frontend/pages/api/submissions.py
```diff
@@ -210,7 +222,204 @@ def API_POST(self, courseid, taskid):  # pylint: disable=arguments-differ
...
+            return {
+                "courseid": s.courseid,
+                "taskid": s.taskid,
+                "username": s.username,
+                "submitted_on": s.submitted_on.strftime("%Y-%m-%dT%H:%M:%SZ"),
+                "result": s.result,
+                "grade": s.grade,
+                "stderr": s.stderr,
+                "stdout": s.stdout,
+                "input": json.loads(json_util.dumps(s.get_input())),
```
anthonygego: What is the exact purpose of this ?

inginious/frontend/pages/api/submissions.py
```diff
@@ -210,7 +222,204 @@ def API_POST(self, courseid, taskid):  # pylint: disable=arguments-differ
...
+        if response_format == "csv":
+            return 200, self._to_csv_response(submissions_list)
+
+        return 200, submissions_list
+
+
+    # send back text feedback ? -> text sent back to the student after submitting its work, not the feedback sent back by the grader
+    # send back submission id ?
+
+    # filter on date range ?
```
anthonygego: Do not let thoughts in the code

inginious/frontend/pages/api/submissions.py
```diff
@@ -210,7 +222,204 @@ def API_POST(self, courseid, taskid):  # pylint: disable=arguments-differ
...
+                "username": s.username,
+                "submitted_on": s.submitted_on.strftime("%Y-%m-%dT%H:%M:%SZ"),
+                "result": s.result,
+                "grade": s.grade,
+                "stderr": s.stderr,
+                "stdout": s.stdout,
+                "input": json.loads(json_util.dumps(s.get_input())),
+            }
+
+        submissions_list = [serialize(s) for s in submissions]
```
anthonygego: Here you store the whole list of submissions with their input in memory. 

This may become very hazardous in some cases. Submissions, with their input, may weigh 50M in some courses where dataset are requested. That's why inputs are stored in the gridfs store. It is not uncommon to process 1000 submissions. In this case, you already fill 50G of memory.

If you need to process the submission inputs, that should be done iteratively, so that the memory usage remains stable whatever the amount of submissions to process is.

inginious/frontend/pages/api/submissions.py
```diff
@@ -210,7 +222,204 @@ def API_POST(self, courseid, taskid):  # pylint: disable=arguments-differ
...
+        # Collect the union of all "input" keys across submissions, to build stable CSV columns.
+        input_fields = []
+        seen_input_fields = set()
+        for s in submissions_list:
+            for key in s.get("input", {}).keys():
+                if key not in seen_input_fields:
+                    seen_input_fields.add(key)
+                    input_fields.append(key)
+
+        fieldnames = base_fields + ["input." + f for f in input_fields]
```
anthonygego: I'm not fully comfortable with the idea of flattening the submission inputs in CSV files :

- The CSV file is a text file that is generally meant to be human readable where each line corresponds to an entry. 
   - Our most common inputs are multiline codes and binary files.
   - The subproblem architecture allows many other kind of complex data (blockly, network traces, parsons) that doesn't fit in CSV either.
- There is no sense in flattening the submissions input for different tasks. There even is a very high chance of collision between keys.
- Submissions can weigh up to 50M (and there actually is no limit, this is a very local observation)

It's probably more sane to provide an endpoint to download the input files of a batch of submissions. This may require more code at the client side for parsing inputs, but input parsing should be done in a subproblem context, as the GUI does.

I'd like the thought of @nrybowski on this.

inginious/frontend/pages/api/tasks.py
```diff
@@ -80,16 +82,16 @@ def API_GET(self, courseid, taskid):  # pylint: disable=arguments-differ
...
+            task_cache = self.user_manager.get_task_cache(username, course.get_id(), task.get_id())
 
             data = {
                 "id": taskid,
-                "name": task.get_name(session.language),
-                "authors": task.get_authors(session.language),
-                "contact_url": task.get_contact_url(session.language),
+                "name": task.get_name("en"),
+                "authors": task.get_authors("en"),
+                "contact_url": task.get_contact_url("en"),
```
anthonygego: The user language can be found in database. There is no reason to provide localisable fields in english if the user is authenticated. 

An alternative valid option is to provide a way for the user to ask for a specific localisation.

inginious/frontend/models/__init__.py
```diff
@@ -13,7 +13,7 @@
 from inginious.frontend.models.lti1_3 import LTIGrade, LaunchData
 from inginious.frontend.models.session import Session, LTIData
 from inginious.frontend.models.submission import Submission
-from inginious.frontend.models.user import User
+from inginious.frontend.models.user import User, APIToken
```
anthonygego, line 16: Is there something missing with this import ?

inginious/frontend/models/user.py
```diff
@@ -5,8 +5,14 @@
 
 import tzlocal
 
-from mongoengine import Document,  StringField, ListField, MapField, BooleanField, DynamicField
+from mongoengine import Document,  StringField, ListField, MapField, BooleanField, DynamicField, EmbeddedDocument, EmbeddedDocumentField, DateTimeField, EmbeddedDocumentField
```
anthonygego:
```suggestion
from mongoengine import Document,  StringField, ListField, MapField, BooleanField, DynamicField, EmbeddedDocument, EmbeddedDocumentField, DateTimeField
```

inginious/frontend/pages/api/_api_page.py
```diff
@@ -98,17 +101,36 @@ def API_OPTIONS(self, *args, **kwargs):  # pylint: disable=unused-argument
...
+        """ Verify that the given token is valid """
+
+
+        auth_header = flask.request.headers.get("Authorization", "")
+        if not auth_header.startswith("Bearer "):
+            raise APIForbidden("Missing or malformed Authorization header")
+        token = auth_header.removeprefix("Bearer ").strip()
+
+        API_JWT_SECRET = current_app.config.get('API_JWT_SECRET')
+        API_JWT_ALGORITHM = current_app.config.get('API_JWT_ALGORITHM')
```
anthonygego: All caps are not something pythonic. Flask do an exception with its config dict. I'd rather keep that style for accessing the config dict only.

inginious/frontend/models/user.py
```diff
@@ -19,6 +25,7 @@ class User(Document):
     ltibindings = MapField(MapField(DynamicField())) # TODO: use custom validation or refactor
     tos_accepted = BooleanField(default=False)
     apikey = StringField(default=None)
+    apitokens = ListField(EmbeddedDocumentField(APIToken), default=[])
```
anthonygego: Looking at the rest of the code, I wonder if a `MapField` won't be a better choice to avoid iterating on them. Besides, there should not be two identical tokens per user and a dictionary ensures this intrinsically.

inginious/frontend/pages/preferences/apitoken.py
```diff
@@ -0,0 +1,87 @@
...
+            token_dict[token.token] = token
+
+        return self.show_page()
+
+    def POST_AUTH(self):
+        """ POST request, generates a new token for the user """
+
+        API_JWT_SECRET = current_app.config.get('API_JWT_SECRET')
+        API_JWT_ALGORITHM = current_app.config.get('API_JWT_ALGORITHM')
+        API_JWT_LIFETIME = datetime.timedelta(days=current_app.config.get('API_JWT_LIFETIME'))
```
anthonygego: Same regarding all caps.

inginious/frontend/pages/preferences/apitoken.py
```diff
@@ -0,0 +1,87 @@
...
+
+            return self.show_page(generated_token=token)
+
+        # save the token and description in the database
+        if "save" in request.form:
+            generated_token = request.form.get("generated_token")
+            description = request.form.get("description")
+
+            if description == "" :
+                return self.show_page(errors=["Description is required to save the token."])
```
anthonygego: You'll probably have more chance to ask for the description before generating it. In all cases, the clear token should never be fed back for storage by the user, even if there is a validity check. In this case, there is even none.

An alternative is to accept a new description once it is generated.

inginious/frontend/templates/apitoken.html
```diff
@@ -0,0 +1,114 @@
...
+                    <p class="card-text ">
+                        {% if token.expires < token.expires.now().astimezone() %}
+                            <small class="text-danger ml-1">{{ _("Expired on:") }} {{ show_time(token.expires) }}</small>
+                        {% else %}
+                            <small class="text-muted ml-1">{{ _("Expires on:") }} {{ show_time(token.expires) }}</small>
+                        {% endif %}
+                    </p>
+                </div>
+                <form method="post" action="{{ url_for('.apitokenpage') }}" >
+                    <input type="hidden" name="token" value="{{ token.token }}">
```
anthonygego: As for sending clear tokens in form for storage, it's a good practice to reduce sending hashes through http requests. Because this remains a valuable data for hackers when the algorithm becomes poor.

It's probably useful here to add a decorelated id to the token so that we know how to update its description without exposing the hash on the internet.

@nrybowski what do you think ?

inginious/frontend/templates/apitoken.html
```diff
@@ -0,0 +1,114 @@
...
+                        <button type="submit" class=" btn btn-danger btn-block ">
+                            <i class="fa fa-close"></i> {{ _("Close") }}
+                        </button>
+                    </form>
+                </div>
+            </div>
+        </form>
+    </div>
+</div>
+{%else %}
```
anthonygego: See other comments on token creation and description updates.

inginious/frontend/submission_manager.py
```diff
@@ -101,14 +101,14 @@ def _before_submission_insertion(self, course, task, inputdata, debug, obj):
...
+        if "username" not in obj:
+            username = session.username
+
         is_group_task =course.get_task_dispenser().get_group_submission(task.get_id())
 
         if is_group_task and not self._user_manager.has_staff_rights_on_course(course, username):
             group = Group.objects.get(courseid=course.id, students=username)
             obj.update({"username": group["students"]})
-        else:
-            obj.update({"username": [username]})
```
anthonygego, line 111: Why this deletion ?

inginious/frontend/submission_manager.py
```diff
@@ -226,10 +226,11 @@ def add_job(self, course, task, inputdata, task_dispenser, debug=False):
         :type debug: bool or string
         :returns: the new submission id and the removed submission id
         """
-        if not session.loggedin:
-            raise Exception("A user must be logged in to submit an object")
-
-        username = session.username
+        if username is None:
+            if not session.loggedin:
+                raise Exception("A user must be logged in to get his submissions")
```
anthonygego: If this is not a heavy change, it would be useful to move this check at the function caller side. The less the submission manager relies on the flask context, the more we'll be able to convert it in a nice, testable controller.

inginious/frontend/pages/api/_api_page.py
```diff
@@ -98,17 +101,36 @@ def API_OPTIONS(self, *args, **kwargs):  # pylint: disable=unused-argument
...
+
+        try:
+            payload = jwt.decode(token, API_JWT_SECRET, algorithms=[API_JWT_ALGORITHM])
+        except jwt.ExpiredSignatureError:
+            raise APIForbidden("Your token has expired, please generate a new one.")
+        except jwt.InvalidTokenError:
+            raise APIForbidden("Invalid token. It is not correctly formatted.")
+
+        self.user = User.objects(username=payload["username"]).first()
```
anthonygego: I don't think Flask instantiate a new instance of the class for each request. So here you have a potential user concurrency issue.

The secure possibilities are :
- keeping the value on the stack: making the parent call the children class methods with the fetched user from database, or make the children class call the parent method to fetch the user.
- using Flask global, which is a global storage accessible through the app tha exist in the context and lifetime of a request.

anthonygego (changes requested):
In summary, there are 3 things that need attention here:
- The way the user is propagated between parent and children API classes.
- Tokens and hashes sent in post forms.
- The way submission input are served.

inginious/frontend/pages/api/submissions.py
AlexandreDoneux: The `serialize` function allows the format change of `submitted_on` and `input` of the submissions retrieved in the database. Would you prefer I do it in a loop without the function ?

anthonygego: I mean...dumping the dict into json before loading it again. There should be a reason.

inginious/frontend/pages/api/submissions.py
AlexandreDoneux: I wasn't comfortable with the idea either but it was a request from [redacted: a client] although it seemed strange.

anthonygego: Clients aren't always right, and clients requests aren't always realistic. You should stick to what your software architecture allows you to. Or refactor it in order to make it flexible.

And then only provides the realistic solution to the client that fits most or all their needs.

The most appropriate solution given the current architecture and such a request is flattening (and therefore generating the CSV file) at the API client side. For the client, that remains the same: the final product is the one expected.

inginious/frontend/pages/api/tasks.py
AlexandreDoneux: I feel like there is no real use for allowing users to choose the language through the endpoint call. There might not even be a translation available. I'll just use the user's language.

inginious/frontend/models/__init__.py
AlexandreDoneux: Not sure I understand the question. I import it to be able to import it into `apitoken.py`

anthonygego: Right, not obvious from the diff.

inginious/frontend/pages/api/submissions.py
AlexandreDoneux: Right, sorry. I do all this to get rid of the byte values that are  not serializable by json. Although, I must admit it is not really clear. Maybe I should move that to the `get_input()` method or do it differently.

Originally, `.get_input()` provides a python dict of the inputs of a submission. And the API handler transforms that dict into a string with `json.dumps()`. However, if an input is a file submission, the dict contains bytes that are not serializable into a JSON.

So, the different stages of this new version : 
1. The `.get_input()` model method uses `bson.BSON.decode()` to transform the BSON from mongodb into a python dict (as previously).
2. The `json_util.dumps()` method transforms it into a JSON string that handles unsupported JSON types.
3. The `json.loads()` transforms it back into a python dict. Now without the unsupported types.
4. `json.dumps()` transforms it back into a string before sending the response.

anthonygego: > The json.loads() transforms it back into a python dict. Now without the unsupported types.

So what happens to them ? Are they silently omitted ?

Besides, this process is very heavy. The only valid solution is to provide the endpoint to download the submission input in the native format and process it at the client side, using the bson library if needed.

AlexandreDoneux: No, they are transformed into supported types.

How would you prefer that ? I could send the submission id and a list of the problems in the task with their type. The user would then call another endpoint where he provides the submission id and the id of the problem he wants the input of.

anthonygego: > No, they are transformed into supported types.
If this is actually the case, the appropriate way is to do it only when it is necessary, by iterating on items.

> The user would then call another endpoint where he provides the submission id and the id of the problem he wants the input of.

An endpoint that serves the input file is enough. Further processing should be done at client side.

inginious/frontend/flask/mapping.py
```diff
@@ -143,22 +140,22 @@ def init_flask_mapping(flask_app):
                            view_func=CourseDangerZonePage.as_view('coursedangerzonepage'))
     flask_app.add_url_rule('/admin/<courseid>/stats',
                            view_func=CourseStatisticsPage.as_view('coursestatisticspage'))
-    flask_app.add_url_rule('/api/v0/auth_methods',
-                           view_func=APIAuthMethods.as_view('apiauthmethods'))
-    flask_app.add_url_rule('/api/v0/authentication',
-                           view_func=APIAuthentication.as_view('apiauthentication'))
-    flask_app.add_url_rule('/api/v0/courses', view_func=APICourses.as_view('apicourses.alias'),
+    flask_app.add_url_rule('/api/v1/courses', view_func=APICourses.as_view('apicourses.alias'),
```
nrybowski, line 143: Bumping the API version is a breaking change that is not even documented in the PR description. Is the bump mandatory? If yes, document it extensively and maybe add dummy endpoints for the old version. That is the kind of silent breaking change we really do not want in production.

inginious/frontend/pages/api/_api_page.py
```diff
@@ -98,17 +101,36 @@ def API_OPTIONS(self, *args, **kwargs):  # pylint: disable=unused-argument
...
+        """ Verify that the given token is valid """
+
+
+        auth_header = flask.request.headers.get("Authorization", "")
+        if not auth_header.startswith("Bearer "):
+            raise APIForbidden("Missing or malformed Authorization header")
+        token = auth_header.removeprefix("Bearer ").strip()
+
+        api_jwt_secret = current_app.config.get('API_JWT_SECRET')
+        api_jwt_algorithm = current_app.config.get('API_JWT_ALGORITHM')
```
nrybowski: So the entirety of the stored tokens is silently invalidated in case of configuration change? This is not a desired behavior at all. Either, the admin can configure the preferred hashing algorithm for **new** tokens, either we hardcode the algorithm. Same comment for the secret.

inginious/frontend/pages/api/_api_page.py
```diff
@@ -98,17 +101,36 @@ def API_OPTIONS(self, *args, **kwargs):  # pylint: disable=unused-argument
...
+            payload = jwt.decode(token, api_jwt_secret, algorithms=[api_jwt_algorithm])
+        except jwt.ExpiredSignatureError:
+            raise APIForbidden("Your token has expired, please generate a new one.")
+        except jwt.InvalidTokenError:
+            raise APIForbidden("Invalid token. It is not correctly formatted.")
+
+        flask.g.user = User.objects(username=payload["username"]).first()
+
+        if not any(UserManager.verify_hash(api_token["token"], token) for api_token in flask.g.user.apitokens) :
+            raise APIForbidden("Invalid token. It is correctly formatted but does not belong to the user in the JWT.")
```
nrybowski: Why not store a mapping of token hash and validity information rather than just a list? It replaces an iteration by a hit or miss operation.
The current approach does not allow an easy integration for future dedicated token authorizations.
Also, DoS vector if there is no limit on the number of tokens a user can generate.

inginious/frontend/pages/api/submissions.py
```diff
@@ -63,10 +66,10 @@ def _get_submissions(submission_manager, user_manager, courseid, taskid, with_in
...
 
         if submission["status"] == "done":
-            data["grade"] = submission.get("grade", 0)
-            data["result"] = submission.get("result", "crash")
-            data["feedback"] = submission.get("text", "")
-            data["problems_feedback"] = submission.get("problems", {})
+            data["grade"] = submission.grade
+            data["result"] = submission.result
+            data["feedback"] = submission.text
+            data["problems_feedback"] = submission.problems
```
nrybowski, line 65: This does not belong to a PR adding API tokens, please, open a dedicated PR for that.

inginious/frontend/pages/api/submissions.py
```diff
@@ -188,20 +194,24 @@ def API_POST(self, courseid, taskid):  # pylint: disable=arguments-differ
...
+        else:
+            user_input = flask.request.form.copy()
+            for problem in task.get_problems():
+                pid = problem.get_id()
+                if problem.input_type() == list:
+                    user_input[pid] = flask.request.form.getlist(pid)
+                elif problem.input_type() == dict:
+                    user_input[pid] = flask.request.files.get(pid)
+                else:
+                    user_input[pid] = flask.request.form.get(pid)
```
nrybowski, line 207: Is the validation really not necessary when the input is json-encoded?

inginious/frontend/pages/api/submissions.py
```diff
@@ -188,20 +194,24 @@ def API_POST(self, courseid, taskid):  # pylint: disable=arguments-differ
...
 
         user_input = task.adapt_input_for_backend(user_input)
 
-        if not task.input_is_consistent(user_input, current_app.config('ALLOWED_FILE_EXTENSIONS'),
-                                        current_app.config.get('MAX_FILE_SIZE')):
+        if not task.input_is_consistent(user_input, current_app.config['ALLOWED_FILE_EXTENSIONS'],
+                                        current_app.config['MAX_FILE_SIZE']):
```
nrybowski, line 212: How is that related to the API token addition?

inginious/frontend/pages/api/submissions.py
```diff
@@ -210,7 +220,142 @@ def API_POST(self, courseid, taskid):  # pylint: disable=arguments-differ
...
+
+    def API_POST(self, courseid, taskid
+    =None):
+        """
+            List all the submissions from a course that were evaluated (done). Or all submissions for a particular task in case a task id is given.
+            # TODO : have two different docs ?How to display them separately in the documentation ?
+            Only accessible to staff members of the course.
+            Returns a 200 OK if the endpoint is reachable and the user has access to it.
+            Returns 403 Forbidden if the user does not have access to the course/task.
+            Returns 404 Not Found if the course does not exist.
```
nrybowski, line 244: Hard no if not authorized. This could expose private courses to unauthorized users.

inginious/frontend/pages/api/submissions.py
```diff
@@ -210,7 +220,142 @@ def API_POST(self, courseid, taskid):  # pylint: disable=arguments-differ
...
+                -H "Content-Type: application/json"  -d '{ "select": "last", "username" : ["user1"] }'
+
+        """
+
+        username = flask.g.user.username
+
+        try:
+            course = Course.get(courseid)
+        except:
+            raise APINotFound("Course not found")
```
nrybowski, line 289: Hard no, a student could poke for hidden courses. Raise an APIForbidden exception instead and log that event, this should be reported to administrators since it could reveal an ongoing attack.

Ideally, loading the list of admins of a course should not trigger the loading of the course. Such information could be stored in the db. But this is for another PR.

inginious/frontend/pages/api/submissions.py
```diff
@@ -210,7 +220,142 @@ def API_POST(self, courseid, taskid):  # pylint: disable=arguments-differ
...
+        try:
+            _ = course.get_task(taskid) if taskid else None
+        except:
+            raise APINotFound("Task not found")
```
nrybowski, line 293: Hard no. DoS vector, loading a task could be costly and an unauthorized user could trigger this code path. Load the task after user authorization.

inginious/frontend/models/user.py
```diff
@@ -19,6 +25,7 @@ class User(Document):
     ltibindings = MapField(StringField())
     tos_accepted = BooleanField(default=False)
     apikey = StringField(default=None)
+    apitokens = ListField(EmbeddedDocumentField(APIToken), default=[])
```
nrybowski: Agree with @anthonygego

inginious/frontend/pages/api/submissions.py
```diff
@@ -210,7 +220,142 @@ def API_POST(self, courseid, taskid):  # pylint: disable=arguments-differ
...
+
+        if not self.user_manager.has_staff_rights_on_course(course, username, include_superadmins=True):
+            raise APIForbidden("You cannot access this course")
+
+        data = request.get_json(silent=True) or {}
```
nrybowski, line 298: I'm not sure it makes sense to allow default behavior. An authorized user should know the content they are looking for.
Also, we should implement some basic checks on the payload size, an attacker could try to dos the endpoint.

inginious/frontend/pages/api/submissions.py
```diff
@@ -210,7 +220,142 @@ def API_POST(self, courseid, taskid):  # pylint: disable=arguments-differ
...
+
+        select = data.get("select", "all")
+        usernames = data.get("username", None)
+
+        if select not in ("all", "best", "last"):
```
nrybowski, line 303: Use StrEnum and validate before loading any additional argument, we want to fail fast.

inginious/frontend/pages/api/submissions.py
```diff
@@ -210,7 +220,142 @@ def API_POST(self, courseid, taskid):  # pylint: disable=arguments-differ
...
+            raise APIInvalidArguments()
+        if usernames is not None and not isinstance(usernames, list):
+            raise APIInvalidArguments()
```
nrybowski, line 306: Again, load then validate. We should ensure that we got real usernames and not gigantic binary blobs before propagating them further.

inginious/frontend/pages/api/submissions.py
```diff
@@ -210,7 +220,142 @@ def API_POST(self, courseid, taskid):  # pylint: disable=arguments-differ
...
+
+        query = {"courseid": courseid, "status": "done"} if taskid is None else {"courseid": courseid, "taskid": taskid, "status": "done"}
```
nrybowski, line 308: We do not want dict built on the fly anymore. Specify the request format in a dataclass and dump it later with `asdict`.

inginious/frontend/pages/api/submissions.py
```diff
@@ -210,7 +220,142 @@ def API_POST(self, courseid, taskid):  # pylint: disable=arguments-differ
...
+
+        if select in ("best", "last"):
+            seen = set()
+            result = []
+            for s in submissions:
+                key = (tuple(sorted(s.username)), s.taskid)
+                if key not in seen:
+                    seen.add(key)
+                    result.append(s)
+            submissions = result
```
nrybowski: So if a user requests for 1k submissions, we iterate over each of them? And if we have ten users that trigger that code path at the same time? DoS vector, adapt the db query accordingly rather than filtering afterwards.

inginious/frontend/pages/api/submissions.py
```diff
@@ -210,7 +220,142 @@ def API_POST(self, courseid, taskid):  # pylint: disable=arguments-differ
...
+            query["username__in"] = usernames
+
+        if select == "best":
+            submissions = Submission.objects(**query) \
+                .only("courseid", "taskid", "username", "submitted_on", "result", "grade", "stderr", "stdout", "input") \
+                .order_by("-grade", "-submitted_on")
+        else:  # select == "last" or select == "all"
+            submissions = Submission.objects(**query) \
+                .only("courseid", "taskid", "username", "submitted_on", "result", "grade", "stderr", "stdout", "input") \
+                .order_by("-submitted_on")
```
nrybowski, line 319: The only difference between the two queries is the `-grade` in order_by. Do not duplicate that code for a single field.

inginious/frontend/pages/api/submissions.py
```diff
@@ -210,7 +220,142 @@ def API_POST(self, courseid, taskid):  # pylint: disable=arguments-differ
...
+            return {
+                "courseid": s.courseid,
+                "taskid": s.taskid,
+                "username": s.username,
+                "submitted_on": s.submitted_on.strftime("%Y-%m-%dT%H:%M:%SZ"),
+                "result": s.result,
+                "grade": s.grade,
+                "stderr": s.stderr,
+                "stdout": s.stdout,
+                "input": json.loads(json_util.dumps(s.get_input())),
```
nrybowski: Beside the questionable use of this function, this code does not belong to the API, but to the Submission definition itself.
On the input question, the submission API endpoint has nothing to do with the input of a submission. The client should know which task generated the submission and what kind of input is requested from the student. The API should not try to interpret that field, pipe it directly to the client.

inginious/frontend/pages/api/submissions.py
```diff
@@ -210,7 +220,142 @@ def API_POST(self, courseid, taskid):  # pylint: disable=arguments-differ
...
+            }
+
+        submissions_list = [serialize(s) for s in submissions]
```
nrybowski: For large amounts of submission the client may want to process them as soon as they arrive. This API endpoint should leverage HTTP streaming when the number of submissions cross a given threshold.
This also resolves the issue of memory pressure by trying to send everything in a single response.

inginious/frontend/pages/api/submissions.py
```diff
@@ -210,7 +220,142 @@ def API_POST(self, courseid, taskid):  # pylint: disable=arguments-differ
...
+                key = (tuple(sorted(s.username)), s.taskid)
+                if key not in seen:
+                    seen.add(key)
+                    result.append(s)
+            submissions = result
+
+        def serialize(s):
+            return {
+                "courseid": s.courseid,
+                "taskid": s.taskid,
```
nrybowski, line 325: Why keep these fields since they will be the same for each submission?

inginious/frontend/pages/api/tasks.py
```diff
@@ -62,12 +62,14 @@ def API_GET(self, courseid, taskid):  # pylint: disable=arguments-differ
             Found.
         """
 
+        username = flask.g.user.username
+
         try:
             course = Course.get(courseid)
         except:
             raise APINotFound("Course not found")
```
nrybowski, line 70: Same comment as above.

inginious/frontend/pages/preferences/apitoken.py
```diff
@@ -0,0 +1,87 @@
...
+    """ Page to view or generate an API token for the user """
+
+    def GET_AUTH(self):
+        """ GET request """
+
+        user = User.objects(username=session["username"]).first()
+        token_dict = {}
+
+        for token in user.apitokens:
+            token_dict[token.token] = token
```
nrybowski: Dict comprehension.

inginious/frontend/pages/preferences/apitoken.py
```diff
@@ -0,0 +1,87 @@
...
+
+        return self.show_page()
+
+    def POST_AUTH(self):
+        """ POST request, generates a new token for the user """
+
+        api_jwt_secret = current_app.config.get('API_JWT_SECRET')
+        api_jwt_algorithm = current_app.config.get('API_JWT_ALGORITHM')
+        api_jwt_lifetime = datetime.timedelta(days=current_app.config.get('API_JWT_LIFETIME'))
```
nrybowski: Should be user choice. Take a look at the token API from github.

inginious/frontend/pages/preferences/apitoken.py
```diff
@@ -0,0 +1,87 @@
...
+        # generate a new token
+        if "generate" in request.form:
+            expiration = datetime.datetime.now(tz=timezone.utc) + api_jwt_lifetime
+            payload = {
+                "username": user.username,
+                "exp": expiration.timestamp(),
+            }
+            token = jwt.encode(payload, api_jwt_secret, algorithm=api_jwt_algorithm) # TODO : hash token
+
+            return self.show_page(generated_token=token)
```
nrybowski: So we generate a new token, show it to the user, but it is not saved? Please, take a look at the github token page.

inginious/frontend/pages/preferences/apitoken.py
```diff
@@ -0,0 +1,87 @@
...
+    def show_page(self, generated_token=None, errors=None):
+        """ Prepares and shows the course marketplace """
+        if errors is None:
+            errors = []
+
+        user = User.objects(username=session["username"]).first()
+        token_list = []
+
+        for token in user.apitokens:
+            token_list.append(token)
```
nrybowski: List comprehension.

inginious/frontend/pages/preferences/apitoken.py
```diff
@@ -0,0 +1,87 @@
...
+
+            if description == "" :
+                return self.show_page(errors=["Description is required to save the token."])
+
+            expiration = jwt.decode(generated_token, api_jwt_secret, algorithms=[api_jwt_algorithm])["exp"]
+            expiration  = datetime.datetime.fromtimestamp(expiration, tz=timezone.utc)
+            user.apitokens.append(APIToken(token=UserManager.hash_password(generated_token), expires=expiration, description=description))
+            user.save()
+
+            return self.show_page()
```
nrybowski: So the workflow to register a new token is:
- In a first POST call, encode the token with its expiration date.
- Show it to the user.
- If the user likes their new shiny token, they can save it with an optional description in a second POST call.
- Within the second POST call, we decode the token we previously encoded to get the expiration date back and re-parse it.
- Finally, we save the token.

Does that make any sense? Its preferable to minimize the number of network requests and cryptographic operations server-side. When the user clicks on generate, it should reveal a hidden form that asks for additional data, then when they click on save, make a single POST request.

inginious/frontend/pages/preferences/apitoken.py
```diff
@@ -0,0 +1,87 @@
...
+
+        # invalidates a token
+        if "delete" in request.form:
+            token_to_delete = request.form.get('token')
+            user.apitokens = [token for token in user.apitokens if token.token != token_to_delete]
```
nrybowski: Store the tokens in a map.

inginious/frontend/pages/preferences/apitoken.py
```diff
@@ -0,0 +1,87 @@
...
+
+            expiration = jwt.decode(generated_token, api_jwt_secret, algorithms=[api_jwt_algorithm])["exp"]
+            expiration  = datetime.datetime.fromtimestamp(expiration, tz=timezone.utc)
+            user.apitokens.append(APIToken(token=UserManager.hash_password(generated_token), expires=expiration, description=description))
+            user.save()
+
+            return self.show_page()
+
+        # invalidates a token
+        if "delete" in request.form:
```
nrybowski, line 75: Maybe use the dedicated HTTP code to request the deletion of a resource?

inginious/frontend/templates/apitoken.html
```diff
@@ -0,0 +1,114 @@
...
+        <button type="button" class="close" data-dismiss="alert" aria-hidden="true">&times;</button>
+        {{ error }}
+    </div>
+{% endfor %}
+
+<div class="col-sm-12">
+    <p>
+        {{ _("Your API token allows external applications and scripts to authenticate as you and interact with "
+             "INGInious on your behalf, without needing your password. Keep it secret: anyone who has it can "
+             "access the API using your account permissions. If you believe your token has been compromised, "
```
nrybowski, line 35: How is the user supposed to overwrite the list of token you store?

inginious/frontend/templates/apitoken.html
nrybowski: As commented above, this form should appear on a click on generate token.

inginious/frontend/app.py
```diff
@@ -60,6 +60,9 @@ def _put_configuration_defaults(config):
                                               [".c", ".cpp", ".java", ".oz", ".zip", ".tar.gz", ".tar.bz2", ".txt"]),
         "ALLOW_DELETION": config.get("allow_deletion", True),
         "ALLOW_REGISTRATION": config.get("allow_registration", True),
+        "API_JWT_SECRET": config.get("api_jwt_secret", "jwt_secret_key"),
+        "API_JWT_LIFETIME": config.get("api_jwt_lifetime", 365),
+        "API_JWT_ALGORITHM": "HS256",
```
nrybowski, line 63: The administrator could request an algorithm that requires a private key but there is not mean to provide it... We will just hardcode `HS512` for now.

inginious/frontend/models/user.py
```diff
@@ -5,8 +5,14 @@
...
 
-from mongoengine import Document,  StringField, ListField, MapField, BooleanField, DynamicField
+from mongoengine import Document,  StringField, ListField, MapField, BooleanField, DynamicField, EmbeddedDocument, EmbeddedDocumentField, DateTimeField
 
+class APIToken(EmbeddedDocument):
+    """ Embedded document for API tokens. Contains the token hash, expiration date, description and hash algorithm used. """
+    token = StringField(required=True)
+    expires = DateTimeField(required=True)
+    description = StringField(required=True)
+    algorithm = StringField(required=True, default="HS256")
```
nrybowski: You set `HS256` as the default value but its not overwritten in `apitoken.py`, it is not even used anywhere. Remove it.

inginious/frontend/submission_manager.py
```diff
@@ -411,20 +406,18 @@ def kill_running_submission(self, submissionid, user_check=True):
         self._client.kill_job(submission["jobid"])
         return True
 
-    def user_is_submission_owner(self, submission):
+    def user_is_submission_owner(self, submission ):
         """ Returns true if the current user is the owner of this jobid, false else """
         if not session.loggedin:
-            raise Exception("A user must be logged in to verify if he owns a jobid")
+            raise Exception("A user must be logged in to get his submissions")
```
nrybowski, line 412: their

nrybowski (changes requested):
All in all, this PR requires some heavy lifting.

nrybowski:
> > * endpoint names : I have for the moment used `/api/v0/token` in the endpoints I have added. However since we don't use the flask session anymore all endpoints use th token authentication and this distinction is not needed anymore. I will change that. But do you want to change the version of the api ?
> 
> Yes, I think we can, even if the interface doesn't change. That will invalide any old usage of the API. No need to preserve any retrocompatibility, the thing was broken for years.

I'm uncomfortable with such a breaking change, we should deploy dummy endpoints documenting the API changes for any scripts out there in the wild that may still use the old (and maybe broken) API. Maybe some 301 with a payload rather than just plain 404 out of nowhere. Such endpoints should only be available for a single release.

nrybowski:
Regarding to https://github.com/INGInious/INGInious/pull/1121/changes#r3880972260, we could generate a token per course upon admin user request, the authorized course id would be embedded within the JWT. Hence, we should not even load to course to ensure that the user is authorized to call the submission API for the specified course.
This could be implemented in the same manner as the webdav access token on the course administration page.
@anthonygego your thoughts?

anthonygego:
> Regarding to https://github.com/INGInious/INGInious/pull/1121/changes#r3880972260, we could generate a token per course upon admin user request, the authorized course id would be embedded within the JWT. Hence, we should not even load to course to ensure that the user is authorized to call the submission API for the specified course. This could be implemented in the same manner as the webdav access token on the course administration page. @anthonygego your thoughts?

Yes that was planned for #1041 and #1042 but the work was not finished by the end of the internship. I was in favor of both user (mainly for submission) and course (mainly for stats) tokens. This can in practice just be implemented as scopes, I don't know. The main requirements is that students should be able to generate tokens too without being manageable by course admins.

But I really think that one feature must be implemented at a time in order to move on.

inginious/frontend/flask/mapping.py
AlexandreDoneux: In this PR we change from authentication using the flask session to a token. Using the endpoints as before will inevitably fail. Bumping the version feels to me like a good way to indicate such a big change.

Although I will grant you this is missing some documentation. Plus, when calling an non-existent url, INGInious send back an HTML page. I will implement specific messages for "/api" and "/api/v0" to indicate an endpoint is not available or is obsolete. 

Should I keep legacy endpoints rather than returning an error ?

AlexandreDoneux:
I chose to hardcode the algorithm used for JWT. Allowing the choice for it could become messy due to different algorithms using different values (secrets, public keys, private keys). We would need to keep track of all the algorithms as well as all the different secrets and keys used to allow old tokens.

On top of that, changing from a symmetric to an asymmetric algorithm without changing the secret could lead to an algorithm-confusion attack.

inginious/frontend/pages/api/submissions.py
AlexandreDoneux: It is a fix. `current_app.config("ALLOWED_FILE_EXTENSIONS)` is missing the `.get`. I just replaced the `.get()` by a bracket notation. Would you prefer I keep the `.get()` ?

inginious/frontend/pages/api/submissions.py
AlexandreDoneux: Not sure I understand your remark. When you say "should", do you mean that it should be implemented ? Or that the client should already know that and therefor should not be implemented ? 

If it is the first case : The client knows the task through the taskid. Or had you something else in mind ? For the type of input requested to the student, do you expect a field giving the problems and their type ?

inginious/frontend/templates/apitoken.html
AlexandreDoneux: Updated message. This one was valid when we were allowing one token at a time.

inginious/frontend/pages/preferences/apitoken.py
AlexandreDoneux: It is doable but I feel like this is a lot of changes for not much. I would have to add the different methods to `INGIniousPage` and `INGIniousAuthPage`, plus a jquery event listener because HTML does not handle the delete method for forms.

inginious/frontend/pages/api/submissions.py
AlexandreDoneux: I have set limits for the username lengths and the number of usernames. I have set arbitrary values but plan to change them when implementing a limit on certain values in a following PR.

For example, there is no upper limit for usernames.

AlexandreDoneux:
Closing this PR. We will implement the submissions API and the generation and use of tokens through smaller PRs. Starting with #1170.

PR #1122, opened 2026-07-16
deploy: Bump to RL10 and Python3.14

nrybowski:
Also bump msgpack to 1.2.1 and pyzmq to 27.1.0 due to build failures for previous versions with Python3.14.


anthonygego (commented):
There are issues with ssh debugging and ``ssh_student`` due to the migration to EL10 and Python3.14

- EL10 ships with openssh 9. The ssh session is no more advertised as a root `sshd` process with `<user>@pts/X` in the command line but rather as a user `sshd-session` process with still `<user>@pts/X` in the command line.

  Therefore lines :
   https://github.com/nrybowski/INGInious/blob/87b4f97ab3e9b23762bdf1594202967154379cbf/base-containers/base/inginious_container_api/utils.py#L94

  and 
  https://github.com/nrybowski/INGInious/blob/87b4f97ab3e9b23762bdf1594202967154379cbf/base-containers/base/inginious_container_api/utils.py#L105

  don't work and user detection fails.
- Python3.10 deprecates the automatic creation of the loop using `asyncio.get_event_loop()` : https://docs.python.org/3.10/library/asyncio-eventloop.html#asyncio.get_event_loop
  And this is removed in Python 3.14.
 
As I can't push the fixes to this PR I'll merge it and push them afterwards

PR #1127, opened 2026-07-16
frontend: Bump old HTTP reference to explicit HTTPS for external fonts.

nrybowski:


inginious/frontend/static/css/lti.css
```diff
@@ -309,6 +309,11 @@ body {
     transform: rotate(90deg);
   }
 }
+@media (max-width: 700px) {
+  h2 {
+    font-size: 20px;
+  }
+}
```
anthonygego, line 316: This addition is a bit intriguing...because obsviously unexpected at this stage 🫠 

It seems this piece of code has been added in the less file here: https://github.com/INGInious/INGInious/commit/59f51db7320b6043b2cbc342cdabe1f33f575093

However,  the `lti.css` file hasn't been changed in that commit. I suspect it has been added manually in both files or the `lti.css` file has'nt been recompiled since.

In all cases, this is the way to go, so I'll merge this and we'll fix the potential upcoming issue in another PR.

anthonygego (approved):

PR #1138, opened 2026-08-06
Platform-wide LTI1.3 and Deep Linking

### LINKED ISSUE #1020: [LTI] Allow a single per-course activity
When using many Inginious tasks in Moodle, we have to configure and copy-paste by hand a dozen fields for each activity. It's very cumbersome.

Simplest A) Have a per-course activity instead of per-task.

When creating an instance of the LTI tool on Moodle (or any LMS), the first view can simply let me select one INGInious task that is the one I want to use in this activity. Then afterwards it shows the right INGInious task directly. I think Gradescope kind of does that.

The LTI exposes enough course information and activity id to create a unique mapping between the unique activity and the selected task. 

This is simpler because currently the INGInious LTI is organised course-wide. We could even maybe have another URL (using / instead of /task for instance) to enable both the old way and course-wide way.

Alternative B) Moodle-wide INGinious integration.

The advantage of having simply a global, INGInious-wide integration is that a Moodle administrator (eg Moodle UCLouvain) might add the activity for everybody, from there it would help foster the use of INGInious. But that part requires much more changes as the LTI keys etc would need to be INGInious-wide.

From a user point of view, when creating an activity I would have to select one of my courses and then one of my tasks, it's not much more complicated.

**Describe alternatives you've considered**
Ask assistants to do it.

**Additional context**
Screenshot of the pain:
image

### LINKED ISSUE #1054: [frontend/lti] use custom paramters to select course and task

Currently, the LTI laucnh URLs include the course and task ids. This is quite cumbersome when configuring some LMS, such as Moodle, which require a tool definition that embeds URLs.

Moreover, LTI1.3 Deep Linking could not be implemented in a useful way with this constraint. Besides, defining LTI1.3 tasks is currently really painful.

Course and task ids should be passed as custom arguments/parameters in the LTI message. 

This way defining activities in popular LMS would be simplified and Deep Linking could be implemented more simply, allowing teachers to select the task in an INGInious window, that would fill the LTI custom arguments field in those LMS.

anthonygego:
This PR deprecates LTI1.1 and improves LTI1.3 implementation, more specifically : 

- It separates the LTI1.1 and LTI1.3 into two separate tabs in the course admin and mentions LTI1.1 is deprecated.
- It removes the courseid and taskid from the stored clear launch message. 
  This one is a bit tricky. The default storage to keep LTI1.3 launch messages in the library is cookies. To ensure no cookies without valid OIDC login are forged, the library uses the session id for the data keys. In our case the data are stored in database and alreday linked with the flask session. The risk is therefore limited and there is no need to add any additional contextual information. Besides, these keys were useless as their values are also passed in the launch messages.
- It allows to specify a platform-wide LTI1.3 config in the configuration file under the key `lti_config` and provides adapted endpoints, while preserving retrocompatibility.
- It introduces the `courseid`, `taskid` and `secret` LTI custom parameters to select the INGInious task through the platform-wide LTI1.3 launch process. Secrets are simple clear password strings that prevents anyone from importing a task without being the course administrator. It is defined and unique at the LTI1.3 deployment level. There is no reason for multiple secrets by deployment.
- It replaces LTI1.3 task access check by secret comparison.
- It refactors the `ltibindings` in database to remove the course id from the LTI1.3 context and flattens everything. The platform identifier is the full deployment string `iss/client_id/deployment_id` for LTI1.3 and `courseid/lti_key` for LTI1.1. This allows deployment linking and therefore implementation of deep linking. A database migration process is provided.
- It implements Deep Linking on both the launch URL and an alias route to support platforms that require a different route for LTI deep linking (such as edX). When a deeplink launch is received, INGInious redirects to a new task selection page instead. That latter returns the aforementioned LTI custom parameters required for the task to integrate. If the `secret` does not exist yet, it is generated randomly.
- It provides a GUI to show available LTI1.3 clients in the course settings and manually generate deployment `secret`s in case the target LMS does not support Deep Linking. The secret is generated randomly using a simple javascript UUID method. There is no need for a specific algorithm or security scheme here. It is just more convenient for users to get it automatically generated instead of choosing one themselves. The `secret` is, in all cases, stored as a clear password in the YAML file, as this is the case for LTI1.1 or registration passwords.
- It updates the LTI documentation to reflect these changes for Moodle and edX.

This PR keeps full retrocompatibility for now with course-specified platforms. LTI configurations are merged on the fly when validating requests. It is not expected that more than 2-3 configurations are used in total so this may not be very clean but does not impact performance. Keeping this ability relies on the fact that some platforms, such as edX, do not provide a way to define global configuration without an action from the support staff, making prototyping less convenient. 
I guess we may remove this ability in the future.

Closes #1054


inginious/frontend/pages/lti/v1_1/__init__.py
```diff
@@ -87,6 +88,26 @@ def get_client_secret(self, client_key, request):
...
+    def is_lti_page(self):
+        return True
+
+    def GET_AUTH(self):
+        data = session.lti
+        if data is None:
+            raise Forbidden(description=_("No LTI data available."))
+        (courseid, taskid) = data['task']
+
+        return BaseTaskPage(self).GET(courseid, taskid, True)
```
nrybowski: Maybe add an helper for that since the same snippet is called in the POST.

inginious/frontend/pages/lti/v1_3/__init__.py
```diff
@@ -149,12 +198,70 @@ def _find_realname(self, launch_data):
...
+        if not session.is_lti:
+            raise Exception("Not an LTI session")
+
+        lti_courses = {
+            courseid: course for courseid, course in Course.get_all().items()
+            if self.user_manager.has_admin_rights_on_course(course) and course.is_lti()
+        }
+
+        if "courseid" in request.args:
+            courseid = request.args.get("courseid")
```
nrybowski: `if (courseid := request.args.get("courseid") is not None:` ?

inginious/frontend/pages/lti/v1_3/__init__.py
```diff
@@ -5,33 +5,71 @@
...
+        except CourseNotFoundException as ex:
             raise NotFound(description=_(str(ex)))
 
-        lti_config = course.lti_config()
+        global_config = current_app.config.get("LTI_CONFIG")
+
+        # Merge with global config
+        for iss in global_config:
+            lti_config.setdefault(iss, [])
+            lti_config[iss] += global_config[iss]
```
nrybowski: Detail:
``` python
for iss, cfgs in global_config.items():
    lti_config.setdefault(iss, [])
    lti_config[iss] += cfgs
```

nrybowski (changes requested):
Small details from a code viewpoint, they may be addressed in further PRs if we are in a hurry. The PR has been tested against a Moodle instance and is functional.
I will add documentation for the database transition process in a compose-based deployment.

inginious/frontend/pages/lti/v1_1/__init__.py
anthonygego: Fixed : https://github.com/INGInious/INGInious/blob/platform_lti/inginious/frontend/pages/lti/v1_1/__init__.py#L95

nrybowski (approved):
LGTM.

PR #1139, opened 2026-08-09
[backend/agents] replace deprecated ZMQEventLoop with asyncio's default loop

### LINKED ISSUE #800: [inginious-backend] ZMQ functions deprecated
```
inginious-backend:39: DeprecationWarning: ZMQEventLoop and zmq.asyncio.install are deprecated in pyzmq 17. Special eventloop integration is no longer needed.
  loop = ZMQEventLoop()
```

vjymisal0:
Fixes #800

## Problem

`zmq.asyncio.ZMQEventLoop` and `zmq.asyncio.install()` have been deprecated since pyzmq 17, and every backend/agent process prints this warning on startup:

```
inginious-backend:39: DeprecationWarning: ZMQEventLoop and zmq.asyncio.install are deprecated in pyzmq 17. Special eventloop integration is no longer needed.
  loop = ZMQEventLoop()
```

Since pyzmq 17, `zmq.asyncio` sockets work directly with the regular asyncio event loop, so the special `ZMQEventLoop` subclass is unnecessary.

## Fix

Replaced `loop = ZMQEventLoop()` with `loop = asyncio.new_event_loop()` (and dropped the now-unused `ZMQEventLoop` import) in all four places that had this pattern:

- `inginious/scripts/backend.py`
- `inginious/scripts/agent_mcq.py`
- `inginious/scripts/agent_docker.py`
- `inginious/frontend/arch_helper.py` (the local-arch helper used when `BACKEND: local`)

`zmq.asyncio.Context()` and everything downstream (`loop.set_debug()`, `loop.run_until_complete()`, `loop.close()`, `context.destroy()`) is unchanged — only the loop construction changes.

## Test plan

These are process entrypoints that open real ZMQ sockets/Docker, so there's no existing unit test harness for them. I verified the fix directly against `pyzmq` (26.2.1):

- **RED**: constructing `zmq.asyncio.ZMQEventLoop()` under `warnings.catch_warnings(record=True)` reproduces the exact `DeprecationWarning` from the issue.
- **GREEN**: with the new `asyncio.new_event_loop()` + `zmq.asyncio.Context()` pattern, a REQ/REP round-trip over `inproc://` completes successfully with zero `DeprecationWarning`s raised.
- `python -m py_compile` on all four changed files succeeds.

I did not touch `zmq.asyncio.Context()` usage or any other part of the async lifecycle in these files — scope is limited to the deprecated loop construction call, as requested in the issue.

### DIFF
--- a/inginious/frontend/arch_helper.py
+++ b/inginious/frontend/arch_helper.py
@@ -16,9 +16,6 @@ def start_asyncio_and_zmq(debug_asyncio=False):
     """ Init asyncio and ZMQ. Starts a daemon thread in which the asyncio loops run.
     :return: a ZMQ context and a Thread object (as a tuple)
     """
-    # ZMQEventLoop/zmq.asyncio.install() are deprecated since pyzmq 17: zmq
-    # sockets work with the regular asyncio event loop, no special
-    # integration is needed anymore.
     loop = asyncio.new_event_loop()
     asyncio.set_event_loop(loop)
     if debug_asyncio:
--- a/inginious/frontend/arch_helper.py
+++ b/inginious/frontend/arch_helper.py
@@ -8,7 +8,7 @@ import asyncio
 import multiprocessing
 import threading
 
-from zmq.asyncio import ZMQEventLoop, Context
+from zmq.asyncio import Context
 
 from inginious.client.client import Client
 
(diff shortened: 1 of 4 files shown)


nrybowski (changes requested):
Hi, could you sign the CLA so that we can consider your PR? Also, do you think it has any interest to repeat the exact same comment in four different files to indicate the replacement of a single deprecated function call? Please, remove the comments as they do not add any value. The contribution is already detailed in the commit message, which is the expected way to document such a small change.

vjymisal0:
Fixed in c6977816 — trimmed the duplicated ZMQEventLoop comment to a single reference in arch_helper.py, with the other three call sites (agent_docker.py, agent_mcq.py, backend.py) pointing to it. On the CLA: that requires vjymisal0's own action (signing/e-signature), so I'm not able to complete it here — flagging for you to handle.

vjymisal0:
Fixed in 037cdd6 — removed the ZMQEventLoop comments entirely from all four files rather than leaving a cross-reference; the rationale stays in the commit message as suggested. (Note: CLA signing needs to happen through the linked CLA-assistant flow, which I can't do on your behalf.)

vjymisal0:
have signed the CLA

nrybowski:
After discussing with @anthonygego, we have decided to not consider furthermore obviously low effort, AI-only, PRs. Closing.

PR #1143, opened 2026-08-11
deploy: Fix duplicate superadmin insertion on DB init.

nrybowski:
Since 11a6873ef393afffa9cd2a4bf5020c5acdaf2cfe, the database of docker-compose based deployments is persitent on the host disk. Because the db_setup.py script is launched at each deployment, the superadmin user was added multiple times in the DB.

This commit fixes the issue by first testing the existence of the superadmin before adding it.

### DIFF
--- a/deploy/db_setup.py
+++ b/deploy/db_setup.py
@@ -1,3 +1,5 @@
+import sys
+
 from pymongo import MongoClient
 from gridfs import GridFS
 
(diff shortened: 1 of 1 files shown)


deploy/db_setup.py
```diff
@@ -41,6 +41,16 @@ def try_mongodb_opts(host="localhost", database_name='INGInious'):
 
     database = try_mongodb_opts('db')
 
+    # Test whether the superadmin is already set.
+    entries = database.users.find({"username": username})
```
anthonygego: I think you can simply use `find_one` that returns `None` if there is no match, to avoid exception catching.

anthonygego (changes requested):

deploy/db_setup.py
nrybowski: Fixed in bb97f7e4ffa930c6a7ceeec8933b19fbc6b3b77f

anthonygego (approved):

PR #1119, opened 2026-07-09
agents/docker: Refactor Docker agent.

nrybowski:
This PR is a major refactor of the Docker agent which focuses on two main points: (i) deprecating the kata-runtime as it is unused and adds a lot of unnecessary complexity and, (ii) implementing a robust communication scheme between the agent, the grading containers and the student containers.

Regarding to the first point, we experimented a few years ago with launching custom kernels within student container (#949). This is a solid alternative to the kata-runtime for running student code as root, e.g., for networking exercises, without compromising INGInious security. This feature has been blocked since then because of the way the Docker agent exposes its capabilities, which in turns impacts the way jobs are distributed among available agents.

  Regarding to the second point, the Docker agent communicates with grading and student containers through their std{in,out} interfaces which is quite fragile and blocks a current effort of removing the Docker dependency (see #1115, #1116)  in favor of generic [OCI compatibility](https://opencontainers.org/about/overview/). Also, when the kata-runtime is used, the Docker agent serves as a proxy between the grading and student containers while they can communicate directly when both of them run on the same kernel. This is not required by #949.
  The Docker agent also relies on the ZMQ framework for communications. This is an heavy framework whose advanced features are not used. We deprecate it in favor of a simpler and lower level approach based on simple bytestreams over TCP and Unix sockets which is already in use in the existing code. By leveraging the [Transport and Protocol features of the asyncio](https://docs.python.org/3/library/asyncio-protocol.html#) framework, we also can get rid of the custom buffering that has been implemented. The goal here is to mainly rely on features of the standard python library and simplify the code base.
  Furthermore, the messages sent between each element are not formally described and are encoded with dictionaries built-on the fly. This approach is not reliable and makes the messages hard to correctly extend.

The kata-runtime is so intermingled with the Docker agent I/O that it is nearly impossible to deprecate it without refactoring the whole Docker agent I/O.
This PR is an effort in the continuity of #1115 and #1116, and should be a cleaner basis for #949.
It implements on the following changes:

- [ ] Docker agent.
  - [x] Formalizing messages exchanged with the grading containers.
  - [ ] Refactoring the I/O loop.
    - [x] Exposing TCP endpoint for communication rather than containers std{in,out}.
    - [x] Leveraging formalized messages.
    - [ ] Deprecating kata-runtime related code.
    - [ ] Deprecate ZMQ.
  - [ ] Refactor the Docker interface to reflect  the communication changes.
    - [ ] Remove Unix socket mount point.
[...]

PR #1145, opened 2026-08-12
Removing support for kata runtime

nrybowski:
This PR deprecates the kata-runtime.

Follow-up of #1119.


anthonygego (approved):
LGTM.

It seems we have an issue here :

https://github.com/INGInious/INGInious/blob/cf574b3b6b2d58b3133571044f6bd2295045d7d3/inginious/agent/docker_agent/__init__.py#L500

when running `run_student --run-as-root` with no root environment available, but this is actually unrelated to this PR.

PR #1146, opened 2026-08-13
Fixing install script for container multi-staged build

AlexandreDoneux:
This PR fixes the inginious install script to adapt to the multi staged build for default containers. It also ignores the new superadmin added if it already exists.

### DIFF
--- a/inginious/frontend/installer.py
+++ b/inginious/frontend/installer.py
@@ -387,10 +387,10 @@ class Installer:
     #             CONTAINERS              #
     #######################################
 
-    def _build_container(self, name, folder):
+    def _build_container(self, name, folder, target=None):
         self._display_info("Building container {}...".format(name))
         docker_connection = docker.from_env()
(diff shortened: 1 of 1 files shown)


inginious/frontend/installer.py
```diff
@@ -560,7 +560,10 @@ def configure_authentication(self):
 
         password = self._ask_with_default("Enter the password of the superadmin", "superadmin")
 
-        User(username=username, realname=realname, email=email, password=UserManager.hash_password(password)).save()
+        if not User.objects(username=username):
```
anthonygego, line 563: I'm pretty sure you need `.first()` here since otherwise you'll get a mongoengine iterator.

AlexandreDoneux: It is not required as `.objects()` returns a QuerySet that implements truthiness by checking it's length. 

I could add `.first()` for efficiency purposes to avoid pulling data for more than one user. But it should not happen due to username being unique.

PR #1148, opened 2026-08-13
Allow empty categories input in task settings

AlexandreDoneux:
In the task settings, removing all categories from a task would not work because it would reject empty tags strings. I removed that restriction and made sure `Categories.get_values()` does not return those empty strings. The problem is those empty strings still show up in the `course.yaml`.

Changing the `parse_tasks_config()` method to allow updating the configuration could fix that but it causes multiple issues in other places. Would you like me to address those or do you think having empty string tags in the course.yaml is acceptable.

### DIFF
--- a/inginious/frontend/templates/course_admin/task_dispensers/config_items/categories.html
+++ b/inginious/frontend/templates/course_admin/task_dispensers/config_items/categories.html
@@ -31,7 +31,7 @@
 
         tasks.each(function () {
             var taskid = $(this).data("taskid");
-            dispenser_config[taskid]["categories"] = value.split(",");
+            dispenser_config[taskid]["categories"] = value.split(",").filter(item => item !== "");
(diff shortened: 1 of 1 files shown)


inginious/frontend/task_dispensers/util.py
```diff
@@ -148,7 +148,7 @@ def get_name(cls):
     def get_value(cls, task_config):
         categories = task_config.get(cls.get_id(), cls.default)
         if "" in categories:
-            raise InvalidTocException("All categories must have a name but are :" + str(categories))
+            categories = [category for category in categories if category != ""]
```
anthonygego: This method is called on the raw data. It is nice to prevent users from writing bad YAML files.

The fix should be done earlier, probably in https://github.com/INGInious/INGInious/blob/main/inginious/frontend/templates/course_admin/task_dispensers/config_items/categories.html 

There is no reason the task settings form produce the wrong format.

Issue #1137, opened 2026-08-04
dnf is unavailable in containers


AlexandreDoneux:
**Describe the bug**
When using dnf inside of a container based on the env-base(inside the container, in a Dockerfile for an image based on it, etc.), this error is thrown : 

```
Traceback (most recent call last):
  File "/usr/bin/dnf", line 61, in <module>
    from dnf.cli import main
ModuleNotFoundError: No module named 'dnf'
```

This appears to be due to the system python being updated to 3.14 in recent commit 2d59b5cc9c7a57120d4e32512645efdce47ee057. Rocky Linux 10 comes with python 3.12 as system default. Changing this default versions causes an issue with the python3-dnf package and making dnf unavailable.

**INGInious installation details**
- Version: commit 37f4438ea8aff3e61841a8d1cf068af7fedf7452

**To Reproduce**
1. Create and enter a container with `docker run --rm -it ghcr.io/inginious/env-base:latest bash`
2. Use any dnf command, for example : `dnf -v`

This issue also appears when building a container image based on env-bas or env-default.

PR #1152, opened 2026-08-14
Refactor base container build and add support for python version

anthonygego:
This PR provides a first solution to ship another Python version for use with the API without breaking the system Python (cf #1137).

What has been done:
- Refactored the `inginious_container_api` module so tha it can be installed using pip through a `pyproject.toml` file. This requires to move the scripts to put into a submodule so that they can be installed in the `xxx/bin` folder.
  => pip will automatically handle the python version that needs to load those scripts, instead of using a hardcoded shebang.
- Kept a retrocompatibility module for the older ` import inginious`
- Adapted the path of `inginious-ipython` in API files  to reflect pip installation path.
- Enabled CRB repository alongside epel
- Add the `PYTHON_VERSION` argument removed in d3899cd because of the system tools not working anymore.

Current limitations : 
- Run file extensions (`run.sh`, `run.py`...) have been introduced in https://github.com/INGInious/INGInious/commit/a740c255b388a4a2580443dfd6a94e8f6cce771a 7 years ago but older `run` files have not been deprecated. If the shebang of those files specifies system python, the API won't be available. A migration process consists in renaming run files to bypass the limitation.
- Besides the run file, there may be scripts using shebangs and system Python in task folders that import the INGInious API. Those scripts may not work anymore. However, it is assumed that no one will build a environment not using the system python without an specific purpose. We can assume there is no reason to serve those environment for legacy tasks.

For the future:
- The API packaging simplifies the container setup. The only INGInious configuration remaining is the `worker` user creation. By encapsulating this process with the library install, we could remove the `env-base` constraint and make OS-agnostic containers that simply install the INGInious package.


nrybowski (approved):
LGTM, tested and functionnal. Merging.

PR #1154, opened 2026-08-14
Replace AgentRuntime by AgentCapabilities

### LINKED ISSUE #983: [agent/docker] add suport for additional container capabilities
If we want to use `perf`, for instance, we need to expose more information from the host. `SYS_ADMIN` capability is required at the launch of the container. However, it's a risky operation to add these capabilities for everyone without control.

A good compromise is to allow the INGInious administrator to run a Docker agent with additional capabilities for the containers it runs. This ensures only selected hosts are more exposed than others, and that this choice is made by the platform administrator.

This could be an `inginious-agent-docker` command line argument or stored in a config file (in this case it could be worth moving all the agent arguments, such as problem plugins, in that file).

The remaining question is to know whether this feature should be announced back to the queue so that the jobs are correctly routed to the right agent or if this should remain an *obscure* feature used with specificly designed environments for that host.

nrybowski:
This PR deprecate the concept of AgentRuntime and replace it with the concept of AgentCapabilities.

Until now, when we wanted to advertise a specific flavor of a DockerAgent (e.g. one that passthrough GPUs), we derived a new AgentRuntime (e.g. NvidiaEnvType) that was statically defined in the frontend. Defining a new flavor (e.g. enabling SSH access to student containers) required yet another AgentRuntime. See:
https://github.com/INGInious/INGInious/blob/96878e6c220e0d4709adcf0f296e57b23ead99b8/inginious/frontend/environment_types/__init__.py#L25-L33

Also, the concept is unclear since AgentRuntime is also called EnvironmentType. In reality, the project only support two types of Agents: OCI (Docker) and MCQ.
In this PR, we clarify the concepts:
- There are two `AgentType`: mcq and oci.
- Each Agent may declare `AgentCapabilities`, e.g., a given Docker Agent may expose a GPU but not SSH, while another supports SSH but does not passthrough a GPU.
- `Environment` only refers to a grading environment and not an Agent.

Also, in this PR, we deprecate the static definition of the Agents. They now directly advertise their capabilities to the Backend, which in turns updates the Clients (Frontend). Hence, the Frontend may directly display the Capabilities of the selected `AgentType` in the task edition tab. The course administrator requests some capabilities for its task, then the backend sends the task to an Agent that supports all the requested capabilities.

~This PR builds on #1145 and should not be merged before.~


inginious/agent/docker_agent/__init__.py
```diff
@@ -487,31 +475,28 @@ async def create_student_container(self, parent_info, socket_id, environment_nam
...
-
             ports_needed = [22] if ssh else []
             ports = {}
             for p in ports_needed:
                 if not self._external_ports:
                     self._logger.warning("User asked for a port but no one are available")
                     if self._external_ports is not None:
-                        for port_to_free in ports:
-                            self._external_ports.add(ports[port_to_free])
+                        for p in ports.valuesi():
```
nrybowski, line 498: typo

inginious/agent/docker_agent/i18n/fr/LC_MESSAGES/messages.po
```diff
@@ -0,0 +1,31 @@
...
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=2; plural=(n > 1);\n"
+
+#: capabilities.py:27
+msgid "The task requires a GPU."
+msgstr "La tâche nécessite un GPU."
+
+#: capabilities.py:29
+msgid "The task requires the student code to be ran as root. (EXPERIMENTAL)"
+msgstr "La tâche nécessite que le code étudiant soit lancé en tant que root."
```
nrybowski, line 26: Miss 'expérimental'

inginious/agent/docker_agent/__init__.py
```diff
@@ -63,27 +63,24 @@ class DockerRunningStudentContainer:
...
 class DockerAgent(Agent):
-    def __init__(self, context, backend_addr, friendly_name, concurrency,
-                 address_host=None, external_ports=None, debugger=False, tmp_dir="./agent_tmp", runtimes=None, ssh_allowed=False):
+    def __init__(self, context, backend_addr, friendly_name, concurrency, capabilities: DockerAgentCapabilities, address_host=None, external_ports=None, debugger=False, tmp_dir="./agent_tmp", ):
         """
         :param context: ZeroMQ context for this process
         :param backend_addr: address of the backend (for example, "tcp://127.0.0.1:2222")
         :param friendly_name: a string containing a friendly name to identify agent
         :param concurrency: number of simultaneous jobs that can be run by this agent
+        :param capabilities: the capabilities of the Agent. See DockerAgentCapabolities.
```
nrybowski, line 72: typo

inginious/agent/docker_agent/__init__.py
```diff
@@ -299,6 +287,7 @@ def __new_job_sync(self, message: BackendNewJob, future_results):
         task_id = message.task_id
 
         debug = message.debug
+        # TODO: Ensure that type matches self._type
```
nrybowski, line 290: remove

inginious/agent/docker_agent/_docker_interface.py
```diff
@@ -15,12 +15,13 @@
 
 from docker.types import Ulimit
 
-from inginious.agent.docker_agent._docker_runtime import DockerRuntime
+from inginious.agent.docker_agent.capabilities import DockerAgentCapabilities
+from inginious.common.agents import GradingEnvironment
 
 DOCKER_AGENT_VERSION = 4
```
nrybowski, line 21: @anthonygego Do we bump the version?

inginious/agent/docker_agent/capabilities.py
```diff
@@ -0,0 +1,20 @@
+from dataclasses import dataclass, field
+import gettext
+
+from inginious.common.agents import Capabilities
+
+_ = gettext.gettext
```
nrybowski, line 6: remove as well as the import

inginious/agent/docker_agent/capabilities.py
```diff
@@ -0,0 +1,20 @@
...
+
+@dataclass(frozen=True)
+class DockerAgentCapabilities(Capabilities):
+
+    def __post_init__(self):
+        object.__setattr__(self, '__translation_path__', 'agent/docker_agent/i18n')
+        super().__post_init__()
+
+    """ Indicates whether the Agent supports GPUs. """
+    gpu: bool = field(doc=_("The task requires a GPU."))
```
nrybowski, line 16: Remove the gettext

inginious/client/client.py
```diff
@@ -278,6 +280,7 @@ def new_job(self, priority, job_info, inputdata, callback, launcher_name="Unknow
                 task: Task,                 Task object related to the submission, optional, can be None
                 environment_type: str,      Type of the environment to run the job (ex : docker, docker_ssh, mcq, ...)
                 environment: str            Name of the environment to run the job in.
+                capabilities: list[str]     The Agent capabilities required to run the job.
```
nrybowski, line 283: dict

inginious/frontend/templates/course_admin/edit_tabs/env_generic_docker_oci.html
```diff
@@ -30,11 +30,30 @@
         </label></div>
     </div>
 </div>
-
```
nrybowski, line 33: restore

inginious/frontend/tasks.py
```diff
@@ -47,15 +48,30 @@ def _migrate_from_v_0_6(content):
...
+    capabilities = []
+    if (env := content.get('environment_type')):
+        if "nvidia" in env:
+            capabilities.append('gpu')
+        if 'ssh' in env:
+            capabilities.append('ssh')
+        if env != AgentType.MCQ:
+            content['environment_type'] = AgentType.OCI
+    # Do not save empty capabilities list. There is a default loader.
+    if len(capabilities) > 0:
```
nrybowski, line 62: remove

inginious/agent/docker_agent/_docker_interface.py
anthonygego: This is useful for restricting usage of containers images. If they are still fully compatible, there probably is no need, even if the name is quite misleading. The fact is that such a change requires rebuilding all the images. 

If the patch already requires config changes, then we can ask for images to be rebuilt at the same time and increment this constant. Otherwise it can remain as it is.

inginious/common/agents.py
```diff
@@ -0,0 +1,44 @@
...
+            trad_path = os.path.join(get_root_path(), self.__translation_path__)
+            available = [
+                lang for lang in os.listdir(trad_path) if os.path.isdir(os.path.join(trad_path, lang))
+            ]
+            translations.update({
+                lang: {
+                    c: gettext.translation('messages', trad_path, [lang]).gettext(doc) for c, doc in docs.items()
+                } for lang in available
+            })
+        object.__setattr__(self, 'translations', translations)
```
anthonygego: This might be an over-engineering flaw.

I guess that this is placed here to make the classes availables to the frontend module while being instantiated by the different agent modules only. Otherwise, it will fail.

If message passing constraints do not allow to keep this code in the `agent` modules and the idea is to making these more standalone in the future, that's OK. But we should be careful never to instantiate them outside of `agents` where the localisations are. 

Another temporary solution is to merge all localization files into the common folder to keep this architecturally coherent.

anthonygego, line 29: This looks a bit like an over-engineering flaw. I understand the goal...
- Isn't this possible to put the translation loading at the agent side ? Since you already do `object.__setattr__(self, '__translation_path__', 'agent/docker_agent/i18n')` 
- Isn't this possible to load the translations once at the agent init ? There should not be capabilities added at runtime.

inginious/agent/docker_agent/_docker_interface.py
```diff
@@ -37,67 +38,86 @@ def get_cgroup_version(self) -> str:
...
+            }
+            """
+            if (id := img.attrs.get('Id')) is not None:
+                img_data = {'created': created, 'ports': ports}
+                if (env_data := environments.get(env)):
+                    env_data[id] = img_data
+                else:
+                    environments[env] = {id: img_data}
+
+        # TODO: Filter on release tag.
```
anthonygego, line 114: Coupled with `agent_version`, this might triggers a lot of useless debugging frustration. Maybe we should stick to the `agent_version` or replace it with something more reliable. But probably not both.

Actually I would even prefer using the container tag only and drop the `agent_version` label. It is more flexible and allows to retag them without rebuilding everything. Most of all, it would allow to keep an image on the local registry without it being used by the agent, which is sometimes frustrating in production. I think that if the container tag matches the release version, that should be quite enough.

inginious/agent/docker_agent/capabilities.py
```diff
@@ -0,0 +1,20 @@
...
+
+from inginious.common.agents import Capabilities
+
+_ = gettext.gettext
+
+@dataclass(frozen=True)
+class DockerAgentCapabilities(Capabilities):
+
+    def __post_init__(self):
+        object.__setattr__(self, '__translation_path__', 'agent/docker_agent/i18n')
```
anthonygego, line 12: and I understand that the path would depends on the agent type. However, in a first approximation, I'd rather see translation loading here, or something more generic at the agent side, given the MCQ also loads translation files.

inginious/backend/backend.py
```diff
@@ -237,49 +280,55 @@ async def handle_agent_hello(self, agent_addr, message: AgentHello):
...
 
-        self._registered_agents[agent_addr] = AgentInfo(message.friendly_name,
-                                                        [(etype, env) for etype, envs in
-                                                         message.available_environments.items() for env in envs])
-        self._available_agents.extend([agent_addr for _ in range(0, message.available_job_slots)])
+        self._registered_agents[agent_addr] = AgentInfo(
+            message.agent_type,
+            message.friendly_name,
+            message.environments,
+            [c for c, s in message.capabilities.items() if c != 'translations' and s ] if message.capabilities is not None else []
```
anthonygego, line 287: Is there no alternative to remove `translations` from what `items()` returns ?

inginious/common/messages.py
```diff
@@ -64,6 +68,7 @@ class ClientGetQueue:
 class BackendUpdateEnvironments:
     """ Update the information about the environments on the client, from the informations retrieved from the agents """
     available_environments: Dict[str, List[str]]  # dict of available environment aliases (as keys) and type of the related agent (as value)
+    capabilities: dict[AgentType, dict[str, dict[str, str]]]
```
anthonygego, line 71: yummy 😀 (nothing to do)

inginious/frontend/pages/course_admin/task_edit.py
```diff
@@ -6,8 +6,10 @@
 """ Pages that allow editing of tasks """
 import json
 import logging
+import gettext
```
anthonygego, line 9: Not useful I think

inginious/frontend/pages/course_admin/task_edit.py
```diff
@@ -40,15 +42,22 @@ def GET_AUTH(self, courseid, taskid):  # pylint: disable=arguments-differ
         except TaskNotFoundException:
             raise NotFound()
 
-        environment_types = get_all_env_types()
-        environments = self.submission_manager.get_available_environments()
+        environments = {
+            agent_type: {
+              'envs': envs,
+              'capabilities':
+                  caps.get(session.language, caps.get('en', {})) if (caps := self.client.agents_capabilities.get(agent_type, {})) is not None else {},
```
anthonygego, line 49: I guess that `caps` is a dictionary of strings here, so you have to explicitly set the default value ?

inginious/frontend/pages/course_admin/task_edit.py
```diff
@@ -40,15 +42,22 @@ def GET_AUTH(self, courseid, taskid):  # pylint: disable=arguments-differ
...
+              'obj': get_env_from_type(agent_type)
```
anthonygego, line 50: `obj` is not very clear for a key.

inginious/frontend/tasks.py
```diff
@@ -47,15 +48,30 @@ def _migrate_from_v_0_6(content):
 
     return content
 
+def _migrate_from_v_0_10(content):
```
anthonygego, line 51: It's OK but I think this should be the last time we do this. 

We'll have to versionise task descriptors and write a migrator next.

inginious/scripts/agent_docker.py
```diff
@@ -119,9 +97,7 @@ def main():
         context = Context()
 
         # Create agent
-        agent = DockerAgent(context, args.backend, args.friendly_name, args.concurrency,
-                            address_host=args.debug_host, external_ports=args.debug_ports, debugger=args.debugger, tmp_dir=args.tmpdir,
-                            runtimes=args.runtime, ssh_allowed=args.ssh)
+        agent = DockerAgent(context, args.backend, args.friendly_name, args.concurrency, capabilities, address_host=args.debug_host, external_ports=args.debug_ports, debugger=args.debugger, tmp_dir=args.tmpdir)
```
anthonygego, line 100: As long as `backend: local` is supported in the config file, we need to update `inginious/frontend/arch_helper.py` too. I think you should also deprecate this and force users to either launch the components apart or use the compose deployment.

inginious/agent/docker_agent/_docker_interface.py
nrybowski: Agree, I will document that in an issue for a later PR.

inginious/scripts/agent_docker.py
nrybowski: Documented in #1158.

inginious/frontend/tasks.py
nrybowski: Documented in #1159.

inginious/agent/docker_agent/_docker_interface.py
nrybowski: The capabilities have been renamed but the sole I'm aware of is `need_gpu` becoming `gpu`, and there is a single public environment requesting that flag (https://github.com/INGInious/containers/blob/master/grading/python3-ai/Dockerfile). I don't think that it requires bumping the agent version and rebuilding all the environments.

inginious/agent/docker_agent/_docker_interface.py
nrybowski: Documented in #1160.

PR #1156, opened 2026-08-17
Fix user admin error feedback

AlexandreDoneux:
While creating a user with a already taken username, no error was displayed. 

This PR fixes the error feedback display for user creation as well as user activation. The feedback elements all used the same id. When displaying an error, the first available element was used. Therefore, the error was added into a hidden modal.


anthonygego:
This is an OK fix. However, I realise this interface allows to create users with invalid username, email and password. The only accidental check is that username and email are not empty :cry:

Issue #1104, opened 2026-03-24
"My courses" page fails for multiple users when course.yaml "tutors" value is empty


AlexandreDoneux:
**Describe the bug**
When the `tutors` key is in a course.yaml but no value is given, the `/mycourses` page returns an internal error. And that is for all users. Even those that are not part of the course (student, tutor or admin). It is due to the concatenation of the tutors list and the admins list in order to get the staff for a course. `self.get_tutors()` returns a None value.

```
  File "/home/alexandre/INGInious/dev/INGInious/inginious/frontend/user_manager.py", line 960, in has_staff_rights_on_course
    return (username in course.get_staff()) or (include_superadmins and self.user_is_superadmin(username))
                        ~~~~~~~~~~~~~~~~^^
  File "/home/alexandre/INGInious/dev/INGInious/inginious/frontend/courses.py", line 135, in get_staff
    return list(set(self.get_tutors() + self.get_admins()))
                    ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~
TypeError: unsupported operand type(s) for +: 'NoneType' and 'list'
```

We need to handle the cases where no value is given and avoid a full crash. Other config parameters might be in the same situation.

**INGInious installation details**
- Version: v0.9.2

**To Reproduce**
1. Create a course with a `tutors` key with no value in the course.yaml
2. Got to your "my courses" page
3. Internal server error

anthonygego:
@AlexandreDoneux It would be nice to address this issue appropriately. We are a bit fed up of those automatic AI PRs that silently fixes local issues. All fields that should never be `None` should trigger a `CourseUnreadableException`. That is, lists, and most probably dictionaries.

We should also return an `InternalServerError` with the error message in case `CourseUnreadableException` is thrown instead of `NotFound` in the course pages. But beware, we should not expose confidential data through this way.

PR #1161, opened 2026-08-19
Handle None config values for course

AlexandreDoneux:
This PR addresses #1104 

It adds a check on list and dict format to be sure they are never None. An error is raised in the UI informing the user which course and config element is at fault in the `course.yaml`. 
An error is also raised if the task config is erroneous, informing  which task config in `course.yaml` is responsible.

### DIFF
--- a/inginious/common/exceptions.py
+++ b/inginious/common/exceptions.py
@@ -25,7 +25,14 @@ class TaskNotFoundException(Exception):
 
 
 class CourseUnreadableException(Exception):
-    pass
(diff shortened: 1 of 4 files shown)


anthonygego:
This patch is quite complex for what it does, suffers from redundancy and additional risks due to repeated dictionary keys.

We should take profit of more reliable mechanisms. For instance, you can use dataclasses with pydantic to validate the loaded YAML dictionaries according to the expected types of values for each fields, and probably set default values too. This reduces the probability of getting KeyError and issues related to typos as non-existing attributes will automatically trigger a runtime exception.

PR #1162, opened 2026-08-25
Pydantic course config

### LINKED ISSUE #1104: "My courses" page fails for multiple users when course.yaml "tutors" value is empty

AlexandreDoneux:
This PR tackles bad course configurations by using Pydantic as a validation model. That way, instead of a generic "Unreachable Course", the user can see why exactly the course is not properly configured. The error show what course has a faulty config, said config element and why. 

This Pydantic validation model : 
- Checks for the fields' types
- Provides default values
- Forces lti parameters (previously in the Course object)
- Checks for a correct task dispenser
- Checks for correct `accessible` and `registration` parameters and their transformation into `AccessibleTime` objects

Fixes #1104

### DIFF
--- a/inginious/frontend/courses.py
+++ b/inginious/frontend/courses.py
@@ -43,8 +43,6 @@ def _load_course(course_fs : FileSystemProvider, courseid : str):
 class CourseDescriptor(BaseModel):
     model_config = ConfigDict(arbitrary_types_allowed=True) # to allow AccessibleTime() objects in the model
 
-    # TODO : use strict validation ? -> no automatic strict conversion of types, we will need the archive_date validator
-
     name: str
     admins: list[str] = []
     tutors: list[str] = []
@@ -109,7 +107,7 @@ class CourseDescriptor(BaseModel):
             self.accessible = AccessibleTime(self.accessible)
             self.registration = AccessibleTime(self.registration)
         except Exception as e:
-            raise ValueError(f"Invalid accessible or registration time format: {e}")
+            raise ValueError(f"Invalid time format for 'accessible' or 'registration': {e}")
         return self
 
 
@@ -118,10 +116,11 @@ class Course(object):
 
     def __init__(self, courseid, content):
         self._id = courseid
+        self._pre_validated_content = content
         try:
             self._content = CourseDescriptor(**content)
         except ValidationError as e:
-            raise Exception(f"Course has an invalid YAML spec: {courseid}. Validation error: {e}")
+            raise CourseUnreadableException(f"Course '{courseid}' has an invalid '{e.errors()[0]["loc"][0]}' YAML spec: {e.errors()[0]["msg"]}")
 
         self._fs = get_fs_provider().from_subfolder(courseid)
         self._new_doc = not self._fs.exists()
@@ -160,7 +159,7 @@ class Course(object):
             # Here we use a lambda to ensure we do not pass a fixed list of tasks to the task dispenser
             self._task_dispenser = task_dispenser_class(lambda: self.get_tasks(), self._content.dispenser_data, self.get_id())
         except Exception as e:
-            raise Exception("Course has an invalid task dispenser: " + self.get_id() + ". Error: " + str(e))
+            raise CourseUnreadableException(f"Course {self.get_id()} has an invalid task dispenser : {str(e)}")
 
         # Build the regex for the ACL, allowing for fast matching. Only used internally.
         self._registration_ac_regex = self._build_ac_regex(self._registration_ac_list)
@@ -187,9 +186,8 @@ class Course(object):
         return Task.get(self._id, taskid)
 
     def get_descriptor(self):
-        """ Get (a copy) the description of the course """ # change comment to say that it returns a dict instead of a copy of the CourseDescriptor object
-        #return copy.deepcopy(self._content)
-        return self._content.model_dump(exclude_none=False) # python mode to accept non-json-serializables (AccessibleTime, ...)
+        """ Get the description of the course as a python dict. """
+        return self._pre_validated_content # -> is this an alternative ? Needed ?
 
 
     def get_staff(self):
@@ -349,12 +347,11 @@ class Course(object):
         return self._archive_date
 
     def set_descriptor_element(self, key: str, value: Any):
-        self._content[key] = value
+        self._pre_validated_content[key] = value
 
     def save(self):
         """ Saves the Course into the filesystem """
-        #self._fs.put("course.yaml", get_json_or_yaml("course.yaml", self._content))
-        self._fs.put("course.yaml", get_json_or_yaml("course.yaml",  self._content.model_dump()))
+        self._fs.put("course.yaml", get_json_or_yaml("course.yaml", self._pre_validated_content))
         if self._new_doc:
             logging.getLogger("inginious.course").info("Course %s created in the factory.", self._fs.prefix)
 
--- a/inginious/frontend/courses.py
+++ b/inginious/frontend/courses.py
@@ -47,7 +47,7 @@ class CourseDescriptor(BaseModel):
     admins: list[str] = []
     tutors: list[str] = []
     description: str = ""
-    accessible: Optional[str | bool | AccessibleTime] = None # TODO : avoid AccessibleTime instance in the descriptor ?
+    accessible: Optional[str | bool | AccessibleTime] = None
     registration: Optional[str | bool | AccessibleTime] = None
     registration_password: Optional[str] = None
     registration_ac: Optional[Literal["username", "binding", "email"]] = None
@@ -100,15 +100,22 @@ class CourseDescriptor(BaseModel):
             )
         return task_dispenser
 
-    # check AccessibleTime() format
-    @model_validator(mode="after")
-    def verify_accessible_time(self):
+    # check given accessible and registration values are valid
+    @field_validator("accessible", mode="after")
+    @classmethod
+    def verify_accessible(cls, v):
         try:
-            self.accessible = AccessibleTime(self.accessible)
-            self.registration = AccessibleTime(self.registration)
+            return AccessibleTime(v)
         except Exception as e:
-            raise ValueError(f"Invalid time format for 'accessible' or 'registration': {e}")
-        return self
+            raise ValueError(f"Invalid time format.")
+
+    @field_validator("registration", mode="after")
+    @classmethod
+    def verify_registration(cls, v):
+        try:
+            return AccessibleTime(v)
+        except Exception as e:
+            raise ValueError(f"Invalid time format.")
 
 
 class Course(object):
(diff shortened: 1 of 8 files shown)


inginious/frontend/pages/course_admin/utils.py
```diff
@@ -48,7 +49,9 @@ def get_course_and_check_rights(self, courseid, taskid=None, allow_all_staff=Tru
                 return course, None
             else:
                 return course, course.get_task(taskid)
-        except:
+        except CourseUnreadableException as e:
+            raise NotFound(description=_(e))
```
anthonygego: That won't be detected by pybabel. gettext must be called on the raw string, not variables.

inginious/frontend/pages/course.py
```diff
@@ -34,8 +35,8 @@ def get_course(self, courseid):
         """ Return the course """
         try:
             course = Course.get(courseid)
-        except:
-            raise NotFound(description=_("Course not found."))
+        except ValueError as e:
+            raise CourseUnreadableException(str(e))
```
anthonygego, line 39: This does not catch anymore any other exception that may arise. 
Is it the only one that can be raised ? How are managed I/O errors ?

Btw it would probably be much better if the `get` method would raise the `CourseUnreadableException`

inginious/frontend/courses.py
```diff
@@ -37,70 +39,127 @@ def _load_course(course_fs : FileSystemProvider, courseid : str):
...
+
+class CourseDescriptor(BaseModel):
+    model_config = ConfigDict(arbitrary_types_allowed=True) # to allow AccessibleTime() objects in the model
+
+    name: str
+    admins: list[str] = []
+    tutors: list[str] = []
+    description: str = ""
+    accessible: Optional[str | bool | AccessibleTime] = None # TODO : avoid AccessibleTime instance in the descriptor ?
+    registration: Optional[str | bool | AccessibleTime] = None
```
anthonygego: As long as a validator can be written there is no need to avoid.

inginious/frontend/courses.py
```diff
@@ -37,70 +39,127 @@ def _load_course(course_fs : FileSystemProvider, courseid : str):
...
+        self._allow_unregister = self._content.allow_unregister
+        self._allow_preview = self._content.allow_preview
+        self._is_lti = self._content.is_lti
+        self._is_archive = self._content.archived
+        self._archive_date = self._content.archive_date
+        self._lti_url = self._content.lti_url
+        self._lti_keys = self._content.lti_keys
+        self._lti_config = self._content.lti_config
+        self._lti_secrets = self._content.lti_secrets
+        self._lti_send_back_grade = self._content.lti_send_back_grade
```
anthonygego: Is there any need to keep copies or can't we just use `self._content` in getters ?

inginious/frontend/courses.py
```diff
@@ -127,8 +186,9 @@ def get_task(self, taskid):
         return Task.get(self._id, taskid)
 
     def get_descriptor(self):
-        """ Get (a copy) the description of the course """
-        return copy.deepcopy(self._content)
+        """ Get the description of the course as a python dict. """
+        return self._pre_validated_content # -> is this an alternative ? Needed ?
```
anthonygego: So this is the raw dictionary ? It still needs copy to prevent updates by reference

anthonygego:
BTW, tests do not pass.

inginious/frontend/pages/course.py
AlexandreDoneux: `CourseUnreadableException` is already raised by `._load_course()` when it fails to load the course.yaml or when the ´CourseDescriptor` validation fails. 
Or did you mean I should replace the other errors raised by `.get()` by `CourseUnreadableException` ?

anthonygego: You should catch the `CourseUnreadableException` here then and show the detailed error message, but also catch other exceptions. Probably not very useful to have to the message shown to the user then.

inginious/frontend/pages/course_admin/utils.py
```diff
@@ -50,13 +53,17 @@ def get_course_and_check_rights(self, courseid, taskid=None, allow_all_staff=Tru
             else:
                 return course, course.get_task(taskid)
         except CourseUnreadableException as e:
-            raise NotFound(description=str(e))
+            self._logger.error(str(e))
+            raise NotFound(description=_("Course not found."))
```
anthonygego: This one should be explicit, even if there are not many cases this will happen as files are mainly edited through webdav.

By sensitive info I mean clear passwords that may be found in yaml files. In this case, do not expose the values, but you know, as you write the message yourself. The unmanaged exceptions should only be logged as their error message could expose deployment details (path, secrets, passwords...).

AlexandreDoneux:
Added validation for the `archived` and `archive_date` as well as some tests for the Pydantic Model.

inginious/frontend/courses.py
```diff
@@ -37,73 +40,149 @@ def _load_course(course_fs : FileSystemProvider, courseid : str):
 
     return Course(courseid, task_content)
 
+
+class CourseDescriptor(BaseModel):
```
nrybowski, line 44: I would prefer using a built-in dataclass and add the pydantic model decorator. It will simplify the process of changing the model validation library if we need to someday.

inginious/frontend/courses.py
```diff
@@ -37,73 +40,149 @@ def _load_course(course_fs : FileSystemProvider, courseid : str):
...
+    model_config = ConfigDict(arbitrary_types_allowed=True) # to allow AccessibleTime() objects in the model
+
+    name: str
+    admins: list[str] = []
+    tutors: list[str] = []
+    description: str = ""
+    accessible: Optional[str | bool | AccessibleTime] = None
+    registration: Optional[str | bool | AccessibleTime] = None
+    registration_password: Optional[str] = None
+    registration_ac: Optional[Literal["username", "binding", "email"]] = None
```
nrybowski, line 54: Please, put the values in a StrEnum. We need to stop propagating possible keys as plain values in the code.

inginious/frontend/courses.py
```diff
@@ -37,73 +40,149 @@ def _load_course(course_fs : FileSystemProvider, courseid : str):
...
 
+
+class CourseDescriptor(BaseModel):
+    model_config = ConfigDict(arbitrary_types_allowed=True) # to allow AccessibleTime() objects in the model
+
+    name: str
+    admins: list[str] = []
+    tutors: list[str] = []
+    description: str = ""
+    accessible: Optional[str | bool | AccessibleTime] = None
```
nrybowski, line 51: Double check the typing module of 3.14, I think that Optional is not mandatory anymore and that `str | bool | ..` would be sufficient.

inginious/frontend/courses.py
```diff
@@ -37,73 +40,149 @@ def _load_course(course_fs : FileSystemProvider, courseid : str):
...
+        return self
+
+
 class Course(object):
     """ A course with some modification for users """
 
     def __init__(self, courseid, content):
         self._id = courseid
-        self._content = content
+        self._pre_validated_content = content
```
nrybowski, line 156: why keep a copy of maybe malformed data?

inginious/frontend/courses.py
```diff
@@ -37,73 +40,149 @@ def _load_course(course_fs : FileSystemProvider, courseid : str):
...
+        try:
+            self._content = CourseDescriptor(**content)
```
nrybowski, line 158: rename _content to _descriptor so that the field name describes its semantic.

inginious/frontend/courses.py
```diff
@@ -37,73 +40,149 @@ def _load_course(course_fs : FileSystemProvider, courseid : str):
...
+        except ValidationError as e:
+            if bool(set(e.errors()[0]['loc']) & {"admins", "tutors", "registration_password", "registration_ac_list", "lti_secrets", }):  # if the error is related to a sensitive field, do not print the message
+                raise CourseUnreadableException(f"Course '{courseid}' has an invalid '{e.errors()[0]['loc'][0]}' YAML spec.")
+            else:
+                raise CourseUnreadableException(f"Course '{courseid}' has an invalid '{e.errors()[0]['loc'][0]}' YAML spec: {e.errors()[0]['msg']}")
+
```
nrybowski, line 164: Are these static accesses to the error content safe? What if 'loc' is not available in errors()[0]? Why calling two or three times the errors() function while you can keep a local copy of its returned value? Please refactor.

inginious/frontend/courses.py
```diff
@@ -37,73 +40,149 @@ def _load_course(course_fs : FileSystemProvider, courseid : str):
...
-            self._name = self._content['name']
-        except:
-            raise Exception("Course has an invalid name: " + self.get_id())
+        self._name = self._content.name
 
-        if self._content.get('nofrontend', False):
+        if self._content.nofrontend:
             raise Exception("That course is not allowed to be displayed directly in the webapp")
 
+        self._tags = {key: Tag(key, tag_dict, self.gettext) for key, tag_dict in self._content.tags.items()}
```
nrybowski, line 175: Why not validate the tags while loading the course descriptor?

inginious/frontend/courses.py
```diff
@@ -127,48 +206,49 @@ def get_task(self, taskid):
         return Task.get(self._id, taskid)
 
     def get_descriptor(self):
-        """ Get (a copy) the description of the course """
-        return copy.deepcopy(self._content)
+        """ Get the description of the course as a python dict. """
+        return copy.deepcopy(self._pre_validated_content)
```
nrybowski, line 210: Same question as L156, why this function returns a  possibly malformed descriptor?

inginious/frontend/pages/course_admin/utils.py
```diff
@@ -48,8 +52,18 @@ def get_course_and_check_rights(self, courseid, taskid=None, allow_all_staff=Tru
...
+            raise NotFound(description=str(e))
+        except InvalidNameException as e:
+            self._logger.error(str(e))
+            raise NotFound(description=_("Course not found."))
+        except CourseNotFoundException:
+            self._logger.error(f"Course {courseid} not found.")
+            raise NotFound(description=_("Course not found."))
+        except Exception as e:
+            self._logger.error(f"Error while fetching course {courseid}: {str(e)}")
+            raise NotFound(description=_("Course not found."))
```
nrybowski, line 66: I'm unsure about what is happening here, you catch the exact error related to a course, log it then raise a new generic error to display in the frontend I guess. Then you duplicate this exact code in `pages/course.py`. This logic does not belong here nor in `pages/courses.py` but to `frontend/course.py`. If you want to explicitly log the error, pass a reference to a logger when raising the exceptions rather than catching them later to hide the exact error anyway.
To replicate this behavior more cleanly, I suggest you to implement exact course exceptions as subclasses of `CourseException`, so that you emit exact exceptions through subclasses, the exceptions are logged when emitted, then elsewhere you catch the parent class to hide the exact error in the frontend.

inginious/frontend/pages/course.py
```diff
@@ -34,7 +38,17 @@ def get_course(self, courseid):
...
+            self._logger.error(str(e))
+            raise NotFound(description=str(e))
+        except InvalidNameException as e:
+            self._logger.error(str(e))
+            raise NotFound(description=_("Course not found."))
+        except CourseNotFoundException:
+            self._logger.error(f"Course {courseid} not found.")
+            raise NotFound(description=_("Course not found."))
+        except Exception as e:
+            self._logger.error(f"Error while fetching course {courseid}: {str(e)}")
```
nrybowski, line 51: See comment in `course_admin/utils.py`.

nrybowski (changes requested):
A quick first pass on the PR. Please, adapt the test accordingly.

PR #1168, opened 2026-09-07
Add DELETE method for pages

AlexandreDoneux:
Allow pages to use delete methods.

I believe there is no specific behaviour to put in place for delete methods. Is a simple check to make sure the user is logged in enough ?


AlexandreDoneux:
Closing this pull request as it is not the correct way to go. A DELETE method should not redirect to a page as it is the case here if the user is not logged in.

It also concerns the current GET and POST methods. This structure is inherited from our use of webpy in the past. We should refactor it properly and add a proper DELETE instead of just duplicating the problem.

PR #1169, opened 2026-09-09
Changing method signatures to pass username instead of using the session

AlexandreDoneux:
This PR is leading another one that introduces the use of tokens to authenticate for api calls.
It changes multiple method signatures to pass the username  instead of relying on the session, which will be unavailable with the use of the tokens. 

I tried to avoid the use of default None values for the username. Unfortunately, this sometimes produces strange method signatures in my opinion. Do you think it is correct

### DIFF
--- a/inginious/frontend/submission_manager.py
+++ b/inginious/frontend/submission_manager.py
@@ -43,10 +43,10 @@ class WebAppSubmissionManager:
         )
 
     def _job_done_callback(self, submissionid, course, task, result, grade, problems, tests, custom, state, archive, stdout,
-                           stderr, task_dispenser,  newsub=True):
+                           stderr, task_dispenser, username, newsub=True):
         """ Callback called by Client when a job is done. Updates the submission in the database with the data returned after the completion of the
         job """
-        submission = self.get_submission(submissionid, False)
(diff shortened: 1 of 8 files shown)


AlexandreDoneux:
I still need to test the features/pages impacted by these changed.

inginious/frontend/submission_manager.py
```diff
@@ -208,14 +206,14 @@ def get_available_environments(self) -> Dict[str, List[str]]:
         """:return a list of available environments """
         return self._client.get_available_environments()
 
-    def get_submission(self, submissionid, user_check=True):
+    def get_submission(self, submissionid, username, user_check=True):
```
anthonygego: I probably have already said that a number of times: reading your diff is important. And when it comes to refactoring, mandatory.

Introducing a `username` argument here allows a combination of calls that makes no sense given the underlying logic, such as requesting a given submission, with a random `username` and `user_check=False`. This is particulary confusing when reading calls.

You should realize that `username` and `user_check` are actually bound and rewrite the signature to avoid any confusion and restrict its usage to managed cases. Thinking of what are the cases can help.

anthonygego (changes requested):

anthonygego (approved):

PR #1170, opened 2026-09-09
Generation of api tokens and the use of said tokens for the api endpoints

AlexandreDoneux:
As discussed, we divide PR #1121 into smaller ones. This is the first of those.

This PR focuses on the token generation and it's use for the current endpoints. A user can generate up to 20 tokens and giving to each a description. With these tokens he can use the submissions, courses and tasks endpoints, instead of using the flask session and authenticating with the related endpoint. Note the flask session is still a valid way to authenticate to the endpoints.
A user can of course delete a token if needed. 

A PR will follow addressing the security concerns made by @nrybowski and hide information sent back to the user (course not found, task not found, etc.).

I thought about adding a page to the docs regarding the use of the endpoints (authentication with the two different methods). Where should it be located in the docs ?


AlexandreDoneux:
Also note this PR is based on #1169.

inginious/frontend/models/user.py
```diff
@@ -19,9 +25,23 @@ class User(Document):
     ltibindings = MapField(StringField())
     tos_accepted = BooleanField(default=False)
     apikey = StringField(default=None)
+    apitokens = MapField(EmbeddedDocumentField(APIToken), default={})
```
anthonygego: `dict` function or `lambda: {}` is the right way for the `default` argument, otherwise the same reference will be shared accross the app.

inginious/frontend/models/user.py
```diff
@@ -19,9 +25,23 @@ class User(Document):
...
     reset = StringField()
 
     meta = {"collection": "users", "indexes": ["username", "email"]}
+
+    def clean(self):
+        """
+        Custom validation for the User model.
+        """
+        if len(self.apitokens) > 20:
+            raise ValidationError("A user can have at most 20 API tokens.")
```
anthonygego: Use gettext for error messages.

inginious/frontend/models/user.py
```diff
@@ -19,9 +25,23 @@ class User(Document):
...
+        if len(self.apitokens):
+            for token in self.apitokens.values():
+                if len(token.description) > 40:
+                    raise ValidationError("Description must be at most 40 characters long.")
+                if len(token.description) == 0:
+                    raise ValidationError("Description cannot be empty.")
```
anthonygego, line 47: Can't we use `min_length` and `max_length` of `StringField` instead ?

inginious/frontend/pages/api/_api_page.py
```diff
@@ -106,9 +114,35 @@ def _handle_api(self, handler, handler_args, handler_kwargs):
         return APIPage._handle_api(self, (lambda *args, **kwargs: self._verify_authentication(handler, args, kwargs)), handler_args, handler_kwargs)
 
     def _verify_authentication(self, handler, args, kwargs):
-        """ Verify that the user is authenticated """
+        """
+        Verify that the user is authenticated.
+        Checks if the user is logged in through the session. If not, checks for a valid token in the request headers.
+        Otherwise, raises an APIForbidden exception.
+        """
         if not session.loggedin:
```
anthonygego: What is the best thing according to you for retrocompatibility ?

- Trying the session, then only the token, even if it is explicitely mentioned in the request.
- Trying the session only when `Authorization` header is missing ?

inginious/frontend/pages/api/_api_page.py
```diff
@@ -106,9 +114,35 @@ def _handle_api(self, handler, handler_args, handler_kwargs):
...
-            raise APIForbidden()
+            auth_header = flask.request.headers.get("Authorization", "")
+            if not auth_header.startswith("Bearer "):
+               raise APIForbidden("Not authenticated, missing or malformed Authorization header")
+            token = auth_header.removeprefix("Bearer ").strip()
+
+            try:
+               payload = decode_jwt(token)
+            except jwt.ExpiredSignatureError:
+                self.logger.warning("A user has used an expired token.")
```
anthonygego: Not sure this log provides more insight on the issue as the final user is returned the error message explicitely. At best it can be set to debug level, but I'm not even sure this would help in debug.

inginious/frontend/pages/api/_api_page.py
```diff
@@ -106,9 +114,35 @@ def _handle_api(self, handler, handler_args, handler_kwargs):
...
+               self.logger.warning(f"A user has used an invalid token. {str(e)}")
+               raise APIForbidden("Invalid token.")
+            except Exception as e:
+               self._logger.exception(f"Unexpected error while decoding JWT: {str(e)}")
+               raise APIForbidden("Invalid token.")
+
+            flask.g.user = User.objects(username=payload["username"]).first()
+
+            if payload["id"] not in flask.g.user.apitokens.keys():
+                self._logger.exception(f"User {flask.g.user} attempted to use an API token that is not present in the database.")
```
anthonygego: At best warn level. Exception level should be reserved for...exceptions, critical issues.

inginious/frontend/pages/api/courses.py
```diff
@@ -54,19 +56,24 @@ def API_GET(self, courseid):  # pylint: disable=arguments-differ
             except:
                 raise APINotFound("Course not found")
 
-        username = session.username
+        if session.loggedin:
+            username = session.username
+        else:
+            username = flask.g.user.username
```
anthonygego: Same question on retrocompatibility: session first, even if the `Authorization` header has been explicitely passed ?

BTW, maybe you can simply set `flask.g.user` in both case in `_verify_authentication` to avoid a second check.

inginious/frontend/pages/api/submissions.py
```diff
@@ -111,7 +110,12 @@ def API_GET(self, courseid, taskid, submissionid):  # pylint: disable=arguments-
         """
         with_input = "input" in flask.request.args
 
-        return _get_submissions(self.submission_manager, self.user_manager, courseid, taskid, with_input, submissionid)
+        if session.loggedin:
+            username = session.username
+        else:
+            username = flask.g.user.username
```
anthonygego: Same remark, set `flask.g.user` only once, do not perform multiple tests.

inginious/frontend/pages/api/submissions.py
```diff
@@ -152,7 +156,12 @@ def API_GET(self, courseid, taskid):  # pylint: disable=arguments-differ
         """
         with_input = "input" in flask.request.args
 
-        return _get_submissions(self.submission_manager, self.user_manager, courseid, taskid, with_input)
+        if session.loggedin:
+            username = session.username
+        else:
+            username = flask.g.user.username
```
anthonygego: And again.

inginious/frontend/pages/api/submissions.py
```diff
@@ -172,7 +181,10 @@ def API_POST(self, courseid, taskid):  # pylint: disable=arguments-differ
         except:
             raise APINotFound("Course not found")
 
-        username = session.username
+        if session.loggedin:
+            username = session.username
+        else:
+            username = flask.g.user.username
```
anthonygego: ....and again.

inginious/frontend/pages/api/submissions.py
```diff
@@ -188,20 +200,31 @@ def API_POST(self, courseid, taskid):  # pylint: disable=arguments-differ
...
+        if flask.request.is_json:
+            user_input = flask.request.get_json()
+            for problem in task.get_problems():
+                pid = problem.get_id()
+                if problem.input_type() == list:
+                    value = user_input.get(pid, [])
+                    user_input[pid] = value if isinstance(value, list) else [value]
+                elif problem.input_type() == dict:
+                    # File inputs are not supported in JSON requests.
+                    raise APIInvalidArguments()
```
anthonygego: 1) Why ?
2) If that should be the case (but it shouldn't), an explicit error message should be returned, and this should be documented.
3) What about the other input types here (that is, str) ?

inginious/frontend/pages/api/tasks.py
```diff
@@ -67,7 +69,14 @@ def API_GET(self, courseid, taskid):  # pylint: disable=arguments-differ
         except:
             raise APINotFound("Course not found")
 
-        if not self.user_manager.course_is_open_to_user(course, lti=False):
+        if session.loggedin:
+            username = session.username
+        else:
+            username = flask.g.user.username
```
anthonygego: once again....

inginious/frontend/pages/api/tasks.py
```diff
@@ -67,7 +69,14 @@ def API_GET(self, courseid, taskid):  # pylint: disable=arguments-differ
...
+
+        user = User.objects(username=username).first()
```
anthonygego: If you do not expect a possible `None` value, use `User.get` that will raise an Exception if the user is not found. If you stick with `.first()`, check for `None` before continuing.

inginious/frontend/pages/preferences/apitoken.py
```diff
@@ -0,0 +1,87 @@
...
+
+            if expires_in == "custom":
+                expires_in = request.form.get("custom_days")
+
+            try:
+                days = int(expires_in)
+            except (TypeError, ValueError):
+                return self.show_page(errors=["Please select a valid expiration duration."])
+
+            if days < 10 or days > 365:
```
anthonygego:
```suggestion
            if not 10 <= days <= 365:
```

inginious/frontend/pages/preferences/apitoken.py
```diff
@@ -0,0 +1,87 @@
...
+    """ Page to view or generate an API token for the user """
+
+    def GET_AUTH(self):
+        """ GET request """
+        return self.show_page()
+
+    def POST_AUTH(self):
+        """ POST request, generates a new token for the user """
+
+        user = User.objects(username=session["username"]).first()
```
anthonygego: Same, `get` or `.first()` + check for `None`.

inginious/frontend/pages/preferences/apitoken.py
```diff
@@ -0,0 +1,87 @@
...
+            description = request.form.get("description")
+            expires_in = request.form.get("expires_in", 10)
+
+            if expires_in == "custom":
+                expires_in = request.form.get("custom_days")
+
+            try:
+                days = int(expires_in)
+            except (TypeError, ValueError):
+                return self.show_page(errors=["Please select a valid expiration duration."])
```
anthonygego: Use gettext for end-user error messages

inginious/frontend/pages/preferences/apitoken.py
```diff
@@ -0,0 +1,87 @@
...
+            }
+
+            token = encode_jwt(payload)
+
+            try:
+                new_token = APIToken(token=UserManager.hash_password(token), expires=expiration, description=description)
+                user.apitokens[token_id] = new_token
+                user.save()
+            except ValidationError as e:
+                return self.show_page(errors=[list(e.to_dict().values())[0]])
```
anthonygego, line 62: Why only keeping the first ?

inginious/frontend/pages/preferences/apitoken.py
```diff
@@ -0,0 +1,87 @@
...
+            if token_id not in user.apitokens:
+                return self.show_page(errors=["Token not found."])
+
+            del user.apitokens[token_id]
+            user.save()
+
+            return self.show_page()
+
+    def show_page(self, generated_token=None, errors=None):
+        """ Prepares and shows the course marketplace """
```
anthonygego: marketplace ?

inginious/frontend/pages/preferences/apitoken.py
```diff
@@ -0,0 +1,87 @@
...
+        if errors is None:
+            errors = []
+
+        user = User.objects(username=session["username"]).first()
```
anthonygego: `get` if no `None` is expected.

inginious/frontend/pages/jwt_utils.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+
+    secrets = [current_app.config["API_JWT_SECRET"]] + current_app.config["API_JWT_OLD_SECRETS"]
+
+    for secret in secrets:
+        try:
+            return jwt.decode(token, secret, algorithms=[current_app.config["API_JWT_ALGORITHM"]])
+        except jwt.ExpiredSignatureError as e:
+            # correct secret found, token expired
+            raise e
```
anthonygego: Why catching an error and just raising it afterwards ?

inginious/frontend/pages/jwt_utils.py
```diff
@@ -0,0 +1,37 @@
...
+Helpers to encode/decode the API JWT tokens.
+
+"""
+
+import jwt
+
+from flask import current_app
+
+
+def encode_jwt(payload):
```
anthonygego: This method is used once. Maybe it is not necessary to put it in a separate module.

inginious/frontend/pages/jwt_utils.py
```diff
@@ -0,0 +1,37 @@
...
+    return jwt.encode(payload, current_secret, algorithm=current_app.config["API_JWT_ALGORITHM"])
+
+
+def decode_jwt(token):
+    """
+    Decodes a JWT, trying the current secret first, then falling back to the previously used secrets
+    to support older tokens.
+    """
+
+    secrets = [current_app.config["API_JWT_SECRET"]] + current_app.config["API_JWT_OLD_SECRETS"]
```
anthonygego: What is the purpose of keeping old secrets ?

inginious/frontend/templates/apitoken.html
```diff
@@ -0,0 +1,128 @@
...
+{% endfor %}
+
+<div class="col-sm-12">
+    <p>
+        {{ _("Your API tokens allows external applications and scripts to authenticate as you and interact with "
+             "INGInious on your behalf, without needing your password. Keep them secret: anyone who has them can "
+             "access the API using your account permissions. If you believe on of your tokens has been compromised, "
+             "delete it from the list and use another one.") }}
+    </p>
+</div>
```
anthonygego, line 39: Maybe put this in a dismissable alert

inginious/frontend/templates/apitoken.html
```diff
@@ -0,0 +1,128 @@
...
+
+        <div class="mb-2">
+            <input type="text" class="form-control" value="{{ generated_token }}" readonly/>
+        </div>
+
+        <form method="get" action="{{ url_for('.apitokenpage') }}">
+            <button type="submit" class="btn btn-primary btn-block">
+                <i class="fa fa-check"></i> {{ _("Close") }}
+            </button>
+        </form>
```
anthonygego: What is the purpose of this ? There is an input out of the `<form/>` and the form only contains a submit button...

inginious/frontend/templates/apitoken.html
```diff
@@ -0,0 +1,128 @@
...
+    </div>
+</div>
+{% else %}
+    <form method="post" action="{{ url_for('.apitokenpage') }}">
+        <div class="col-sm-12"><br />
+            <button type="button" id="show-generate-fields-btn" class="btn btn-primary btn-block"
+                    onclick="document.getElementById('generate-fields').classList.remove('d-none'); this.classList.add('d-none');">
```
anthonygego: You can use JQuery `on("click", function() {...} )` instead, and use JQuery selectors instead of vanilla JS `getElementById`. Also check if we can't do that with toggling methods.

inginious/frontend/templates/apitoken.html
```diff
@@ -0,0 +1,128 @@
...
+            <div id="generate-fields" class="d-none mt-2">
+                <div class="mb-2">
+                    <input type="text" class="form-control" name="description"
+                           placeholder="{{ _('Provide a description for your token.') }}" />
+                </div>
+
+                <div class="mb-2">
+                    <label for="expires_in" class="form-label">{{ _("Expiration") }}</label>
+                    <select id="expires_in" name="expires_in" class="form-control"
+                            onchange="document.getElementById('custom-expiration-field').classList.toggle('d-none', this.value !== 'custom');">
```
anthonygego: Same comment, you can use JQuery `onChange` in a JS snippet put aside.

inginious/frontend/installer.py
```diff
@@ -591,6 +594,14 @@ def configure_authentication(self):
 
         return options
 
+    def configure_api(self):
+        """ Configure the API parameters """
+        options = {}
+        options["api_jwt_secret"] = hexlify(os.urandom(32)).decode('utf-8')
```
anthonygego: Use `secrets` module instead

configuration.deploy.yaml
```diff
@@ -14,3 +14,6 @@ superadmins:
 tasks_directory: /inginious/tasks
 use_minified_js: true
 webdav_host: https://localhost/dav
+api_jwt_old_secrets:
+- c9f5542edfa41190589bce633ae9901ed2cb6c99be367016c8d88fc698836c41
+api_jwt_secret: c9f5542edfa41190589bce633ae9901ed2cb6c99be367016c8d88fc698836c41
```
anthonygego: Currently please at least put a comment in `README.md` where it is mentioned to change the Flask `secret_key` . You can add a `# CHANGE THIS` comment in this file.

anthonygego (changes requested):
Things are getting better but there is still  issues to be addressed.

Please also add some docs in the developer section.

inginious/frontend/models/user.py
AlexandreDoneux: We can, but the ValidationError has a different format. Same if we moved that custom validation in `APIToken`. I was thinking about a shared method among models to display errors. But I feel it should not belong to this PR.

inginious/frontend/pages/preferences/apitoken.py
AlexandreDoneux: As I implemented validation through the `clean` method alone, only one error was listed by validation. I will change that behaviour when implementing a general error displaying method for all models as discussed.

AlexandreDoneux:
I will address the model validation remarks in a following PR to not overcharge this one.

doc/dev_doc/API/api_usage.rst
```diff
@@ -0,0 +1,78 @@
...
+The current API allows :
+
+- Checking available authentication methods, authentication through session and checking if the session is valid
+- Listing courses, and their informations, available to the authenticated client
+- Listing tasks, and their informations, available to the authenticated client
+- Listing all the submissions that the authenticated client made for a specific course or task
+
+For more information about the available endpoints, please refer to the :ref:`API documentation.<rest_api_documentation>`.
+
+Authentication through session
```
anthonygego, line 24: Please indicate this is deprecated and will be removed in a future release.

inginious/frontend/pages/api/submissions.py
```diff
@@ -188,20 +192,34 @@ def API_POST(self, courseid, taskid):  # pylint: disable=arguments-differ
...
+                if problem.input_type() == list:
+                    value = user_input.get(pid, [])
+                    user_input[pid] = value if isinstance(value, list) else [value]
+                elif problem.input_type() == dict:
+                    # File inputs are not supported in JSON requests. Needs to be sent in base64 encoded format
+                    value = user_input.get(pid)
+                    if isinstance(value, dict) and "filename" in value and "value" in value:
+                        user_input[pid] = {
+                            "filename": value["filename"],
+                            "value": base64.b64decode(value["value"])
```
anthonygego, line 208: Beware, this can raise an exception.

inginious/frontend/pages/preferences/apitoken.py
anthonygego: Where is it changed ?

inginious/frontend/installer.py
```diff
@@ -591,6 +594,13 @@ def configure_authentication(self):
 
         return options
 
+    def configure_api(self):
+        """ Configure the API parameters """
+        options = {}
+        options["api_jwt_secret"] = secrets.token_hex(16)
+
+        return options
```
anthonygego, line 602:
```suggestion
        return {"api_jwt_secret": secrets.token_hex(16)}
```

inginious/frontend/task_problems.py
```diff
@@ -133,6 +133,10 @@ def get_type_name(cls, language):
         return _("file upload")
 
     def adapt_input_for_backend(self, input_data):
+        # if file is already in the correct format
+        if isinstance(input_data[self.get_id()], dict) and set(input_data[self.get_id()].keys()) == {"filename", "value"}:
+            return input_data
+
```
anthonygego, line 139: I'm wondering if this is possible to avoid this "multi-format" tweak and either:
- Generate the dict before this method is called, as there is in all cases a specific logic to fetch flask file inputs. That means putting the logic of this method outside. Not sure to win anything by doing this.
- Mimick Flask behaviour in `/api/submissions.py` and produces a ``werkzeug.datastructures.FileStorage`` when the request is made using base64.
  Based on the constructor here: https://github.com/pallets/werkzeug/blob/main/src/werkzeug/datastructures/file_storage.py
  That would consist in something like :
       
        from werkzeug.datastructures import FileStorage
        from io import BytesIO

        user_input[pid] = FileStorage(BytesIO(base64.b64decode(value["value"]), 
                                      value["filename"])
   which would be more elegant given the current implementation at the cost of a `BytesIO` encapsulation.   @nrybowski  your thought on this ?

inginious/frontend/pages/preferences/apitoken.py
AlexandreDoneux: I was thinking adding that in a future PR to avoid confusion. Plus it is not only related to the API.

doc/dev_doc/API/api_usage.rst
```diff
@@ -0,0 +1,80 @@
+INGInious API
+==============
+
+This page is dedicated to the RESTful API of INGInious. It is intended for developers to access INGInious features outside of the UI.
+
+If you are looking for the runfile API, learn more about it the :ref:`runfile introduction.<run_file_api>`
+
+Using the API
+-------------
+Multiple endpoints are available to interact with INGInious. The API is RESTful and uses session or token authentication.
```
anthonygego, line 10:
```suggestion
Multiple endpoints are available to interact with INGInious. The API is RESTful and uses cookie or token authentication.
```

doc/dev_doc/API/api_usage.rst
```diff
@@ -0,0 +1,80 @@
...
+The current API allows :
+
+- Checking available authentication methods, authentication through session and checking if the session is valid
+- Listing courses, and their informations, available to the authenticated client
+- Listing tasks, and their informations, available to the authenticated client
+- Listing all the submissions that the authenticated client made for a specific course or task
+
+For more information about the available endpoints, please refer to the :ref:`API documentation.<rest_api_documentation>`.
+
+Authentication through session
```
anthonygego, line 24:
```suggestion
Authentication through cookie
```

This better fits what is actually done by the user.

doc/dev_doc/API/api_usage.rst
```diff
@@ -0,0 +1,80 @@
...
+```````````````````````````````
+Using the API with the flask session will need the use of the authentication endpoint.
+The following example shows how to authenticate with /api/v0/authentication endpoint.
+You can check if the session is valid by sending a GET request to the same endpoint.
+
+Note this method is deprecated and will be removed in a future release. It is recommended to use the token authentication method instead.
```
anthonygego, line 30: Put this in a warning or danger admonition so that this is clearly visible.

doc/dev_doc/API/api_usage.rst
```diff
@@ -0,0 +1,80 @@
...
+    -F "studentupload=@/path/to/file/my_file.txt" -F "studentcode=some_code" -F "thecode=some_other_code"
+
+
+Authentication through a token
+```````````````````````````````
+
+Authentication through a token does not require the use of the authentication endpoint.
+It is also safer because is generated by the user in the UI and can be revoked at any time.
+
+To create a token, connect to your profile in the webapp and go tou your user preferences.
```
anthonygego, line 58:
```suggestion
To create a token, connect to your profile in the webapp and go to your user preferences.
```

inginious/frontend/models/user.py
```diff
@@ -19,9 +25,23 @@ class User(Document):
     ltibindings = MapField(StringField())
     tos_accepted = BooleanField(default=False)
     apikey = StringField(default=None)
+    apitokens = MapField(EmbeddedDocumentField(APIToken), default=dict())
```
anthonygego, line 28:
```suggestion
    apitokens = MapField(EmbeddedDocumentField(APIToken), default=dict)
```
`dict() == {}`.

inginious/frontend/pages/api/_api_page.py
```diff
@@ -7,15 +7,22 @@
...
+from inginious.frontend.models import User
 
 
 class APIPage(INGIniousPage):
     """ Generic handler for all API pages """
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._logger = logging.getLogger("inginious.frontend.api")
+
```
anthonygego, line 25:
```suggestion
    self._logger = logging.getLogger("inginious.frontend.api")
```

as a class attribute is enough.

inginious/frontend/pages/api/submissions.py
```diff
@@ -32,26 +34,26 @@ def _get_submissions(submission_manager, user_manager, courseid, taskid, with_in
...
-        submissions = submission_manager.get_user_submissions(course, task)
+        submissions = submission_manager.get_user_submissions(course, task, username)
     else:
         try:
-            submissions = [submission_manager.get_submission(submissionid)]
+            submissions = [submission_manager.get_submission(submissionid, as_username=username)]
         except:
             raise APINotFound("Submission not found")
-        if submissions[0].taskid != task.get_id() or submissions[0].courseid != course.get_id():
+        if submissions[0]["taskid"] != task.get_id() or submissions[0]["courseid"] != course.get_id():
```
anthonygego, line 43: `submissions[0]` will exist and be `None` in case we try to access an existing submission that isn't ours. Besides, `get_submission` raises an exception in case the submission does not exist.

If calls to `get_submission` allows that, it's probably better to also return `None` (that is, use `.first()`) in case the submission does not exist. But before that double check that the calls would correctly handle that.

inginious/frontend/app.py
```diff
@@ -54,12 +54,20 @@ def _put_configuration_defaults(config):
...
+        print("Refer to the documentation for more information.", file=sys.stderr)
+        exit(1)
+
     # Populate a sanitized new dict with upper chars for Flask
     new_config = {
         "ALLOWED_FILE_EXTENSIONS": config.get('allowed_file_extensions',
                                               [".c", ".cpp", ".java", ".oz", ".zip", ".tar.gz", ".tar.bz2", ".txt"]),
         "ALLOW_DELETION": config.get("allow_deletion", True),
         "ALLOW_REGISTRATION": config.get("allow_registration", True),
+        "API_JWT_ALGORITHM": "HS256",
```
anthonygego, line 69: At this stage this is not useful here, you can hardcode this in the code.

inginious/frontend/submission_manager.py
```diff
@@ -262,7 +263,7 @@ def add_job(self, course, task, inputdata, task_dispenser, debug=False):
 
         # Send LTI information to the client except "consumer_key"
         # to_dict() to avoid sending mongoengine BaseLists to ZMQ
-        if session.is_lti:
+        if session.is_lti: # TODO ; check correct behavior when no session
```
anthonygego, line 266:
```suggestion
        if session.is_lti: # TODO ; refactor not to use session here
```

It will run as `is_lti` defaults to `False`. But will have to be re-worked later.

anthonygego (changes requested):

inginious/frontend/pages/api/submissions.py
AlexandreDoneux: What about what I've done here ? Returning an empty list of submissions if the submission does not belong to the user and if the submission does not exist. 

I will add logging for the error (task and course not corresponding) in the next PR.

inginious/frontend/submission_manager.py
```diff
@@ -187,7 +188,7 @@ def replay_job(self, course, task, submission, task_dispenser, copy=False, debug
         jobid = self._client.new_job(1, job_info, inputdata,
                                      (lambda result, grade, problems, tests, custom, state, archive, stdout, stderr:
                                       self._job_done_callback(submissionid, course, task, result, grade, problems, tests,
-                                                              custom, state, archive, stdout, stderr, task_dispenser, copy)),
+                                                              custom, state, archive, stdout, stderr, task_dispenser, username, copy)),
```
anthonygego: Once again, please read your diffs before asking for review.

This rebase issue is easily detectable both with the eye and by running the code as no submission will end.

anthonygego (changes requested):

inginious/frontend/submission_manager.py
anthonygego, line 273: Note : I actually don't see any good reason for keeping this code. I don't find any usage in our own task database, it forwards host-platform specific information, it is not documented, and it only filters out LTI 1.1 critical fields.

I think it can just be removed in another PR.

anthonygego (approved):
This has eventually reached my acceptance level.

@nrybowski can you perform a last pass on this ?

PR #1171, opened 2026-09-11
[lti1.3] add launch messages expiration

anthonygego:
This PR adds support for expiring LTI1.3 launch messages from the database storage. These documents are not expected to live longer than the LTI session.

- It adds the expirable index on the mongo collection
- It adds the expiration date given by `lti1p3` lib through the `LaunchDataStorage` interface.
- It provides a migration for the collection, making documents expire the next day (the default from `lti1p3`)
- Since launch messages should be available during the whole LTI session lifetime for AGS service to send scores, it disable LTI session lifetime extension. It only lives the time set in config file.
- It sets the launch message expiration time to the flask session expiration time.


anthonygego:
@nrybowski Pushed commit to disable LTI sessions expiration extension and put the same expiration time to sessions and launch message, so that the launch messages will expire with the session.

PR #1173, opened 2026-09-14
Limiting API feedback to client

AlexandreDoneux:
Based on #1170, this PR changes the amount of information returned to the client calling the API endpoints. As mentioned in the comments of #1121,  returning too much information can allow users to look for hidden courses, tasks, etc. 

Also adding logging to monitor plausible attacks and troubleshoot.


AlexandreDoneux:
Should I be more clear for each log by indicating the endpoint/class logging the error ?

PR #1176, opened 2026-09-16
Fix KeyError when course staff don't have accounts yet

thaugdahl:
**Problem**
Pages such as the course submissions page combine students and staff into one username list and pass it into `get_users_info` in `UserManager`. The Mongo query only returns users that exist, so the returned dict has no key for staff who haven't logged in yet. Callers that index into it with a staff name hit `KeyError`s.

This scenario happened on our self-hosted INGInious where we have OIDC login and a handwritten tutor-list in `course.yaml`.

**Fix**
`get_users_info` now includes every requested username in its result. Usernames that aren't found in the query are mapped to `None`. Callers seem to already handle `None` users fine.

Missing usernames are only added when there's no pagination (`limit` and `skip` are unset). With pagination, foregoing that predicate might include users the page shouldn't contain.


thaugdahl:
It seems this PR also included a separate fix I was working on. Closing and reopening them as separate PRs.

PR #1177, opened 2026-09-16
[usermanager]: Fix userinfo for non-existent staff users

thaugdahl:
**Problem**
Pages such as the course submissions page combine students and staff into one username list and pass it into `get_users_info` in `UserManager`. The Mongo query only returns users that exist, so the returned dict has no key for staff who haven't logged in yet. Callers that key into the returned value may then cause `KeyError`s.

This scenario happened on our self-hosted INGInious where we have OIDC login and a handwritten tutor-list in `course.yaml`.

**Fix**
`get_users_info` now includes every requested username in its result. Usernames that aren't found in the query are mapped to `None`. Callers seem to already handle `None` users fine.


anthonygego:
Thanks for notifying the issue. The fix you suggested made me realise this method needed some refactor. I did it in #1179, credited you as well.

I'm closing this one.

PR #1178, opened 2026-09-16
Fixed skip and limit filters on course admin submissions page

thaugdahl:
**Problem:**
Pagination is silently ignored on the course admin submissions page. 

**Fix**:
MongoEngine `QuerySet.limit()` and `skip()` return a new `QuerySet` instead of mutating in place. This fix assigns the new queries.


anthonygego:
Good catch, thanks !

PR #1179, opened 2026-09-16
[user_manager/user_info] refactor get_users_info and UserInfo

anthonygego:
This addresses issue reported in #1177 and provides a first refactor `UserManager.get_users_info`. 

- `get_users_info` was actually used in two cases with different signatures (`usernames` or `limit, skip` actually used)
   - recovered the previous signature without optional arguments, as well as the default to `None`.
   - usage with limit and skip did not need a dictionary mapping: replaced by a simple mongoengine request as a first approximation, and updated the template accordingly.
- `UserInfo` was a redundant data structure since mongoengine migration. Removed it and returned `User` model directly. Adapted code. 
   I recall this was introduced to limit the risk of exposing the raw dictionary with confidential data in templates, but mongoengine does not allow to dump the whole model values implicitly, so the requirements are kept.


inginious/frontend/courses.py
```diff
@@ -146,7 +146,7 @@ def is_open_to_non_staff(self):
         """ Returns true if the course is accessible by users that are not administrator of this course """
         return self.get_accessibility().is_open()
 
-    def is_registration_possible(self, user_info: UserInfo):
+    def is_registration_possible(self, user_info: User):
```
nrybowski: We could also rename the variable

inginious/frontend/courses.py
```diff
@@ -229,7 +229,7 @@ def lti_send_back_grade(self):
         """ True if the current course should send back grade to the LTI Tool Consumer """
         return self._is_lti and self._lti_send_back_grade
 
-    def is_user_accepted_by_access_control(self, user_info: UserInfo):
+    def is_user_accepted_by_access_control(self, user_info: User):
```
nrybowski: Same comment

nrybowski (commented):
LGTM from a first pass, I still have to give it a try on a test instance.

nrybowski (approved):

PR #1184, opened 2026-09-21
[studio] refactor task file upload

anthonygego:
This PR :

- removes the buggy ``dd_upload`` endpoint.
- makes the drag and drop perform the same request as the file upload modal.
- removes the hidden `<input type="file" />` that does not seem to be of any use.

Superseeds and closes #1167.


nrybowski (commented):
LGTM, still have to give it a try.

nrybowski (approved):

